PR #24, opened 2016-12-10
fixed probable bugs

pinpong:


src/main/java/com/iota/iri/IRI.java
```diff
@@ -122,13 +122,13 @@ private static void validateParams(final String[] args) {
...
+	    	StatusPrinter.print((LoggerContext)LoggerFactory.getILoggerFactory());
 	    }
 	    
 	    Configuration.put(DefaultConfSettings.API_PORT, cport);
-	    Configuration.put(DefaultConfSettings.NEIGHBORS, cns.toString());
-	
-	    if (Integer.parseInt(cport) < 1024) {
+	    Configuration.put(DefaultConfSettings.NEIGHBORS, cns);
+
+		if (cport != null && Integer.parseInt(cport) < 1024) {
```
davassi: cport here cannot be null. There's a check before.

 if (cport == null) {
    		log.error("Invalid arguments list. Provide api port number with -p or --port");
			printUsage();
	    }

PR #28, opened 2016-12-12
fixed probable bugs

pinpong:


src/main/java/com/iota/iri/service/dto/AbstractResponse.java
```diff
@@ -22,10 +22,16 @@ public int hashCode() {
     }
 
     @Override
-    public boolean equals(Object obj) {
+    public boolean equals(final Object obj) {
+        if (this == obj) {
+            return true;
+        }
+        if ((obj == null) || !(obj instanceof AbstractResponse)) {
```
davassi: those 2 conditions are already in the reflectionEquals()

pinpong: Updated

PR #34, opened 2016-12-20
Add jre version to node info

mark-bradshaw:
Minor change to nodeInfo command to also return the jre version.  The intent is to provide a bit more info when troubleshooting nodes (mostly other people's) who are having issues.  It's also a way to get my feet wet a bit with IRI.


davassi:
:+1:

PR #35, opened 2016-12-23
Threading issues in PearlDiver.java

alfredch:
Tests with attachToTangle for a bundle with 4 txs showed the following situation:
In 7 of 21 cases, one of the 4 nonces was returned as 999999999...

PearlDiver.search() has an IN-OUT-argument. Tracing proved that the the transactionTrits are calculated correctly - for the fault nonces as well - but after return from search(), when using the trits for building the transaction object, the nonce is 999999999... This happens most frequently when the spam loop is active and then a concurrent call to attachToTangle is made.

A further problem is created by the missing join() within the PearDiver thread structure. The worker threads have a chance to continue running when they should terminate because the creator does not care to wait for their termination.

### DIFF
--- a/src/main/java/com/iota/iri/hash/PearlDiver.java
+++ b/src/main/java/com/iota/iri/hash/PearlDiver.java
@@ -91,10 +91,12 @@ public class PearlDiver {
             }
         }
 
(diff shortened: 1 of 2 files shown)


davassi:
@alfredch nice work

Issue #33, opened 2016-12-18
Race condition in PearlDiver (iri version 1.1.2.2 dev)


alfredch:
There is this severe problem in PearlDiver which causes a sporadic return of 99999999... nonce:
The flag 'state' controls the inner and outer thread(s) in the search() function.
When one of the inner threads finds a success (minWeightMagnitude  counted down to 0), then it set state = COMPLETED and notifyALL().
That causes the main thread to terminate and the function search to return.
However, the inner thread continues to run because the function search() is called again in the mean time. 
Probably (not checked by myself) the same instance of PearlDiver is used for a multi-tx bundle. So the new call to search(...) can set state = RUNNING again before the old inner thread executes the 'while (state == RUNNING)' check. So that one continues and messes up the work of the regular diver thread.

If in doubt, trace the situation with log.debug within the diver thread. I did it and it happens every so often (Maybe you have to repeat a 4tx bundle' attachToTangle 20 times until that overlapping is visible, but on a average it happen every 4th time).

alfredch:
The IRI-version that I was referring to is 1.1.2.2-dev, not 1.2.2.2-dev.

davassi:
is it related to #31?

alfredch:
Yes - it's the same. Dominik asked for help - sorry I didn't see he had already created an issue.

alfredch:
I have identified 2 problems related to that issue:
1) PearlDiver search() function is not thread-safe because of in-out-argument 'transactionTrits' and the usage thereof in the internal worker threads.
2) The internal worker threads are not controlled safely by the state flag. There is a time window where a caller can re-establish state RUNNING before the worker thread tests for the actual state. That means the inner threads will potentially continue executing also the intention was that they terminate.
I will make a PR request with the minimal fixes, avoiding a major re-coding of the class.

Issue #31, opened 2016-12-18
PearlDiver Threading Issue


domschiener:
I run the testnet version on a two core DigitalOcean VPS (Ubuntu 16.04, 2GB RAM,40gb ssd). It seems that PearlDiver leads to an incorrect output in about 25% of the time. What I mean with that specifically, is that when doing `attachToTangle`, one of the transactions in the bundle usually has an invalid transaction hash, as well as an all 9's nonce. 

Here is an example output on two different machines, with the exact same input (trytes, branch and trunk, as well as minWeightMagnitude):

4 cores, MacOS: http://pastebin.com/Du5JfZWG 

2 cores, Ubuntu 16.04: http://pastebin.com/zV7h51Dc

Considering that PearlDiver / Curl uses n - 1 cores, this is probably related with doing POW on a single core machine. Any thoughts?

paulhandy:
It sets the number of threads to 1 if there is only a single core. Threading in this case would just result in the main process being paused while the child process is executed.

PR #37, opened 2016-12-28
Thread safety fix and loop elimination

paulhandy:
Should solve #31 and #33, unit test included!


davassi:
Please never PR directly on master, but prefer dev branch instead.

PR #39, opened 2016-12-28
Obviated the bitIndex loop

paulhandy:
Instead of bitshifts and a loop through bitIndex, this change directly calculates a successful nonce by combining an inverse XOR of the low and high transformed states with a mask originally set to high bits. Then, only one bit of this is taken for the output hash.
In the greatest case, this combines 64 iterations into one. In the least case, it performs as well as its previous counterpart.


davassi:
@Come-from-Beyond

paulhandy:
I have junit tests that I can add, as well. Not sure if that's welcome, but I think it would be beneficial.

PR #41, opened 2016-12-29
Retry if bundle not complete.

alfredch:
If the bundle information (transaction list) was not available for a new milestone, then no retry was attempted to validate the new milestone.
Tests show that the bundle information is complete with high probability in the next call to 'updateLatesMilestone()'. That function is called every 5000 ms by the 'Latest Milestone Tracker' thread.

The logic suggested in this fix could be improved by counting the number of retries. If the bundle is not complete after some time - let's say 1 minute - there is little chance that it will ever be complete.


alfredch:
I closed this pull request because the code in it is not complete.

PR #43, opened 2016-12-29
Queue milestone tx for retry if bundle incomplete

alfredch:
Often, the bundle information is not available when the function 'updateLatestMilestone()' is called and a higher milestone index is found.
When the function is called again in 5000 ms by the updater thread, the bundle is complete and the index update is possible.
Without this fix, a new milestone is evaluated a 2nd time.


davassi:
@alfredch Sorry just only a favour please: do you mind to recreate that PR against dev branch and not master?

alfredch:
Added a pull request with identical code to dev branch.

PR #49, opened 2017-01-17
Created iota extension interface

paulhandy:
Javascript files can be created in a folder, defaulted to "./ixi/", subject to javax.script.scriptengine constraints. Currently, they are automatically loaded into the IRI, and have full access to all classes, except the IXI class itself.

A script can register a "shutdown" function, so that it can be cleanly removed at the call of the IRI shutdown.

API endpoints can be registered, and are callable by `<script name>.<name of registered command>`.


paulhandy:
Cancelling this pull request

PR #59, opened 2017-01-25
CORS preflight channel did not succeed fix

davassi:
CORS preflight channel did not succeed


paulhandy:
Closing because stale.

PR #60, opened 2017-01-26
Update TipsManager.java

Come-from-Beyond:
Artificial latency


paulhandy:
Closing because stale.

PR #62, opened 2017-02-01
New Tip Selection, reduce mutual locking.

alfredch:
The new tip selection controls the depth of the branch that a tip may have at maximum.

The reduced locking strategy uses memory maps instead of sharing the same file-system mapped Bytebuffers with other threads.

The aim of the game:

- Better tip selection will make a better tangle structure and increase the stability for txs getting confirmed.

- Reduced mutual thread locking will allow a node to stay synced.


alfredch:
Dev-release to be tested by spammers.

PR #69, opened 2017-02-10
IXI Fix for IRI call arguments & script API response

paulhandy:
A script can now properly create an arbitrary API response by calling IXIResponse.create(<some object>), and the folder that will contain all ixi is set by `-x path/to/ixi/dir` or `--ixi-dir path/to/ixi/dir`.


paulhandy:
Merged into #70

PR #70, opened 2017-02-14
Add IXI / MAM

paulhandy:
Usage:
```
String plainText = "SOMETRINARYSTRING9";
String key = "TRINARYKEYSOME9";
String cipherText = ISC.encrypt(key).apply(plainText);
plainText == ISC.decrypt(key).apply(cipherText);
```


paulhandy:
Should there be any specific behaviour when `key.length()  text.length()`, besides to panic? Currently, it only checks that the key length is at least text length, and truncates the key. Should it fill past the plainText.length() with random data instead?

paulhandy:
merged #69 into this branch

PR #97, opened 2017-05-02
Extracted System.exit() from printUsage()

vatcher:
Otherwise method should be called printUsageAndExit. But the major reason is, that System.exit(2) (some lines before) would be never reached.


paulhandy:
Please check branch `dev` for this, and re-submit pull request there if necessary.

PR #180, opened 2017-06-25
Feature/unused code

janprill:
This PR marks API methods as @SuppressWarnings("unused") before it deleted multiple methods which at least seem to be unused.

I'm in a situation where I try to getter a deeper understanding of your superinteresting project. IMO it is important for people in the same situation like myself to don't get distracted from code not any longer used in the project. That said it might very well be, that methods I've deleted are still important through API calls, zmq-calls and the like. All I can currently say is that the code still compiles, tests are still without failures and the behaviour in the --testnet seems to be as it has been before my changes.


paulhandy:
Some of the unused code is exposed for the purpose of IXI modules, by the way.

paulhandy:
Also, closing this since pull requests to master are not accepted.
Feel free to re-open a pull request to `dev` branch.

janprill:
Fair enough, might well resubmit to dev after I've got a thorough understanding of IXI. Cheers, Jan

janprill:
IMO you should put comments on methods exposed as an API nevertheless.

paulhandy:
that's for sure :)

Issue #211, opened 2017-07-04
IXI plugin loader behavior inconsistent


bluedigits:
## Issue 1
1. If you have an IXI module inside of the `iri` folder it get's loaded -> OK
2. Move this folder out of the `iri`folder -> Nothing happens
3. Move it back inside of the `iri`folder -> Script get's loaded again

## Issue 2
1. Delete `index.js` of a deployed IXI module -> Log shows something similar to:
```2017-07-04 15:35:07 [Thread-10] DEBUG com.iota.iri.IXI - detach child: ixi/nstats/index.js```
2. Try to request the API with a command introduced by this particular IXI module (e.g. ```nstats.getHealth```) -> Still works

## Issue 3
Although #182 has been fixed there is another double loading issue which only occurs on IRI startup on Linux systems due to the fact, that Java Files.list(dir) returns a list of files with different file sortings on Windows/Linux. On Windows I have no double loading issue anymore. There might be a little flaw at L69 in `IRI.java`.

Proposal: Revamp the loading mechanics by improving the WatchService/WatchKey polling implementation

nicoschtein:
Related to #182 maybe ?

bluedigits:
Nope, that one is fixed. You can check by yourself: Edit index.js while it's loaded already -> it get's reloaded once, but was twice before

PR #219, opened 2017-07-08
Bugfix/ixi loader

bluedigits:
Revamped IXI plugin loader


bluedigits:
This should solve problems described in #211

PR #229, opened 2017-07-14
usage: port is mandatory

robertkowalski:
fix notation for users that rely on the info output.


paulhandy:
Please submit pull requests to `dev` branch

PR #230, opened 2017-07-14
usage: port is mandatory

robertkowalski:
fix notation for users that rely on the info output.

old PR: https://github.com/iotaledger/iri/pull/229


jakubcech:
Will be fixed in #724

PR #244, opened 2017-07-29
Only publish new stored tx to ZMQ

viossat:
Prevent from publishing old already stored transactions to the "tx" ZMQ topic.


alfredch:
This change would expose another bug in IRI: Transactions stored locally via API service call are not explicitly published by zmq. There are only published when the echo broadcast arrives from a peer node. With your pull request, locally stored txs will not be published at all.
I suggest moving the publishing of txs into the  persistenc provider. I will submit a tested PR for that change.

PR #246, opened 2017-07-31
Move zmq publish operation for tx items to a new ZmqPublishProvider.

alfredch:
Previously, the txs stored by API serivce where not published directly but only on arrival of a broadcast echo from a peer node.


alfredch:
Exception in stresstest, fix needs revising.

PR #255, opened 2017-08-08
Added Height column, and Tip search by height

paulhandy:


jakubcech:
@paulhandy can you please describe the motivations for this as a new issue? This is now outdated but it could be a useful thing to do if we define the motivations for it.

PR #270, opened 2017-08-16
Update README.md (master branch)

rajivshah3:
Fix config in INI example because it may mislead users that they can set the API port as the same as the UDP port


capaj:
I have it set up like that but I don't see any errors. Is that normal behaviour?

PR #272, opened 2017-08-19
Fix IXI command selection

viossat:
If the name of an IXI module is longer that the entire name of a called IXI command, `String.substring` raises `IndexOutOfBoundsException`.

Steps to reproduce:
- create module `LongNameModule` (must appears before the other module in `ixiAPI`)
- create module `Foo` with the command `bar`
- call `Foo.bar`

Note that this PR also makes the dot between the module name and the sub-command mandatory (calling `Module.subCommand` or `ModuleXsubCommand` had the same effect).


paulhandy:
Great! I'll have a look at this soon

GalRogozinski:
Seems to be merged

PR #281, opened 2017-08-30
adds recommendation for number of neighbors

capaj:
also showcase how to define multiple neighbors in the ini file


jakubcech:
Going to close this as it's been open for too long. We have multiple neighbors in the readme now and we have recommendations for the number of neighbors in the documentation.

PR #301, opened 2017-09-16
replace java alpine to openjdk-jre version

kelu27:
alpine version didn't work
```Exception in thread "main" java.lang.UnsatisfiedLinkError: /tmp/librocksdbjni2898419474021443955.so: Error loading shared library libstdc++.so.6: No such file or directory (needed by /tmp/librocksdbjni2898419474021443955.so)```


paulhandy:
Ah, this is a musl issue?

PR #303, opened 2017-09-16
Feature: use nonce to add attachment timestamp.

alon-e:
```    
NONCE_TRINARY_SIZE = 243;
```
will be changed to
```    
TAG_TRINARY_SIZE = 81
ATTACHMENT_TIMESTAMP_TRINARY_SIZE = 27
ATTACHMENT_TIMESTAMP_LOWER_BOUND_TRINARY_SIZE = 27
ATTACHMENT_TIMESTAMP_UPPER_BOUND_TRINARY_SIZE = 27
NONCE_TRINARY_SIZE = 81
```
to allow specifying a timestamp for each attached transaction (so re-attachments will have a different attachment timestamp).
also the tag has been moved out to the attachment section, to allow a different tag for each re-attachment. 

in this PR, the tag is the same as obsolete tag.


alon-e:
yes, currently the value of (attachment) TAG will be the same as (bundle) TAG.
as this is what is impl. in IRI's `attachToTangle` API call.

however, this opens the option to assign a different TAG to a single instance of the attached bundle in the future.

PR #305, opened 2017-09-17
Curl: differentiated P81 with P27

paulhandy:


src/main/java/com/iota/iri/hash/SpongeFactory.java
```diff
@@ -5,15 +5,17 @@
...
-    public static Curl create(Mode mode){
+    public static Sponge create(Mode mode){
         switch (mode) {
-            case CURL: return new Curl();
+            case CURLP81: return new Curl(mode);
+            case CURLP27: return new Curl(mode);
             case KERL: return new Kerl();
-            case BCURLT: return new Curl(true);
+            //case BCURLT: return new Curl(true, mode);
             default: return null;
```
th0br0: this is just a nit, but shouldn't we throw a `RuntimeException` here?

PR #314, opened 2017-09-26
Make SpongeFactory fail fast

eukaryote31:
This may help weed out future NPE or unaccounted cases.


paulhandy:
Not sure that I agree with this approach. What merits do you see of this approach vs defaulting to, i,e. CurlP81?

eukaryote31:
The exception would only be thrown when passed either null or an enum value with no case, both of which are most likely a result of human error. Imagine getting the wrong sponge type and pulling your hair out over why your hashes are wrong.

paulhandy:
I suppose that I agree, then. I'll approve this.

PR #316, opened 2017-10-10
Added support for supplemental REMOTE option

scott79:
Patch for 'REMOTE' configuration file option.

REMOTE = true 

Setting this configuration option is equivalent to setting "API_HOST" or adding cli option "--remote".

### DIFF
--- a/src/main/java/com/iota/iri/IRI.java
+++ b/src/main/java/com/iota/iri/IRI.java
@@ -182,7 +182,8 @@ public class IRI {
             configuration.put(DefaultConfSettings.TCP_RECEIVER_PORT, vrporttcp);
         }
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/iota/iri/IRI.java
```diff
@@ -182,7 +182,8 @@ private static void validateParams(final Configuration configuration, final Stri
             configuration.put(DefaultConfSettings.TCP_RECEIVER_PORT, vrporttcp);
         }
 
-        if (parser.getOptionValue(remote) != null) {
+	final Boolean remoteBindingOption = configuration.booling(DefaultConfSettings.REMOTE);
```
GalRogozinski: final is redundant

jakubcech:
Fixed as part of #724

PR #317, opened 2017-10-10
Fixed neighbor list if clients are using dynamic dns service.

scott79:
- Added log stmt for ip change
 - Ensure existing neighbor 'proto://<address>:<port>' is removed from list
 - Added neighbor as 'proto://<hostname>:<port>'


marcel-burkhard:
Nice 😃  
If you want, we can test it some time 👍

GalRogozinski:
Fix conflicts please

alexsporn:
I am having this issue with my node. No way to remove the duplicate neighbors with an API call

PR #322, opened 2017-10-14
Handled address NullPointerException thrown

scott79:
Corrected reported exception thrown for invalid host resolution.


GalRogozinski:
Outdated

PR #325, opened 2017-10-17
Optimize neighbor operations

pRizz:
This changes the neighbor data structure to be a `HashSet` instead of an `ArrayList`. This optimizes operations like `add`, `remove`, `contains`, etc. This PR also abstracts how neighbors are stored in the `Node` class as well as public views of it, `getNeighbors()`.

### DIFF
--- a/src/main/java/com/iota/iri/network/Node.java
+++ b/src/main/java/com/iota/iri/network/Node.java
@@ -6,14 +6,16 @@ import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.security.SecureRandom;
 import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
-import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.concurrent.ConcurrentSkipListSet;
-import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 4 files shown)


pRizz:
I know this is anecdotal and there could be other factors, but the average CPU on my node dropped from 15-20% to 4-6% with this change

I will do some more testing and make sure this is stable.

pRizz:
Just fyi, I think the 4-6% was a fluke. Either way, this is more efficient in theory.

alon-e:
I would argue that most common operation on `neighbors` is actually iterations:
[Node.java#L241](https://github.com/iotaledger/iri/blob/dev/src/main/java/com/iota/iri/network/Node.java#L241)
[Node.java#L520](https://github.com/iotaledger/iri/blob/dev/src/main/java/com/iota/iri/network/Node.java#L520)
[Node.java#L551](https://github.com/iotaledger/iri/blob/dev/src/main/java/com/iota/iri/network/Node.java#L551)
(by order)

so, while `add, remove, contains` are optimized by a `HashSet`, I think that an `ArrayList` is still faster for the common case here which is iterating. [stackoverflow's thoughts](https://stackoverflow.com/questions/10799417/performance-and-memory-allocation-comparision-between-list-and-set/10799668#10799668)

OTOH, [Node.java#L241](https://github.com/iotaledger/iri/blob/dev/src/main/java/com/iota/iri/network/Node.java#L241) logic can be changed to `contains()`. 
but I would still argue that L520 & L551 are more common than the ops above.

any ideas on how to optimize them as well?

pRizz:
Those are good points. And there are iterations of those neighbors outside Node.java as well, so those might hurt too. I can revert the Set-List change, but can we keep the type abstraction change? We could revisit optimizing later.

paulhandy:
Good call, @alon-e

pRizz:
Will revisit this later

PR #330, opened 2017-10-24
Add X-IOTA-API-Version to CORS allowed headers

### LINKED ISSUE #329: Header issues in IRI 1.4.1
The latest IRI (1.4.1) requires the http header X-IOTA-API-Version to be set. But IRI does not report that header as a valid header when the browsers performs a CORS prefligh check:

This is that IRI reports as valid:
```
Access-Control-Allow-Headers:Origin, X-Requested-With, Content-Type, Accept
Access-Control-Allow-Origin:*
```

Note the absence of the X-IOTA-API-Version header. Until this is fixed we can't use say fetch() to do requests to IRI.

th0br0:
Fixes #329

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,14 +8,6 @@ sudo: required
 #services:
 #- docker 000ctGQcXVGUjeURqQeHclLLpst2Td0uR6SgAGLEOxjeQ9ghzQWC/4x9BPW7xf7SkxhESWB2MMTvxma8xaV7hmNSsUK7tHyMLZPGdhDnPuUx2E6eNG4IZzUvcmHI4U=
 
-before_install:
-- openssl aes-256-cbc -K $encrypted_5a15fa813cca_key -iv $encrypted_5a15fa813cca_iv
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -8,9 +8,6 @@ sudo: required
 #services:
 #- docker 000ctGQcXVGUjeURqQeHclLLpst2Td0uR6SgAGLEOxjeQ9ghzQWC/4x9BPW7xf7SkxhESWB2MMTvxma8xaV7hmNSsUK7tHyMLZPGdhDnPuUx2E6eNG4IZzUvcmHI4U=
 
-before_install:
-- openssl aes-256-cbc -K $encrypted_5a15fa813cca_key -iv $encrypted_5a15fa813cca_iv
-  -in codesigning.asc.enc -out codesigning.asc -d
 
 before_script: gpg --fast-import codesigning.asc
```
mlouielu: Should this be `codesigning.asc.enc`?

PR #339, opened 2017-11-06
refactoring iota.rescan_db

magussiro:
refactoring iota.rescan_db


GalRogozinski:
It seems that the changes are merged

PR #340, opened 2017-11-06
Use new tag field to index transactions

viossat:
The `obsoleteTag` field is still used to index transactions. The result of `findTransactions` may be different if `obsoleteTag` differs from `tag`.

### DIFF
--- a/src/main/java/com/iota/iri/controllers/TransactionViewModel.java
+++ b/src/main/java/com/iota/iri/controllers/TransactionViewModel.java
@@ -161,7 +161,7 @@ public class TransactionViewModel {
         hashesList.add(new Pair<>(getBundleHash(), new Bundle(hash)));
         hashesList.add(new Pair<>(getBranchTransactionHash(), new Approvee(hash)));
         hashesList.add(new Pair<>(getTrunkTransactionHash(), new Approvee(hash)));
-        hashesList.add(new Pair<>(getObsoleteTagValue(), new Tag(hash)));
+        hashesList.add(new Pair<>(getTagValue(), new Tag(hash)));
         setAttachmentData();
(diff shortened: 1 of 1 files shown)


paulhandy:
I like this. What do you think, @alon-e ?

paulhandy:
@viossat could you update this PR?

mlouielu:
Will this cause some backward compatibility with the behavior changed at #303 ?

alon-e:
>Will this cause some backward compatibility with the behavior changed at #303 ?

well the default behavior of `attachToTangle` copies the `obsoleteTag`  value to `tag` (if no attachment `tag` was specified explicitly - by pre-filling the field in sent trytes).
so I would say that it's 90% backward compatible - in cases were a special attachment tag isn't used, but also expands the ability to query cases were a special attachment tag was used 
-for example to differentiate between re-attached tx versions by incr. tags.

the case where I can see this not back. comp. is if someone uses custom PoW, that doesn't follow the changes in #303 and uses all 81 trytes as nonce - then querying by the tag wouldn't work.
but I would argue needing both is rare (custom PoW & querying by tag), and the benefits are much greater.

so I think it's a good PR as well :+1:

mlouielu:
@alon-e so be clear, if we apply this PR, that the function of the `obsoleteTag` will only used in milestone as milestone index, and others are deprecated, right?

cyclux:
@alon-e Could this PR please considered for the next release?

GalRogozinski:
Just to be clear @viossat,
This will have to wait for the next global snapshot

jakubcech:
This was fixed as part of https://github.com/iotaledger/iri/pull/728, thanks!

PR #366, opened 2017-11-27
Make examples less confusing

rajivshah3:
I changed the flag and INI examples to use the default ports as some users have been confused by it. The INI example sets API port to the same as UDP port which is not possible in reality.


geminoz:
lgtm

GalRogozinski:
#731 took care of this

PR #369, opened 2017-11-27
Cleanup in API.java

gubatron:
- Typos, "ellapsed" -> "elapsed"
- member visibility, several static methods visibility moved to private or package
- tidy, several static counters that were declared throughout the file moved to the start
- tidy, line spacing between methods
- padTag() implementation optimized to not create unnecessary String instances with += operator using StringBuilder
- Simplified getParameterAsSet(), avoid use of stream api for sake of being cool, a simple copy HashSet constructor performs better avoiding the creation of temporary objects created with .stream().collect, also simpler easier to read/maintain.
- imports optimized

Should be warning free now.

### DIFF
--- a/src/main/java/com/iota/iri/service/API.java
+++ b/src/main/java/com/iota/iri/service/API.java
@@ -1,63 +1,23 @@
 package com.iota.iri.service;
 
-import static io.undertow.Handlers.path;
-
-import java.io.IOException;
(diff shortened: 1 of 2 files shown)


jakubcech:
Closing as stale. API.java needs to be completely refactored from the current state at some point.

PR #380, opened 2017-12-03
Default configuration is a bit misleading

sharpner:
I ran into the issue that the API port conflicted with the tcp receiver port
so IMHO if the configuration is documented like this its much clearer :)


sharpner:
closing this one because https://github.com/iotaledger/iri/pull/366 is much better.

PR #392, opened 2017-12-06
Api command functional routing

yovendielmundo:
1. Use runtime exceptions and handle it all in the same place
Improve readability/maintainability. Ease create proper http responses.
2. Refactor how to validate and get an uri from strings
3. Request body errors as bad request responses
4. Use functional hash mapping for command routing instead of switch cases
Improve readability/maintainability/efficiency  of the code.

### DIFF
--- a/src/main/java/com/iota/iri/service/API.java
+++ b/src/main/java/com/iota/iri/service/API.java
@@ -375,22 +375,13 @@ public class API {
         final AtomicInteger numberOfRemovedNeighbors = new AtomicInteger(0);
 
         for (final String uriString : uris) {
-            try {
-                final URI uri = new URI(uriString);
-
-                if ("udp".equals(uri.getScheme()) || "tcp".equals(uri.getScheme())) {
-                    log.info("Removing neighbor: "+uriString);
-                    if (instance.node.removeNeighbor(uri,true)) {
-                        numberOfRemovedNeighbors.incrementAndGet();
-                    }
-                }
-                else {
-                    return ErrorResponse.create("Invalid uri scheme");
-                }
-            } catch (URISyntaxException e) {
-                return ErrorResponse.create("Invalid uri syntax");
+            final URI uri = getValidUri(uriString);
+            log.info("Removing neighbor: "+uriString);
+            if (instance.node.removeNeighbor(uri,true)) {
+                numberOfRemovedNeighbors.incrementAndGet();
             }
         }
+
         return RemoveNeighborsResponse.create(numberOfRemovedNeighbors.get());
     }
 
@@ -810,38 +801,25 @@ public class API {
     }
 
     private AbstractResponse addNeighborsStatement(final List<String> uris) {
-
         int numberOfAddedNeighbors = 0;
-        for (final String uriString : uris) {
-            try {
-                final URI uri = new URI(uriString);
 
-                if ("udp".equals(uri.getScheme()) || "tcp".equals(uri.getScheme())) {
-                    log.info("Adding neighbor: "+uriString);
-                    // 3rd parameter true if tcp, 4th parameter true (configured tethering)
-                    final Neighbor neighbor;
-                    switch(uri.getScheme()) {
-                        case "tcp":
-                            neighbor = new TCPNeighbor(new InetSocketAddress(uri.getHost(), uri.getPort()),true);
-                            break;
-                        case "udp":
-                            neighbor = new UDPNeighbor(new InetSocketAddress(uri.getHost(), uri.getPort()), instance.node.getUdpSocket(), true);
-                            break;
-                        default:
-                            return ErrorResponse.create("Invalid uri scheme");
-                    }
-                    if (!instance.node.getNeighbors().contains(neighbor)) {
-                        instance.node.getNeighbors().add(neighbor);
-                        numberOfAddedNeighbors++;
-                    }
-                }
-                else {
-                    return ErrorResponse.create("Invalid uri scheme");
-                }
-            } catch (URISyntaxException e) {
-                return ErrorResponse.create("Invalid uri syntax");
+        for (final String uriString : uris) {
+            final URI uri = getValidUri(uriString);
+            log.info("Adding neighbor: "+uriString);
+            // 3rd parameter true if tcp, 4th parameter true (configured tethering)
+            final Neighbor neighbor;
+            if ("tcp".equals(uri.getScheme())) {
+                neighbor = new TCPNeighbor(new InetSocketAddress(uri.getHost(), uri.getPort()), true);
+            } else {
+                neighbor = new UDPNeighbor(new InetSocketAddress(uri.getHost(), uri.getPort()), instance.node.getUdpSocket(), true);
+            }
+
+            if (!instance.node.getNeighbors().contains(neighbor)) {
+                instance.node.getNeighbors().add(neighbor);
+                numberOfAddedNeighbors++;
             }
         }
+
         return AddedNeighborsResponse.create(numberOfAddedNeighbors);
     }
 
@@ -968,6 +946,26 @@ public class API {
             return TagViewModel.load(tangle, indexable);
         } catch (Exception e) {
             throw new IllegalStateException(e.getMessage());
+        }
+    }
+
+    /**
+     * Gets a valid uri{@link URI} from an uri String
+     *
+     * @param uriString uri String
+     * @return valid uri
+     * @throws IllegalArgumentException if scheme is not udp or tcp
+     * @throws IllegalArgumentException if uriString has invalid sintax
+     */
+    private URI getValidUri(final String uriString) {
+        try {
+            final URI uri = new URI(uriString);
+            if (!"udp".equals(uri.getScheme()) && !"tcp".equals(uri.getScheme())) {
+                throw new IllegalArgumentException("Invalid uri scheme");
+            }
+            return uri;
+        } catch (URISyntaxException e) {
+            throw new IllegalArgumentException("Invalid uri syntax");
         }
     }
 }
--- a/src/main/java/com/iota/iri/service/API.java
+++ b/src/main/java/com/iota/iri/service/API.java
@@ -27,6 +27,8 @@ import java.util.stream.Collectors;
 import com.iota.iri.*;
 import com.iota.iri.controllers.*;
 import com.iota.iri.network.*;
+import com.iota.iri.storage.Indexable;
+import com.iota.iri.storage.Tangle;
 import org.apache.commons.io.IOUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -75,8 +77,6 @@ import io.undertow.util.HttpString;
 import io.undertow.util.Methods;
 import io.undertow.util.MimeMappings;
 import io.undertow.util.StatusCodes;
-
-import javax.xml.bind.ValidationException;
 
 @SuppressWarnings("unchecked")
 public class API {
@@ -286,11 +286,12 @@ public class API {
 
                 case "storeTransactions": {
                     try {
-                        storeTransactionStatement(getParameterAsList(request,"trytes", TRYTES_SIZE));
-                    } catch (RuntimeException e) {
+                        storeTransactionStatement(getParameterAsList(request, "trytes", TRYTES_SIZE));
+                    } catch (Exception e) {
                         //transaction not valid
                         return ErrorResponse.create("Invalid trytes input");
                     }
+                    return AbstractResponse.createEmptyResponse();
                 }
                 case "getMissingTransactions": {
                     //TransactionRequester.instance().rescanTransactionsToRequest();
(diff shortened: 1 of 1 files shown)


src/main/java/com/iota/iri/service/API.java
```diff
@@ -804,35 +718,26 @@ public static void incEllapsedTime_PoW(long ellapsedTime) {
...
-                    numberOfAddedNeighbors++;
-                }
+        for (final String uriString : uris) {
+            final URI uri = getValidUri(uriString);
+            log.info("Adding neighbor: "+uriString);
+            // 3rd parameter true if tcp, 4th parameter true (configured tethering)
+            final Neighbor neighbor;
+            if ("tcp".equals(uri.getScheme())) {
+                neighbor = new TCPNeighbor(new InetSocketAddress(uri.getHost(), uri.getPort()), true);
+            } else {
```
gubatron: probably not a good assumption to assume the else case is going to be "udp", better be specific.

src/main/java/com/iota/iri/service/API.java
```diff
@@ -804,35 +718,26 @@ public static void incEllapsedTime_PoW(long ellapsedTime) {
...
+                neighbor = new UDPNeighbor(new InetSocketAddress(uri.getHost(), uri.getPort()), instance.node.getUdpSocket(), true);
             }
-            else {
```
gubatron: this else should then come back, because then the rest of the logic doesn't make sense if an invalid uri scheme is passed and you're assuming UDP

src/main/java/com/iota/iri/service/API.java
```diff
@@ -912,4 +817,218 @@ public void shutDown() {
             server.stop();
         }
     }
+
+    // TransactionViewModel, ApproveeViewModel, AddressViewModel, BundleViewModel, TagViewModel Interface
+    private TransactionViewModel transactionViewModelFromHash(final Tangle tangle, final Hash hash) {
```
gubatron: this wrapper seems unnecessary, unnecessary symbols.

src/main/java/com/iota/iri/service/API.java
```diff
@@ -700,7 +614,7 @@ private AbstractResponse getBalancesStatement(final List<String> addrss, final i
 
             if (analyzedTips.add(hash)) {
 
-                final TransactionViewModel transactionViewModel = TransactionViewModel.fromHash(instance.tangle, hash);
+                final TransactionViewModel transactionViewModel = transactionViewModelFromHash(instance.tangle, hash);
```
gubatron: unnecessary refactor

src/main/java/com/iota/iri/service/API.java
```diff
@@ -912,4 +817,218 @@ public void shutDown() {
...
+     * @param uriString uri String
+     * @return valid uri
+     * @throws IllegalArgumentException if scheme is not udp or tcp
+     * @throws IllegalArgumentException if uriString has invalid sintax
+     */
+    private URI getValidUri(final String uriString) {
+        try {
+            final URI uri = new URI(uriString);
+            if (!"udp".equals(uri.getScheme()) && !"tcp".equals(uri.getScheme())) {
+                throw new IllegalArgumentException("Invalid uri scheme");
```
gubatron: oh I see.

src/main/java/com/iota/iri/service/API.java
```diff
@@ -804,35 +718,26 @@ public static void incEllapsedTime_PoW(long ellapsedTime) {
...
     }
 
-    private AbstractResponse addNeighborsStatement(final List<String> uris) throws URISyntaxException {
-
+    private AbstractResponse addNeighborsStatement(final List<String> uris) {
         int numberOfAddedNeighbors = 0;
-        for (final String uriString : uris) {
-            final URI uri = new URI(uriString);
 
-            if ("udp".equals(uri.getScheme()) || "tcp".equals(uri.getScheme())) {
```
gubatron: it seems as if this part of the refactor got a little out of the scope of the Pull Request's intention.
But I see how you factored `if ("udp".equals(uri.getScheme()) || "tcp".equals(uri.getScheme())) {` into a function to avoid code repetition.

src/main/java/com/iota/iri/service/API.java
gubatron: nvm, I see how `getValidUri` would throw an exception

src/main/java/com/iota/iri/service/API.java
```diff
@@ -162,25 +181,28 @@ private void processRequest(final HttpServerExchange exchange) throws IOExceptio
 
         final long beginningTime = System.currentTimeMillis();
         final String body = IOUtils.toString(cis, StandardCharsets.UTF_8);
-        final AbstractResponse response;
 
+        final AbstractResponse response;
         if (!exchange.getRequestHeaders().contains("X-IOTA-API-Version")) {
             response = ErrorResponse.create("Invalid API Version");
-        } else if (body.length() > maxBodyLength) {
```
gubatron: why was this check removed?

gubatron: seems like a nice opening for a big payload attack.

src/main/java/com/iota/iri/service/API.java
```diff
@@ -162,25 +181,28 @@ private void processRequest(final HttpServerExchange exchange) throws IOExceptio
...
+
         sendResponse(exchange, response, beginningTime);
     }
 
-    private AbstractResponse process(final String requestString, InetSocketAddress sourceAddress) throws UnsupportedEncodingException {
+    private AbstractResponse process(final String requestString, InetSocketAddress sourceAddress) {
 
         try {
 
+            if (requestString.length() > maxBodyLength) {
```
gubatron: 👍

src/main/java/com/iota/iri/service/API.java
```diff
@@ -912,4 +817,218 @@ public void shutDown() {
...
+
+    private boolean transactionViewModelExist(final Tangle tangle, final Hash hash) {
+        try {
+            return TransactionViewModel.exists(tangle, hash);
+        } catch (Exception e) {
+            throw new IllegalStateException(e.getMessage());
+        }
+    }
+
+    private AddressViewModel addressViewModelLoad(final Tangle tangle, final Indexable indexable) {
```
gubatron: still not sure what you're trying to achieve with these local wrappers, this only creates a bunch of unnecessary symbols

src/main/java/com/iota/iri/service/API.java
```diff
@@ -912,4 +817,218 @@ public void shutDown() {
...
+        try {
+            return AddressViewModel.load(tangle, indexable);
+        } catch (Exception e) {
+            throw new IllegalStateException(e.getMessage());
+        }
+    }
+
+    private BundleViewModel bundleViewModelLoad(final Tangle tangle, final Indexable indexable) {
```
gubatron: 👎

src/main/java/com/iota/iri/service/API.java
```diff
@@ -912,4 +817,218 @@ public void shutDown() {
...
+        try {
+            return BundleViewModel.load(tangle, indexable);
+        } catch (Exception e) {
+            throw new IllegalStateException(e.getMessage());
+        }
+    }
+
+    private TagViewModel tagViewModelLoad(final Tangle tangle, final Indexable indexable) {
```
gubatron: :-1

src/main/java/com/iota/iri/service/API.java
```diff
@@ -912,4 +817,218 @@ public void shutDown() {
...
+
+    /**
+     * Gets a valid uri{@link URI} from an uri String
+     *
+     * @param uriString uri String
+     * @return valid uri
+     * @throws IllegalArgumentException if scheme is not udp or tcp
+     * @throws IllegalArgumentException if uriString has invalid sintax
+     */
+    private URI getValidUri(final String uriString) {
```
gubatron: :+1

src/main/java/com/iota/iri/service/API.java
gubatron: is it about converting the exceptions (whatever they might be) into `IllegalStateException`s?
smells funky.

src/main/java/com/iota/iri/service/API.java
```diff
@@ -120,6 +122,23 @@ public API(Iota instance, IXI ixi) {
...
+        commandRoute = new HashMap<>();
+        commandRoute.put("addNeighbors", addNeighbors());
+        commandRoute.put("attachToTangle", attachToTangle());
+        commandRoute.put("broadcastTransactions", broadcastTransactions());
+        commandRoute.put("findTransactions", findTransactions());
+        commandRoute.put("getBalances", getBalances());
+        commandRoute.put("getInclusionStates", getInclusionStates());
+        commandRoute.put("getNeighbors", getNeighbors());
+        commandRoute.put("getNodeInfo", getNodeInfo());
+        commandRoute.put("getTips", getTips());
```
gubatron: why not just:
`commandRoute.put("getTips", getTipsStatement());`

and save yourself the unnecessary wrapper
`private Function<Map<String, Object>, AbstractResponse> getTips() {`

src/main/java/com/iota/iri/service/API.java
```diff
@@ -912,4 +817,218 @@ public void shutDown() {
...
+                        .create("This operations cannot be executed: The subtangle has not been updated yet.");
+            }
+            final List<String> transactions = getParameterAsList(request, "transactions", HASH_SIZE);
+            final List<String> tips = getParameterAsList(request, "tips", HASH_SIZE);
+
+            return getNewInclusionStateStatement(transactions, tips);
+        };
+    }
+
+    private Function<Map<String, Object>, AbstractResponse> getNeighbors() {
```
gubatron: unnecessary wrapper

src/main/java/com/iota/iri/service/API.java
```diff
@@ -912,4 +817,218 @@ public void shutDown() {
...
+                    Runtime.getRuntime().freeMemory(), System.getProperty("java.version"), Runtime.getRuntime().maxMemory(),
+                    Runtime.getRuntime().totalMemory(), instance.milestone.latestMilestone, instance.milestone.latestMilestoneIndex,
+                    instance.milestone.latestSolidSubtangleMilestone, instance.milestone.latestSolidSubtangleMilestoneIndex,
+                    instance.node.howManyNeighbors(), instance.node.queuedTransactionsSize(),
+                    System.currentTimeMillis(), instance.tipsViewModel.size(),
+                    instance.transactionRequester.numberOfTransactionsToRequest());
+        };
+    }
+
+    private Function<Map<String, Object>, AbstractResponse> getTips() {
```
gubatron: unnecessary wrapper

src/main/java/com/iota/iri/service/API.java
yovendielmundo: This funky approach is nice here. Each request needs a response, thats for sure, with 2xx, 3xx, 4xx, 5xx status code. We can just rise an exception, wherever in the code (IllegalArgumentException for errors in the input, IllegalStateExceptions for internal errors) and handle all exceptions in just one place. The idea is: We don't need to propagate all compile time exceptions in every single function because we know where to handle all the errors and what to do with them (log the error and send the response). This wraper could represents a thin layer between the API and the other components. it could be in another file.

src/main/java/com/iota/iri/service/API.java
yovendielmundo: This is an error in the input data. We just have to rise the `IllegalArgumentException` and then send a bad request response.

src/main/java/com/iota/iri/service/API.java
yovendielmundo: I agree with memory problems in particular devices, specially with java applications. In this case I'm not sure if memory has increased with splitting this commands into functions but I'll check. Anyway I'll have this in mind for the future.

yovendielmundo:
Thank you @gubatron for the review. I hope something can be use.

src/main/java/com/iota/iri/service/API.java
yovendielmundo: You must check `getValidUri` to understand this assumption

GalRogozinski:
I was considering using jax-rs. However, since your solution may also be satisfactory it may be accepted when we come around to refactor the API.

brunoamancio:
I'm coming from https://github.com/iotaledger/iri/pull/743, where the command pattern was cited.

In that PR the Codacy bot is suggesting lowering the complexity of the method process(...) in the API.

In order to have better structure with less complexity, the logic should be something like it was commit in this issue, for precisely https://github.com/iotaledger/iri/pull/392/commits/53c519991f114ca613e3f03a5545d200d842f9cc

GalRogozinski:
Hey @yovendielmundo,

Will you be willing to do an implementation of the API using JAX-RS with RESTEASY?
I liked your use of the commander pattern but I think that for the longer term, using JAX-RS will be better.

If you agree, 
here is a link to a quick tutorial to get started: https://github.com/vdevigere/undertow-cdi-jaxrs/wiki.

brunoamancio:
@GalRogozinski What if we implement the basic functionality he committed in https://github.com/iotaledger/iri/commit/53c519991f114ca613e3f03a5545d200d842f9cc so that the current structure becomes more malleable faster and then go on to using that tool?

GalRogozinski:
That makes good sense actually.
Go for it.

I'll keep an eye on this PR.

PR #404, opened 2017-12-15
Working docker image, new configuration system, updated docs, ci pipeline

PatrickHuetter:
1. implemented new configuration system, integrated lightbend config, made it backward compatible to the old configuration api, added central default configuration as reference.conf; environment and application specific configuration can be provided by application.conf, system environment variables and java properties; refactored validation of params/config paths to do validation only mostly and updated usage info.

2. created new docker image which works fine in production environments and also rancher or kubernetes. Configurable by environment variables. To use the docker image to run an iri node it's enough to provide only one variable "NEIGHBORS". The rest is already configured by best practice and default configuration. So it's easy to set up a working docker iri node within seconds.

3. Integrated gitlab-ci to have automatically built docker images and also free docker image registry hosting.


src/main/java/com/iota/iri/IRI.java
```diff
@@ -93,158 +77,63 @@ public static void main(final String[] args) throws IOException {
...
-                "[{-u,--udp-receiver-port} 14600] " +
-                "[{-t,--tcp-receiver-port} 15600] " +
-                "[{-d,--debug} false] " +
-                "[{--testnet} false]" +
-                "[{--remote} false]" +
-                "[{--remote-auth} string]" +
-                "[{--remote-limit-api} string]"
+                        "[{-DNEIGHBOURS} '<list of neighbors>'] " +
+                        "[{-DPORT} 14600] " +
+                        "[{-Dconfig.file} 'config-file-name'] " +
```
ixidion: config changed to config.file
Why?

ixidion:
Can you explain why a new configuration is needed and what the motivation is for a new one. This approach may frustrate many people, because they expect the API to be stable. 
Why should this tradeoff be made?

PatrickHuetter:
@ixidion We'll make it backward compatible to be stable with the current cli param style and extend this PR soon. The new configuration is needed because it's much more clean and has support for different environments (inheritance, dev, prod, staging and so on) and configuration methods (file, environment also with inheritance). It's also needed to run iri in scale on different cloud providers using docker for example and also support clean catalog-files and provisioning configurations (ansible, rancher-compose and so on).

GalRogozinski:
Thanks for your PR. After inspecting that PR we decided to go for a different solution utilizing JCommander, See #724.

We decided it is better to use only one framework for configuration and it should be one that helps us with putting params via CLI and generating docs.

All docker suggestions have been reviewed by @geminoz. The new docker file will be pushed as part of #744.

PR #405, opened 2017-12-15
Closing of resources and Thread interrupt

marcuslang:
Hi,

I'm a huge fan of your project and wanted to help a little bit. 
While going through a code analysis I found two things that can possibly end in a bug or make it harder to analyse bugs. First I've added closing of resources where it was missing. Second I've reinterrupted the Threads after an InterruptedException is thrown. See [Don't swallow interrupts](https://www.ibm.com/developerworks/java/library/j-jtp05236/index.html?ca=drs-#2.1) for more details on that.

Hope that it helps. Best wishes,
Marcus

### DIFF
--- a/src/main/java/com/iota/iri/IXI.java
+++ b/src/main/java/com/iota/iri/IXI.java
@@ -2,7 +2,6 @@ package com.iota.iri;
 
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
-import com.iota.iri.hash.ISS;
 import com.iota.iri.service.CallableRequest;
 import com.iota.iri.service.dto.AbstractResponse;
 import com.sun.nio.file.SensitivityWatchEventModifier;
@@ -20,7 +19,10 @@ import java.io.Reader;
 import java.nio.file.*;
 import java.nio.file.attribute.BasicFileAttributes;
 import java.time.Instant;
-import java.util.*;
+import java.util.EnumSet;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -87,6 +89,7 @@ public class IXI {
                 key = watcher.poll(1000, TimeUnit.MILLISECONDS);
             } catch (InterruptedException e) {
                 log.error("Watcher interrupted: ", e);
+                Thread.currentThread().interrupt();
             }
             if (key == null) {
                 continue;
--- a/src/main/java/com/iota/iri/storage/FileExportProvider.java
+++ b/src/main/java/com/iota/iri/storage/FileExportProvider.java
@@ -1,7 +1,5 @@
 package com.iota.iri.storage;
 
-import com.iota.iri.controllers.TransactionViewModel;
-import com.iota.iri.model.Hash;
 import com.iota.iri.model.Transaction;
 import com.iota.iri.utils.Converter;
 import com.iota.iri.utils.Pair;
@@ -14,7 +12,6 @@ import java.io.UnsupportedEncodingException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 
 import static com.iota.iri.controllers.TransactionViewModel.TRINARY_SIZE;
@@ -53,17 +50,17 @@ public class FileExportProvider implements PersistenceProvider {
     @Override
     public boolean update(Persistable model, Indexable index, String item) throws Exception {
 
-        if(model instanceof Transaction) {
+        if (model instanceof Transaction) {
             Transaction transaction = ((Transaction) model);
-            if(item.contains("sender")) {
+            if (item.contains("sender")) {
+                PrintWriter writer = null;
                 try {
-                    PrintWriter writer;
                     Path path = Paths.get("export", String.valueOf(getFileNumber()) + ".tx");
                     writer = new PrintWriter(path.toString(), "UTF-8");
                     writer.println(index.toString());
                     writer.println(Converter.trytes(trits(transaction)));
                     writer.println(transaction.sender);
-                    if(item.equals("height")) {
+                    if (item.equals("height")) {
                         writer.println("Height: " + String.valueOf(transaction.height));
                     } else {
                         writer.println("Height: ");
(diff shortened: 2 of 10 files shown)


src/main/java/com/iota/iri/IXI.java
```diff
@@ -87,6 +89,7 @@ private void processWatchEvents() {
                 key = watcher.poll(1000, TimeUnit.MILLISECONDS);
             } catch (InterruptedException e) {
                 log.error("Watcher interrupted: ", e);
+                Thread.currentThread().interrupt();
```
gubatron: 👎 
I believe this wasn't the intention originally.
They clearly wanted to not fail. And now you'll probably cause yet another `InterruptedException` by forcing this thread to not die gracefully.

gubatron:
👎

src/main/java/com/iota/iri/storage/FileExportProvider.java
```diff
@@ -75,7 +72,9 @@ public boolean update(Persistable model, Indexable index, String item) throws Ex
                 } catch (Exception e) {
                     log.error("Transaction load failed. ", e);
                 } finally {
-
+                    if (writer != null) {
+                        writer.close();
```
gubatron: Since this project uses Apache Commons IO, you might want to take a look at `IOUtils.closeQuietly(Closeable closeable)` instead of doing this.

This `writer.close()` could throw an exception too.

I suggest you try here:

```java
} finally {
    IOUtils.closeQuietly(writer);
}
```

src/main/java/com/iota/iri/storage/rocksDB/RocksDBPersistenceProvider.java
```diff
@@ -149,47 +149,47 @@ public boolean exists(Class<?> model, Indexable key) throws Exception {
...
-        iterator.seekToLast();
-        if(iterator.isValid()) {
-            object = (Persistable) model.newInstance();
-            index = (Indexable) indexModel.newInstance();
-            index.read(iterator.key());
-            object.read(iterator.value());
-            ColumnFamilyHandle referenceHandle = metadataReference.get().get(model);
-            if(referenceHandle != null) {
-                object.readMetadata(db.get(referenceHandle, iterator.key()));
+        try (RocksIterator iterator = db.newIterator(classTreeMap.get().get(model))) {
```
gubatron: 👍

src/main/java/com/iota/iri/storage/rocksDB/RocksDBPersistenceProvider.java
```diff
@@ -149,47 +149,47 @@ public boolean exists(Class<?> model, Indexable key) throws Exception {
...
     @Override
     public Set<Indexable> keysWithMissingReferences(Class<?> model, Class<?> other) throws Exception {
         ColumnFamilyHandle handle = classTreeMap.get().get(model);
         ColumnFamilyHandle otherHandle = classTreeMap.get().get(other);
-        RocksIterator iterator = db.newIterator(handle);
-        Set<Indexable> indexables = new HashSet<>();
-        for(iterator.seekToFirst(); iterator.isValid(); iterator.next()) {
-            if(db.get(otherHandle, iterator.key()) == null) {
-                indexables.add(new Hash(iterator.key()));
+        try (RocksIterator iterator = db.newIterator(handle)) {
```
gubatron: :+1

src/main/java/com/iota/iri/storage/rocksDB/RocksDBPersistenceProvider.java
```diff
@@ -553,24 +552,27 @@ private void fillmodelColumnHandles(List<ColumnFamilyHandle> familyHandles) thro
...
-                .collect(Collectors.toList());
-        columnFamilies.add(0, familyDescriptors.get(0));
-        List<ColumnFamilyDescriptor> missingFromDatabase = familyDescriptors.stream().filter(d -> columnFamilies.stream().filter(desc -> new String(desc.columnFamilyName()).equals(new String(d.columnFamilyName()))).toArray().length == 0).collect(Collectors.toList());
-        List<ColumnFamilyDescriptor> missingFromDescription = columnFamilies.stream().filter(d -> familyDescriptors.stream().filter(desc -> new String(desc.columnFamilyName()).equals(new String(d.columnFamilyName()))).toArray().length == 0).collect(Collectors.toList());
-        if (missingFromDatabase.size() != 0) {
-            missingFromDatabase.remove(familyDescriptors.get(0));
-            db = RocksDB.open(options, path, columnFamilies, familyHandles);
-            for (ColumnFamilyDescriptor description : missingFromDatabase) {
-                addColumnFamily(description.columnFamilyName(), db);
+        try (final Options options = new Options().setCreateIfMissing(true)) {
```
gubatron: not sure what `Options` are, but it doesn't sound like this statement will fail at all
`new Options().setCreateIfMissing(true)`

are `Option` a closeable resource?

gubatron: I see, it's an Option for RocksDB, still having a hard time looking at the class family tree and finding an implementation or extension of an option being a resource.
https://github.com/facebook/rocksdb/blob/master/java/src/main/java/org/rocksdb/Options.java


Why the `try(final Options options ...)`

src/main/java/com/iota/iri/IXI.java
marcuslang: Going with this [answer on stackoverflow](https://stackoverflow.com/a/4906814/4604830)  swallowing should only be done if the code implements a interruption policy. I don't know if that's the case here. 
Anyhow, we could leave a comment in the code if swallowing that exception was intentionally.

src/main/java/com/iota/iri/storage/rocksDB/RocksDBPersistenceProvider.java
marcuslang: The inheritance of Options shows that it's a reference to an underlying C++ object and an AutoClosable. So it's pointing on a resource somewhere else and should be closed in any case.

src/main/java/com/iota/iri/storage/FileExportProvider.java
marcuslang: thanks, I'll change that

paulhandy:
Would you please rebase your changes on top of the current dev branch?

marcuslang:
I struggled a little bit with the rebase, but it's done now

GalRogozinski:
Hey,

Sorry for the late response.
Can you break this into two PRs? One for each subject.

Thanks

PR #411, opened 2017-12-18
Created an easier to use docker container

chrootLogin:
Features:
- Everything can be configured through environment.
- Creates a dedicated "IOTA"-user inside the container.
- Uses [tini](https://github.com/krallin/tini) a small but valid init.
- Ports configurable.
- Documentation updated/added docker-compose.


romansemko:
Does it make Dockerfile only usable with environment vars? What if someone wants to continue using config file? On a currently running systems this would mean a breaking change.

PatrickHuetter:
This is already contained in #404 with new configuration system (also config file is possible).

ixidion:
@PatrickHuetter You Dockerfile has also downsides, which this one doesn't have.
Set JVM Parameter, Tini, ...

A Dockerfile with the best of both sides should be the best approach. So just telling that it is already contained in another Pullrequest is not enough, when it doesn't have the same features.

geminoz:
Thanks Simon for the PR. I like that you define a UID/GID for the IRI process, however we need to cater for instances where there are multiple IRI docker containers on the same host, each running with a different UID. Please have a look at PR-744, I am going to add that support on that PR.

Giorgio

GalRogozinski:
Closed in favor of #744

Issue #424, opened 2017-12-20
IRI 1.4.1.4: NullPointerException while reading transaction.obsoleteTag


benoslav:
```
Error updating transactions.
Dez 20 23:25:25 server java[19036]: java.lang.NullPointerException: null
Dez 20 23:25:25 server java[19036]: at com.iota.iri.model.Transaction.metadata(Transaction.java:68) ~[iri-1.4.1.4.jar:na]
Dez 20 23:25:25 server java[19036]: at com.iota.iri.storage.rocksDB.RocksDBPersistenceProvider.update(RocksDBPersistenceProvider.java:385) ~[iri-1.4.1.4.jar:na]
Dez 20 23:25:25 server java[19036]: at com.iota.iri.storage.Tangle.update(Tangle.java:92) ~[iri-1.4.1.4.jar:na]
Dez 20 23:25:25 server java[19036]: at com.iota.iri.controllers.TransactionViewModel.update(TransactionViewModel.java:124) ~[iri-1.4.1.4.jar:na]
Dez 20 23:25:25 server java[19036]: at com.iota.iri.controllers.TransactionViewModel.updateHeights(TransactionViewModel.java:424) ~[iri-1.4.1.4.jar:na]
Dez 20 23:25:25 server java[19036]: at com.iota.iri.TransactionValidator.quickSetSolid(TransactionValidator.java:236) ~[iri-1.4.1.4.jar:na]
Dez 20 23:25:25 server java[19036]: at com.iota.iri.TransactionValidator.updateStatus(TransactionValidator.java:211) ~[iri-1.4.1.4.jar:na]
Dez 20 23:25:25 server java[19036]: at com.iota.iri.network.Node.processReceivedData(Node.java:407) [iri-1.4.1.4.jar:na]
Dez 20 23:25:25 server java[19036]: at com.iota.iri.network.Node.processReceivedDataFromQueue(Node.java:369) [iri-1.4.1.4.jar:na]
Dez 20 23:25:25 server java[19036]: at com.iota.iri.network.Node.lambda$spawnProcessReceivedThread$8(Node.java:579) [iri-1.4.1.4.jar:na]
Dez 20 23:25:25 server java[19036]: at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) ~[na:1.8.0_151]
Dez 20 23:25:25 server java[19036]: at java.util.concurrent.FutureTask.run(FutureTask.java:266) ~[na:1.8.0_151]
[...]

PR #430, opened 2017-12-21
Updated TransactionViewModel.java update function

paulhandy:
Update didn't ensure to include the obsoleteTag previously. This fixes that.

### DIFF
--- a/src/main/java/com/iota/iri/controllers/TransactionViewModel.java
+++ b/src/main/java/com/iota/iri/controllers/TransactionViewModel.java
@@ -119,9 +119,8 @@ public class TransactionViewModel {
         getBundleHash();
         getTagValue();
         getObsoleteTagValue();
-        getAttachmentTimestamp();
-        getAttachmentTimestampUpperBound();
-        getAttachmentTimestampLowerBound();
+        setAttachmentData();
+        setMetadata();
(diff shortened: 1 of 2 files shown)


alon-e:
fixes #424

src/main/java/com/iota/iri/controllers/TransactionViewModel.java
```diff
@@ -118,6 +118,10 @@ public boolean update(final Tangle tangle, String item) throws Exception {
         getBranchTransactionHash();
         getBundleHash();
         getTagValue();
+        getObsoleteTagValue();
+        getAttachmentTimestamp();
```
alon-e: these are actual getters, they don't load values / parse fields.
`setAttachmentData()` should be used instead (also contains `tag`)

src/main/java/com/iota/iri/controllers/TransactionViewModel.java
```diff
@@ -118,6 +118,10 @@ public boolean update(final Tangle tangle, String item) throws Exception {
...
+        getAttachmentTimestampUpperBound();
+        getAttachmentTimestampLowerBound();
```
alon-e: if these values are needed to be pre-loaded, then what about `setMetadata()`?

PR #432, opened 2017-12-22
stop old neighbor thread if neighbor re-added (full credit to @Leibi133)

### LINKED ISSUE #345: Repeated adding/removing of neighbors brakes TCP connection
When a TCP neighbor is repeatedly added and removed via API call it will eventually brake the connection
This behavior varies, but in general after 10 - 20 removals and re-adding's it will trigger this bug.
Reestablishing of the connection is then only possible by restarting the IRI application. Some times if that doesn't help, the affected TCP neighbor also has to restart his node.
UDP neighbors are not affected by this.

alon-e:
@Leibi133 's comment formatted as a Pull Request
fixes #345

### DIFF
--- a/src/main/java/com/iota/iri/network/TCPNeighbor.java
+++ b/src/main/java/com/iota/iri/network/TCPNeighbor.java
@@ -20,6 +20,7 @@ public class TCPNeighbor extends Neighbor {
     private int tcpPort;
 
     private final ArrayBlockingQueue<ByteBuffer> sendQueue = new ArrayBlockingQueue<>(10);
+    private boolean stopped = false;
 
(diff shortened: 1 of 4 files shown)


alon-e:
what if neighbors are replaced with other neighbors,
this fix won't resolve that.

1. while `removeNeighbor()`  closes the (Source & Sink) sockets, the thread keeps on running (same as `while` loop as touched here)

2. the HashMap keeps references to dead threads. so memory would not be freed by GC (?).

bahamapascal:
Have made a test run over the last 24 hours and it's working great and smooth!

alon-e:
merged but didn't close automatically.

Issue #413, opened 2017-12-18
Option to disable reverse DNS lookup


bahamapascal:
It would be great to have a optional command line parameter to disable the reverse DNS lookup.
Currently about 30 min after adding a IP/DNS the `getNeighbours` will not return the added IP/DNS as it was added but instead will return the DNS.
For example after adding `173.212.224.234` it will return `vmi112828.contaboserver.net` instead.

In some special cases this can lead to problems (for programs using the getNeighbours API call).

It would be great if this can be disabled with a extra command line parameter.

ixidion:
I still don't get it, why people share hard to memorize IP adresses.

To fix it:
1. Setup a DNS-Name, because IP adresses are not easy to memorize.
2. Setup Reverse DNS properly.

bahamapascal:
This issue is already fixed and got added to the dev bench. 
Big thanks to Ahab for the fix and Paul H for reviewing and merging!

PR #434, opened 2017-12-22
Add optional parameter to disable dns resolution

ahab94:
#413
Have added an optional parameter to disable dns resolution that can become a problem with applications serving the iota node. 
- optimized imports
- Added boolean argument for disabling dns-resolution
- Added predicate to bypass dns resolution.
- Linted some of the code


alon-e:
merged but didn't close automatically.

Issue #428, opened 2017-12-21
Fully synced node gets stuck after a few hours


torokati44:
The node is perfectly in sync for hours, then it just stops at a random milestone for no apparent reason, without being touched at all. The latestMilestone keeps advancing with COO, the latestSolidMilestone is stuck. It doesn't move at all for hours, but after an IRI restart, it will get back to full sync within a few minutes. Then another few hours pass, and it gets stuck again.

IRI version 1.4.1.4.
I have a couple manual neighbors, but also run Nelson (0.2.2).

torokati44:
At the moment it is 308589 / 308546, if this helps.

fijter:
Same here, with or without Nelson. Node has good neighbors but has been stuck at 243000 for hours now after a reboot.

mstrewe:
Can approve this. But is not stucking.. its hanging behind. I follow my node stats and after ~ 20 minutes the number increase from 308638 to 308639... I will look tomorrow if its increasing

mstrewe:
@fijter was your node already in sync?

fijter:
My node was fine and always in sync up until the moment the COO got paused, after that I had issues. After a couple of restarts my node is in sync again but it didn't move by itself.

stockermans:
I seem to have some issue as well. My node was in sync always with my neighbors, until the upgrade it seems. I had it staying updating the solidSubtangleMilestoneIndex like clockwork for a while yesterday (maybe a couple hours), and then it just stopped, and only the latestMilestoneIndex would continue, but my SolidSubtangleMilestone does not want to move.

I've tried with a fresh db from iota.partners and it will get in sync momentarily, then just will not continue.

I do not use nelson, only manual tethered neighbors which have been stable for a month+.

mstrewe:
OK. Now I can confirm that subtangle stopped completely.

scott79:
Adding some additional info found. Still need time to review further, however unsure if approach listed at end would make sense or cause other issues.

3 nodes

1 - consistently synced with latest milestone

2 - Not synced:

	1 - Public node and is connected to fully synced node
	1 - Private node connected to not synced public node

	Notes on non-synced nodes:

	- Upon clearing of db (on both nodes) and starting up process for non-synced public/private nodes. Public node requested transactions at a rate of about 1-2MB/sec from synced public node. The duration lasted
	for about 2 hours then dropped down to roughly 20kB/sec. 
[...]

PR #435, opened 2017-12-22
Invalid Subtangle Propagation

paulhandy:
Also added attachmentTimestamp filtering to received transactions.

### DIFF
--- a/src/main/java/com/iota/iri/TransactionValidator.java
+++ b/src/main/java/com/iota/iri/TransactionValidator.java
@@ -1,9 +1,11 @@
 package com.iota.iri;
 
+import com.iota.iri.controllers.ApproveeViewModel;
 import com.iota.iri.controllers.TipsViewModel;
 import com.iota.iri.hash.Curl;
 import com.iota.iri.hash.Sponge;
 import com.iota.iri.hash.SpongeFactory;
+import com.iota.iri.model.Transaction;
 import com.iota.iri.network.TransactionRequester;
 import com.iota.iri.controllers.TransactionViewModel;
 import com.iota.iri.model.Hash;
@@ -29,6 +31,7 @@ public class TransactionValidator {
     private final TransactionRequester transactionRequester;
     private final MessageQ messageQ;
     private int MIN_WEIGHT_MAGNITUDE = 81;
+    private static int MIN_TIMESTAMP = 1508760000;
 
     private Thread newSolidThread;
 
@@ -69,9 +72,16 @@ public class TransactionValidator {
         return MIN_WEIGHT_MAGNITUDE;
     }
 
-    private static void runValidation(TransactionViewModel transactionViewModel, final int minWeightMagnitude) {
+    private static boolean invalidTimestamp(TransactionViewModel transactionViewModel) {
+        if (transactionViewModel.getAttachmentTimestamp() == 0) {
+            return transactionViewModel.getTimestamp() < MIN_TIMESTAMP && !transactionViewModel.getHash().equals(Hash.NULL_HASH);
+        }
+        return transactionViewModel.getAttachmentTimestamp() < MIN_TIMESTAMP;
+    }
+
+    public static void runValidation(TransactionViewModel transactionViewModel, final int minWeightMagnitude) {
         transactionViewModel.setMetadata();
-        if(transactionViewModel.getTimestamp() < 1508760000 && !transactionViewModel.getHash().equals(Hash.NULL_HASH)) {
+        if(invalidTimestamp(transactionViewModel)) {
             throw new RuntimeException("Invalid transaction timestamp.");
         }
         for (int i = VALUE_TRINARY_OFFSET + VALUE_USABLE_TRINARY_SIZE; i < VALUE_TRINARY_OFFSET + VALUE_TRINARY_SIZE; i++) {
@@ -109,7 +119,7 @@ public class TransactionValidator {
     private final AtomicInteger nextSubSolidGroup = new AtomicInteger(1);
 
     public boolean checkSolidity(Hash hash, boolean milestone) throws Exception {
-        if(TransactionViewModel.fromHash(tangle, hash).isSolid()) {
+        if(TransactionViewModel.fromHash(tangle, hash).subtangleStatus() == SubtangleStatus.SOLID) {
             return true;
         }
         Set<Hash> analyzedHashes = new HashSet<>(Collections.singleton(Hash.NULL_HASH));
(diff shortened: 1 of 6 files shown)


src/main/java/com/iota/iri/TransactionValidator.java
```diff
@@ -232,12 +246,12 @@ private boolean quickSetSolid(final TransactionViewModel transactionViewModel) t
...
             }
             if(solid) {
-                transactionViewModel.updateSolid(true);
+                transactionViewModel.updateSolid(SubtangleStatus.SOLID);
                 transactionViewModel.updateHeights(tangle);
                 return true;
             }
         }
-        //return isSolid();
+        //return subtangleStatus();
```
alon-e: remove?
`subtangleStatus` doesn't return Boolean.

src/main/java/com/iota/iri/Milestone.java
```diff
@@ -47,6 +47,7 @@
     public Hash latestMilestone = Hash.NULL_HASH;
     public Hash latestSolidSubtangleMilestone = latestMilestone;
 
+    public static Hash INITIAL_MILESTONE_HASH = Hash.NULL_HASH;
```
alon-e: worth completing this change by using it in  Milestone:L47 as well:
```
public Hash latestMilestone = INITIAL_MILESTONE_HASH;
```

src/main/java/com/iota/iri/TransactionValidator.java
```diff
@@ -108,8 +118,8 @@ public static TransactionViewModel validate(final byte[] bytes, int minWeightMag
 
     private final AtomicInteger nextSubSolidGroup = new AtomicInteger(1);
 
-    public boolean checkSolidity(Hash hash, boolean milestone) throws Exception {
-        if(TransactionViewModel.fromHash(tangle, hash).isSolid()) {
+    public boolean checkSolidity(Hash hash, boolean priority) throws Exception {
+        if(TransactionViewModel.fromHash(tangle, hash).subtangleStatus() == SubtangleStatus.SOLID) {
             return true;
         }
         Set<Hash> analyzedHashes = new HashSet<>(Collections.singleton(Hash.NULL_HASH));
```
alon-e: `boolean solid = true;` not used anymore
(line below - couldn't add comment to that line)

src/main/java/com/iota/iri/TransactionValidator.java
```diff
@@ -249,7 +263,34 @@ private boolean checkApproovee(TransactionViewModel approovee) throws Exception
         if(approovee.getHash().equals(Hash.NULL_HASH)) {
             return true;
         }
-        return approovee.isSolid();
+        return approovee.subtangleStatus() == SubtangleStatus.SOLID;
+    }
+
+    public void propagateInvalidSubtangle(Hash invalidHash, boolean force) throws Exception {
```
alon-e: `invalidSubtangleStatus` is an API function name to check solid milestones.
`propagateInvalidSubtangle` + `subtangleStatus` are confusing names in that context.

maybe worth changing  API function name to reflect it's true meaning.

src/main/java/com/iota/iri/TransactionValidator.java
```diff
@@ -119,25 +129,29 @@ public boolean checkSolidity(Hash hash, boolean milestone) throws Exception {
...
+                            analyzedHashes.clear();
+                            return false;
+                        } else {
                             nonAnalyzedTransactions.offer(transaction.getTrunkTransactionHash());
                             nonAnalyzedTransactions.offer(transaction.getBranchTransactionHash());
                         }
-                    }
+                        break;
+                    case INVALID:
+                        if(!priority) {
```
alon-e: if this intended to allow `priority` to change `INVALID` to `SOLID`,
then it will only do so for the first "level" of invalids. as the traversal is only done for `UNKNOWN`.

this could also lead to false marking as solid:
think of `checkSolidity` called from `Milestone.updateLatestSolidSubtangleMilestone` (L227) where `priority` is `true`
if the starting transaction was marked as `INVALID`, then the while loop would end w/o traversing & mark the milestone as `SOLID` (`updateSolidTransactions`) w/o actually traversing & checking.

src/main/java/com/iota/iri/TransactionValidator.java
alon-e: `force` is always `false` when `propagateInvalidSubtangle` is called.

(in `checkSolidity` the call is wrapped w/ `if(!priority) { ... }`)

src/main/java/com/iota/iri/controllers/TransactionViewModel.java
```diff
@@ -369,21 +376,28 @@ public static void updateSolidTransactions(final Tangle tangle, final Set<Hash>
...
-            transactionViewModel.updateSolid(true);
+            transactionViewModel.updateSolid(SubtangleStatus.SOLID);
             transactionViewModel.update(tangle, "solid|height");
         }
     }
 
-    public boolean updateSolid(boolean solid) throws Exception {
-        if(solid != transaction.solid) {
-            transaction.solid = solid;
+    public boolean updateSolid(SubtangleStatus status) throws Exception {
```
alon-e: maybe worth changing the name now that it's not only solid - true / false:
`updateStatus` / `updateSubtangleStatus` / `updateSolidStatus` ?

Pomyk:
Is this a fix for #428 ?

src/main/java/com/iota/iri/controllers/TransactionViewModel.java
```diff
@@ -369,21 +376,28 @@ public static void updateSolidTransactions(final Tangle tangle, final Set<Hash>
...
+                case INVALID: transaction.solid = -1;
+            }
             return true;
         }
         return false;
     }
 
-    public boolean isSolid() {
-        return transaction.solid;
+    public SubtangleStatus subtangleStatus() {
```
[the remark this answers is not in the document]
rafaelbrizola: +1

paulhandy:
It's a step toward voluntary pruning

src/main/java/com/iota/iri/TransactionValidator.java
```diff
@@ -108,36 +124,37 @@ public static TransactionViewModel validate(final byte[] bytes, int minWeightMag
...
-                        solid = false;
-                        break;
-                    } else {
-                        if (solid) {
+                switch (transaction.subtangleStatus()) {
+                    case INVALID:
+                        if(!priority) {
+                            propagateInvalidSubtangle(transaction.getHash(), priority);
+                            return false;
+                        }
```
alon-e: the lack of a `break` is used implicitly to have `priority` treat INVALID as UNKNOWN?
if so, "explicit is better than implicit" - may seem as a bug, not intentional opt.

sreeraj-krishnan:
this does not fix sync issue

Issue #147, opened 2017-06-01
API exception in GUI 2.2.3


willowfoot:
Got a API Exception that flashed on screen trying to log in:

05/31 18:09:33.462 [XNIO-1 task-21] ERROR com.iota.iri.service.API - API Exception: 
java.lang.NullPointerException: null
	at com.iota.iri.service.API.exhaustiveSearchWithinIndex(API.java:518) [iri.jar:na]
	at com.iota.iri.service.API.getNewInclusionStateStatement(API.java:503) [iri.jar:na]
	at com.iota.iri.service.API.process(API.java:265) [iri.jar:na]
	at com.iota.iri.service.API.processRequest(API.java:147) [iri.jar:na]
	at com.iota.iri.service.API.access$000(API.java:81) [iri.jar:na]
	at com.iota.iri.service.API$1.handleRequest(API.java:135) [iri.jar:na]
	at io.undertow.server.Connectors.executeRootHandler(Connectors.java:208) [iri.jar:na]
	at io.undertow.server.HttpServerExchange$1.run(HttpServerExchange.java:809) [iri.jar:na]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142) [na:1.8.0_101]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617) [na:1.8.0_101]
	at java.lang.Thread.run(Thread.java:745) [na:1.8.0_101]

willowfoot:
same error happened with 2.2.4 ...

yillkid:
Same as dev branch:
 [XNIO-1 task-24] ERROR com.iota.iri.service.API - API Exception: 
java.lang.NullPointerException: null
	at com.iota.iri.service.API.exhaustiveSearchWithinIndex(API.java:555) [iri-1.2.4.jar:na]
	at com.iota.iri.service.API.getNewInclusionStateStatement(API.java:540) [iri-1.2.4.jar:na]
	at com.iota.iri.service.API.process(API.java:279) [iri-1.2.4.jar:na]

jakubcech:
Resolved

PR #440, opened 2017-12-24
Added functionality to check balance by tip

paulhandy:
This is a reasonable step toward safe zero-confirmation bundle promoting spends.
The client would query this to test if it can spend from a subtangle not yet checkpointed by a milestone.

### DIFF
--- a/src/main/java/com/iota/iri/service/API.java
+++ b/src/main/java/com/iota/iri/service/API.java
@@ -223,11 +223,11 @@ public class API {
                 }
                 case "getBalances": {
                     final List<String> addresses = getParameterAsList(request,"addresses", HASH_SIZE);
-                    final List<String> hashes = request.containsKey("hashes") ?
-                            getParameterAsList(request,"hashes", HASH_SIZE):
+                    final List<String> tips = request.containsKey("tips") ?
+                            getParameterAsList(request,"tips ", HASH_SIZE):
                             null;
                     final int threshold = getParameterAsInt(request, "threshold");
-                    return getBalancesStatement(addresses, hashes, threshold);
+                    return getBalancesStatement(addresses, tips, threshold);
                 }
                 case "getInclusionStates": {
                     if (invalidSubtangleStatus()) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/iota/iri/service/API.java
```diff
@@ -762,37 +765,44 @@ public void broadcastTransactionStatement(final List<String> trytes2) {
...
-            for (final Hash address : addresses) {
-                balances.put(address,
-                        instance.milestone.latestSnapshot.getState().containsKey(address) ?
-                                instance.milestone.latestSnapshot.getState().get(address) : Long.valueOf(0));
+            referenceSnapshot = new Snapshot(instance.milestone.latestSnapshot);
+        }
+        index = referenceSnapshot.index();
+        for(Hash hash: hashes) {
+            if (!instance.ledgerValidator.isTipConsistent(referenceSnapshot, hash)) {
+                return ErrorResponse.create("Hashes are not consistent");
```
alon-e: non-existing/yet-to-be-seen hashes also return `Hashes are not consistent` - this will be used for newly transmitted transactions, so getting this error message might be confusing.

an explicit check that the hashes are present would be best.

src/main/java/com/iota/iri/service/API.java
```diff
@@ -223,8 +223,11 @@ private AbstractResponse process(final String requestString, InetSocketAddress s
                 }
                 case "getBalances": {
                     final List<String> addresses = getParameterAsList(request,"addresses", HASH_SIZE);
+                    final List<String> hashes = request.containsKey("hashes") ?
```
alon-e: here the new field is called `hashes`, but the dto returns `references`.
also the iota.lib.js PR #147 call the new argument `references`, but passes `hashes` to the API.
also this doesn't match `getInclusionStates`'s naming conventions - `tips`

this is confusing - (I ran the API calls a few times w/ `references` till I found this).

paulhandy: I'll change it to `tips` in both, then :+1:  ?

PR #443, opened 2017-12-25
Added more fine-grained controls over getTxToApprove

paulhandy:
Allows for a user to select

* number of tips requested
* references with which the new tip must be consistent

### DIFF
--- a/src/main/java/com/iota/iri/service/API.java
+++ b/src/main/java/com/iota/iri/service/API.java
@@ -11,15 +11,7 @@ import java.net.URISyntaxException;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.security.SecureRandom;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.Queue;
-import java.util.Set;
+import java.util.*;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.regex.Matcher;
@@ -258,8 +250,10 @@ public class API {
                                 .create("This operations cannot be executed: The subtangle has not been updated yet.");
                     }
 
+                    final List<String> references = request.containsKey("hashes") ? getParameterAsList(request, "hashes", HASH_SIZE): new ArrayList<>();
                     final String reference = request.containsKey("reference") ? getParameterAsStringAndValidate(request,"reference", HASH_SIZE) : null;
                     final int depth = getParameterAsInt(request, "depth");
+                    final int count = request.containsKey("count") ? getParameterAsInt(request, "count"): 2;
                     if(depth < 0 || (reference == null && depth == 0)) {
                         return ErrorResponse.create("Invalid depth input");
                     }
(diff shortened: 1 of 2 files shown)


pRizz:
I think we should put a soft limit on `int tipsToApprove`. Potential CPU DOS attack otherwise.

paulhandy:
agreed

src/main/java/com/iota/iri/service/API.java
```diff
@@ -491,26 +497,40 @@ public static void incEllapsedTime_getTxToApprove(long ellapsedTime) {
...
+
+        referenceHashes = references == null? new ArrayList<>(): references.stream().map(h -> new Hash(h)).collect(Collectors.toList());
+
+        extraTip = referenceHashes.size() > 0 ? referenceHashes.get(0): null;
         synchronized (instance.milestone.latestSnapshot.snapshotSyncObject) {
             referenceSnapshot = new Snapshot(instance.milestone.latestSnapshot);
         }
+
+        for(Hash hash: referenceHashes) {
+            if (!instance.ledgerValidator.isTipConsistent(referenceSnapshot, hash)) throw new RuntimeException("inconsistent tips pair selected");
```
alon-e: this will fail if `hash` doesn't exist, but with the wrong message.

src/main/java/com/iota/iri/service/API.java
```diff
@@ -491,26 +497,40 @@ public static void incEllapsedTime_getTxToApprove(long ellapsedTime) {
...
+        Hash extraTip;
         if(reference != null) {
             referenceHash = new Hash(reference);
             if(!TransactionViewModel.exists(instance.tangle, referenceHash)) {
                 referenceHash = null;
             }
         }
-        Snapshot referenceSnapshot;
+
+        referenceHashes = references == null? new ArrayList<>(): references.stream().map(h -> new Hash(h)).collect(Collectors.toList());
```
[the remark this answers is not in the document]
paulhandy: I'll rename it to referenceTips

src/main/java/com/iota/iri/service/API.java
```diff
@@ -268,7 +264,7 @@ private AbstractResponse process(final String requestString, InetSocketAddress s
                         numWalks = minRandomWalks;
                     }
                     try {
-                        final Hash[] tips = getTransactionToApproveStatement(depth, reference, numWalks);
+                        final Hash[] tips = getTransactionToApproveStatement(depth, reference, referenceTips, count, numWalks);
```
alon-e: The return value (dto) does not support returning more than 2 tips. (explicitly gets `tips[0]` , `tips[1]` )

src/main/java/com/iota/iri/service/API.java
```diff
@@ -491,26 +487,41 @@ public static void incEllapsedTime_getTxToApprove(long ellapsedTime) {
...
-        final SecureRandom random = new SecureRandom();
-        final int randomWalkCount = numWalks > maxRandomWalks || numWalks < 1 ? maxRandomWalks:numWalks;
+    public synchronized Hash[] getTransactionToApproveStatement(final int depth, final String reference, final List<String> referenceTips, final int tipsToApprove, final int numWalks) throws Exception {
+        final Hash[] tips;
+        final SecureRandom random;
+        final int randomWalkCount;
+        final List<Hash> referenceTipsHashes;
+        final Snapshot referenceSnapshot;
+
+        tips = new Hash[tipsToApprove];
```
alon-e: requesting `count`: 1 results in an error - as the returned `tips` size is 1, but the dto is sent `tips[1]`

src/main/java/com/iota/iri/service/API.java
paulhandy: good call - but since the dto defines specifically `trunk` and `branch`, how could we modify this to work to our advantage?

I'd be most in favor of modifying the return value to just return `String[] tips` altogether, and at the very least when `count` is provided. But at that point, it begs the question - should this be a new API call (with the count variable)?

PR #447, opened 2017-12-27
Publish the trytes of a transaction to the ZMQ

pRizz:
- Publish the trytes of a transaction to the ZMQ for dapp purposes.
- Refactored ZmqPublishProvider to reduce nesting and separate concerns. 
- Using StringBuilder instead of StringBuffer because it is more efficient and is guaranteed to be thread-safe in this context.

### DIFF
--- a/src/main/java/com/iota/iri/storage/ZmqPublishProvider.java
+++ b/src/main/java/com/iota/iri/storage/ZmqPublishProvider.java
@@ -9,6 +9,7 @@ import org.slf4j.LoggerFactory;
 import com.iota.iri.controllers.TransactionViewModel;
 import com.iota.iri.model.Hash;
 import com.iota.iri.model.Transaction;
+import com.iota.iri.utils.Converter;
 import com.iota.iri.utils.Pair;
 import com.iota.iri.zmq.MessageQ;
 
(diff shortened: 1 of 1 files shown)


viossat:
I really need the `tx_trytes` ZMQ topic because `tx` doesn't provide all fields like `attachmentTimestamp` or the data inside the transaction. It could be great to document it in `src/main/java/com/iota/iri/zmq/README.md`.

Do you think it is useful to add the arrival time and the sender at the end?

pRizz:
@viossat The intention of this PR was only for adding the trytes to its own topic.

src/main/java/com/iota/iri/storage/ZmqPublishProvider.java
```diff
@@ -48,34 +49,58 @@ public void delete(Class<?> model, Indexable index) throws Exception {
...
+        TransactionViewModel transactionViewModel = new TransactionViewModel(transaction, (Hash)index);
+
+        publishTx(transactionViewModel);
+        publishTxTrytes(transactionViewModel);
+
+        return true;
+    }
+
+    private void publishTx(TransactionViewModel transactionViewModel) {
+        StringBuilder txStringBuffer = new StringBuilder(600);
```
GalRogozinski: change variable name to `txStringBuilder`

src/main/java/com/iota/iri/storage/ZmqPublishProvider.java
```diff
@@ -48,34 +49,58 @@ public void delete(Class<?> model, Indexable index) throws Exception {
...
+
+            messageQ.publish(txStringBuffer.toString());
+        } catch (Exception e) {
+            log.error(txStringBuffer.toString());
+            log.error("Error publishing tx to zmq.", e);
+        }
+    }
+
+    private void publishTxTrytes(TransactionViewModel transactionViewModel) {
+        StringBuilder txTrytesStringBuffer = new StringBuilder(2673);
```
GalRogozinski: change variable name to `txTrytesStringBuilder`

GalRogozinski: instead of `2673` use `TransactionViewModel.TRINARY_SIZE/3`

pRizz:
@GalRogozinski changes have been implemented. Thanks for the feedback

GalRogozinski:
@pRizz did you test this change?

This will put some pressure on the ZMQ feed as you are essentially sending all the transactions.
Let's open a discussion. I see 3 options:
1. Only send the trytes and zmq subscribers will parse it by itself.
2. Don't send the trytes and add missing attributes in accordance to @viossat offer.
3. Send trytes without the signature/message part and tx topics. What is the motivation to send both?

viossat:
@GalRogozinski No, I proposed to add `arrivalTime` and `sender` after the trytes. They may be useful.

I think the current behaviour (option 1) is good.

pRizz:
@GalRogozinski I kept the `tx` topic just to ensure backwards compatibility; I only refactored how it was implemented. I then added the `tx_trytes` topic so that any interested listeners can get the whole transaction. I am interested in the entire `tx_trytes` topic which allows for my applications to run: Chatangle and soon Votangle and more.

PR #455, opened 2018-01-01
workaround for issue #428

Pomyk:
Change lru cache into fifo with expire time.
Looks like sometimes a transaction doesn't get saved properly and
the lru cache was preventing to process it again. That missing
transaction was freezing solid subtangle milestone.

PR #459, opened 2018-01-02
allows "recently seen" caches to forget failed transaction stores

### LINKED ISSUE #428: Fully synced node gets stuck after a few hours

alon-e:
modifies Node caches from LRU to FIFO, 
with additional probabilistic drop cache entry - `P_DROP_CACHE_ENTRY`.

fixes #428 - credit does to @Pomyk that found the bug and offered a workaround (#455)!
this fixes the issue where storing a transaction would fail, but the "recently seen" cache would still hold a copy.

the difference from #455 is that the probabilistic solution has a much smaller memory and comp. footprint.
so not harming the common case, while still allowing the cache to refresh itself and recover from the rare case of failed store.

### DIFF
--- a/src/main/java/com/iota/iri/network/Node.java
+++ b/src/main/java/com/iota/iri/network/Node.java
@@ -758,10 +758,7 @@ public class Node {
 
         public V get(K key) {
             V value = this.map.get(key);
-            if (value == null) {
-                return null;
-            }
-            if (rnd.nextDouble() < this.dropRate) {
+            if (value != null && (rnd.nextDouble() < this.dropRate)) {
                 this.map.remove(key);
                 return null;
             }
--- a/src/main/java/com/iota/iri/network/Node.java
+++ b/src/main/java/com/iota/iri/network/Node.java
@@ -69,8 +69,8 @@ public class Node {
     private double P_PROPAGATE_REQUEST;
 
 
-    private LRUCache<Hash, Boolean> recentSeenHashes;
-    private LRUCache<ByteBuffer, Hash> recentSeenBytes;
+    private FIFOCache<Hash, Boolean> recentSeenHashes;
+    private FIFOCache<ByteBuffer, Hash> recentSeenBytes;
 
     private boolean debug;
     private static AtomicLong recentSeenBytesMissCount = new AtomicLong(0L);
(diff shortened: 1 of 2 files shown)


src/main/java/com/iota/iri/network/Node.java
```diff
@@ -742,35 +743,41 @@ public int getReplyQueueSize() {
...
 
         public V get(K key) {
             V value = this.map.get(key);
             if (value == null) {
-                value = null;
-            } else {
-                this.put(key, value);
+                return null;
+            }
+            if (rnd.nextDouble() < this.dropRate) {
```
th0br0: remove the previous if and replace this with

`if (value != null && rnd.nextDouble() < this.dropRate) {` ?

Pomyk:
I see one problem with this. Since it's a double cache you have to hit dropRate twice in a row. At 2% that's 1 in 2500 times. My first "hacky" solution used probability for recentSeenBytes and time eviction for recentSeenHashes (by storing Long with timestamp instead of Boolean).

alon-e:
@Pomyk good point.
as a result, I've been looking at the `recentSeenHashes` more deeply and running hit/miss analysis on it:
```
with no drop rate:
[pool-2-thread-4] INFO  com.iota.iri.network.Node - RecentSeenHashes cache hit/miss ratio: 55/445
[pool-2-thread-4] INFO  com.iota.iri.network.Node - RecentSeenHashes cache hit/miss ratio: 35/465
[pool-2-thread-4] INFO  com.iota.iri.network.Node - RecentSeenHashes cache hit/miss ratio: 26/474
[pool-2-thread-4] INFO  com.iota.iri.network.Node - RecentSeenHashes cache hit/miss ratio: 18/482

with 2% drop rate:
[pool-2-thread-4] INFO  com.iota.iri.network.Node - RecentSeenHashes cache hit/miss ratio: 16/484
[pool-2-thread-4] INFO  com.iota.iri.network.Node - RecentSeenHashes cache hit/miss ratio: 84/416
[pool-2-thread-4] INFO  com.iota.iri.network.Node - RecentSeenHashes cache hit/miss ratio: 83/417
[pool-2-thread-4] INFO  com.iota.iri.network.Node - RecentSeenHashes cache hit/miss ratio: 39/461
```

The rational behind `recentSeenHashes` was to serve as a memory cache for recently stored transactions, so that we wont need to access the DB to ask if a tx is already stored, but get a faster response.

while the rational behind `recentSeenBytes` is to serve as a filter for duplicate transactions "on the wire" due to gossip redundancy (echos).

however it seems that in practice, the `recentSeenHashes` is rarely "hit", due to the `recentSeenBytes` echo filter (added later in development), which just adds a layer of complexity & comp. resource not needed anymore.

node w/ 2 neighbors (the higher the neighbor count, the higher the hit ratio - more gossip echoes):
```
with no drop rate:
[pool-5-thread-2] INFO  com.iota.iri.network.Node - RecentSeenBytes cache hit/miss ratio: 217/283
[pool-5-thread-2] INFO  com.iota.iri.network.Node - RecentSeenBytes cache hit/miss ratio: 227/273
[pool-5-thread-1] INFO  com.iota.iri.network.Node - RecentSeenBytes cache hit/miss ratio: 219/281

with 2% drop rate:
[pool-5-thread-2] INFO  com.iota.iri.network.Node - RecentSeenBytes cache hit/miss ratio: 209/291
[pool-5-thread-2] INFO  com.iota.iri.network.Node - RecentSeenBytes cache hit/miss ratio: 211/289
[pool-5-thread-2] INFO  com.iota.iri.network.Node - RecentSeenBytes cache hit/miss ratio: 227/273

```

I propose to remove `recentSeenHashes` altogether.

PR #462, opened 2018-01-04
separate rate limit for API

alon-e:
allows setting rate via API call: `setApiRateLimit`
fixes limit counter,
uses a larger rate limit window.

the limit is off by default.

### DIFF
--- a/src/main/java/com/iota/iri/service/API.java
+++ b/src/main/java/com/iota/iri/service/API.java
@@ -91,7 +91,7 @@ public class API {
     private final int maxGetTrytes;
     private final int maxBodyLength;
     private final double newTransactionsRateLimit;
-    private final double newTransactionsLimit;
+    private double newTransactionsLimit;
     private final static String overMaxErrorMessage = "Could not complete request";
     private final static String invalidParams = "Invalid parameters";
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/iota/iri/service/API.java
```diff
@@ -91,7 +91,7 @@
     private final int maxGetTrytes;
     private final int maxBodyLength;
     private final double newTransactionsRateLimit;
-    private final double newTransactionsLimit;
+    private double newTransactionsLimit;
```
th0br0: could we put this below the final vars

src/main/java/com/iota/iri/service/API.java
```diff
@@ -354,7 +363,7 @@ private boolean isBelowNewTransactionLimit(InetAddress sourceAddress, int size)
             broadcastStoreTimer.set(now);
         }
 
-        if(broadcastStoreCounters.putIfAbsent(sourceAddress, new AtomicInteger(size)) != null) {
+        if(broadcastStoreCounters.putIfAbsent(sourceAddress, new AtomicInteger(size)) == null) { //not already in counter
             return size < newTransactionsLimit;
```
th0br0: this is a potential data race (parallel setApiRateLimit + read)
make this atomic?

src/main/java/com/iota/iri/service/API.java
```diff
@@ -977,6 +998,11 @@ private boolean validTrytes(String trytes, int length, char zeroAllowed) {
         Matcher matcher = trytesPattern.matcher(trytes);
         return matcher.matches();
     }
+    
+    public void setApiRateLimit(double apiRateLimit) {
+        newTransactionsLimit = (apiRateLimit * Neighbor.newTransactionsWindow) / 1000;
```
th0br0: same race as above

src/main/java/com/iota/iri/service/API.java
```diff
@@ -353,7 +354,8 @@ private AbstractResponse process(final String requestString, InetSocketAddress s
     }
 
     private boolean isBelowNewTransactionLimit(InetAddress sourceAddress, int size) {
-        if (newTransactionsLimit == 0) {
+        int newTransactionsLimitInt = newTransactionsLimit.get();
```
th0br0: nit: you might want to specify the order as relaxed here, that might give a minor speed boost (default is seq_cst)

paulhandy:
rebased ( but it didn't mark it as merged )

PR #471, opened 2018-01-07
Removed root access from loop-back devices

wbar:
This line was giving unlimited access to IRI remote commands like adding nodes or hosts to spam TCP/UDP packets

### DIFF
--- a/src/main/java/com/iota/iri/service/API.java
+++ b/src/main/java/com/iota/iri/service/API.java
@@ -185,8 +185,7 @@ public class API {
                 return ErrorResponse.create("COMMAND parameter has not been specified in the request.");
             }
 
-            if (instance.configuration.string(DefaultConfSettings.REMOTE_LIMIT_API).contains(command) &&
-                    !sourceAddress.getAddress().isLoopbackAddress()) {
+            if (instance.configuration.string(DefaultConfSettings.REMOTE_LIMIT_API).contains(command)) {
                 return AccessLimitedResponse.create("COMMAND " + command + " is not available on this node");
             }
 


nuriel77:
i don't think it does, because it says "ignore the limitations if the connection is to the loopback/local device". External connection cannot access the API when it is on loopback device unless you add a reverse proxy, which is a different topic.

wbar:
I noticed that someone edited my IRI server with custom nodes.. and this is what I'm afraid..
The point is how do you start the server you type on which IP has the server to listen to. If you want only locally you enter 127.0.0.1 but if you want public server you need to bind on public IP ... however - 99% of people enter the address 0.0 .0.0 telling "listen to all interfaces", i.e. local and public. 
Lines I'm changing in this PR gives ALL PERMISSIONS for requests from loopback interface! so you can download a list of other nods to which you are connected and add other servers.. 
And here the DDoS attack scenario: 
1. just connect to the server and trick that you are from localhost, 
2. download the list of neighbors 
3. add i.e. address of the site you want to disable 
4. repeat for all neighbors downloaded ... 

soon many nodes will want to communicate with the address of the site which we turn off by flooding requests - a DDoS attack

th0br0:
Did you run your server via a proxy?

GhostTyper:
How should it be possible to "trick that you are from localhost" and read the neighbours with that?

I think this change is not good because every automatic node management system and also nelson won't work, then.

footloosejava:
>just connect to the server and trick that you are from localhost,

This doesn't make any sense to me. The loopback address is localhost or 127.0.0.1 and there is not way that I know of to disguise an external packet. I double checked and there does not seem to be any way to spoof localhost that is not already properly identified and handled by the OS.

As GhostTyper said: "I think this change is not good because every automatic node management system and also nelson won't work, then."

GalRogozinski:
As the comments claim, this change is unneccessary

Issue #386, opened 2017-12-05
Replace docker command line options with environment variables


PatrickHuetter:
This would be better to support all orchestration tools without problems. E.g. rancher, google cloud and so on.

themarch:
+1

willianpaixao:
+1

PatrickHuetter:
Pull request for this can be found here: https://github.com/iotaledger/iri/pull/404

ahab94:
@PatrickHuetter, having it done by env variable for containerization is a plus. However, the changes you have made are a bit radical. I can purpose (and even contribute) a cleaner way around that with no change to the codebase. We will still be able to have environment variables defined in the image, that can be used with anysort of orchestration be it, simple `docker run`, `docker-compose` or any `kubernetes workload`.

PatrickHuetter:
@ahab94 I'll update the PR soon to be backward compatible with the current cli parameters.

ahab94:
@PatrickHuetter please take a look at #480

PR #480, opened 2018-01-09
Environment variables support for iri container

ahab94:
- Define and Add environment variables to Docker file, that gets built into the container.
- Change `CMD` to `ENTRYPOINT`.
- Define `docker-entrypoint.sh`

Now all the `iri` command line options can be defined and used inside the container without altering code or image. Which will make it compatible with several orchestration engines.    
  
A simple example:
```BASH
sudo docker run -e DNS_RESOLUTION_FALSE=yes -e MAX_PEERS=9 -it iri:latest 
```


ahab94:
This should make `iri` consumable in any sort of orchestration and resolve the following issue.  #386

geminoz:
Thanks Abdullah for the PR.

I really like your approach to the point that I privately used a similar setup for months. Adding a bash entry point allows greater flexibility and, as you said, allows for easier integration with orchestration tools.

As much as I want to keep passing ENV vars via docker and then have the container to source them in order to build the final java command, I recently took a step back and I am now advocating against using env vars to pass all sorts of IRI options to the final entry point command.

The reasons for this direction are:

the list of environment variables would have to be maintained in the entry point. Although this is possible, it also increases the technical debt over time with the risk of missing out "some" options
the Dockerfile and in return the entry point script would be way too opinionated (what env var to add to the entry point: the most important ones? Who defines which ones are the most important? All of them?
as new options are introduced, sysadmins and full node operators would have to change their tools accordingly and this is expected, but I struggled to find a reason to maintain a list of ENV vars allowed. Say for example a full node operator uses systemd via ansible, the places where changes need to occur in case of a new option is introduced or changed is doubled: the ansible playbook needs to be updated with a new variable that has to be "translated" to a docker run -e option.
while I was adding more and more env vars the list of ENV vars increased quickly and I had the impression the actual result was becoming increasingly more a hack rather than a best practice.
All the scenario above would be eliminated if IRI would source the env var themselves as command line option as this would eliminate the need to map an ENV var with a command line option. At this stage this is not possible.

So for the reasons above, I am proposing PR-744 where I attempt at covering all cases, leaving great flexibility to sysadmins and staying the least opinionated as possible. Arguments can be passed via command line and/or via a INI file which path can be defined at run time.

Hope you like the proposal, please let me know

Giorgio

ahab94:
Hi @geminoz! Thanks for the detail reasoning. I will be going through your pr and give feedback. Thanks.

Issue #460, opened 2018-01-04
Transaction resolves to incorrect ledger balance (thousands of time)


mjollnir14:
The following event keeps my iri blocked, this gets back after several restarts systematically and I got thousands of this same line in logs:

Jan 04 02:21:31 osiris IRI[18344]: 01/04 02:21:31.480 [Solid Milestone Tracker] INFO  com.iota.iri.Snapshot - Transaction resolves to incorrect ledger balance: -4151698081495

I think this is the issue preventing my node to keep in sync with LastMilestone.

> iota@osiris:~$ uname -a
> Linux osiris 3.16.0-4-amd64 #1 SMP Debian 3.16.39-1 (2016-12-30) x86_64 GNU/Linux
> iota@osiris:~$ ./status.sh
>   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
>                                  Dload  Upload   Total   Spent    Left  Speed
> 100   596  100   570  100    26  56846   2592 --:--:-- --:--:-- --:--:-- 63333
> {
>   "appName": "IRI",
>   "appVersion": "1.4.1.4",
>   "jreAvailableProcessors": 12,
>   "jreFreeMemory": 368470240,
>   "jreVersion": "1.8.0_151",
>   "jreMaxMemory": 7249854464,
>   "jreTotalMemory": 710410240,
>   "latestMilestone": "ID9RSTSRZEMSJKZCXTRAOQABOIETYGMIAOMYMXIPFYUUHWGYXVXQSTPMWMOX9ICBCQVDPQRUOBOAZ9999",
>   "latestMilestoneIndex": 319322,
>   "latestSolidSubtangleMilestone": "CWWMMLIJKXBLPFDWEUG9MKOQTXLHABVFTMUFUS9LDWUVHIGRCEKBK9FOAIRHLQVOOQPMYIBPJPMKA9999",
>   "latestSolidSubtangleMilestoneIndex": 319189,
>   "neighbors": 5,
>   "packetsQueueSize": 0,
>   "time": 1515029209112,
>   "tips": 273,
>   "transactionsToRequest": 0,
>   "duration": 1
> }
> 

> iota@osiris:~$ java -version
> java version "1.8.0_151"
> Java(TM) SE Runtime Environment (build 1.8.0_151-b12)
> Java HotSpot(TM) 64-Bit Server VM (build 25.151-b12, mixed mode)
> 
[...]

PR #486, opened 2018-01-18
Solid Milestone initialization fix

Desolatora:
### Bug
LedgerValidator initializes wrong consistent milestone after restart when node is not fully synced.

### Symptoms
When node is synced to solid milestone X and latest milestone is Y, after restart solid milestone is incorrectly initialized to Z (usually very close to Y).

### Cause
Method LedgerValidator.buildSnapshot() is responsible for building the most recent consistent milestone. It uses MilestoneViewModel.firstWithSnapshot() and MilestoneViewModel.nextWithSnapshot() to iterate on every milestone with snapshot then gets the state differences and patches them to the current snapshot. These methods deterministically rely on StateDiffViewModel.exists() which internally uses tangle.maybeHas(). The problem is that tangle.maybeHas() can only guarantee when entry is not present, but does not guarantee if entry is present. This leads to MilestoneViewModel.firstWithSnapshot() and MilestoneViewModel.nextWithSnapshot() sometimes returning milestone without snapshot or more precise milestone without StateDiff. This incorrect behaviour is also supported by the fact that StateDiffViewModel.load() returns an object even if StateDiff is not present. This leads to latestSnapshot being patched with no state differences and incorrect milestone index.

### Problems
When node is initialized with wrong solid milestone this may lead to solid milestone tracker appearing frozen while it is trying to solidify next milestone or compute state diff with both operations involving huge number of transactions. Sometimes wrong solid milestone index may lead to "Transaction resolves to incorrect ledger balance" errors (bug: #460).

### Proposed fix
Method StateDiffViewModel.isEmpty() is introduced to check if current state difference is empty. LedgerValidator.buildSnapshot() now iterates on every candidate milestone and patches latestSnapshot only with nonempty state differences. This guarantees that initialized consistent milestone is solid.

### DIFF
--- a/src/main/java/com/iota/iri/LedgerValidator.java
+++ b/src/main/java/com/iota/iri/LedgerValidator.java
@@ -190,6 +190,8 @@ public class LedgerValidator {
     protected void init() throws Exception {
         MilestoneViewModel latestConsistentMilestone = buildSnapshot();
         if(latestConsistentMilestone != null) {
+            log.info("Loaded consistent milestone: #" + latestConsistentMilestone.index());
+
             milestone.latestSolidSubtangleMilestone = latestConsistentMilestone.getHash();
             milestone.latestSolidSubtangleMilestoneIndex = latestConsistentMilestone.index();
         }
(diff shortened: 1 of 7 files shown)


src/main/java/com/iota/iri/LedgerValidator.java
```diff
@@ -221,20 +223,34 @@ public static boolean isApproved(Snapshot snapshot, Hash hash) {
      */
     private MilestoneViewModel buildSnapshot(Snapshot latestSnapshot) throws Exception {
         MilestoneViewModel consistentMilestone = null;
+
         synchronized (latestSnapshot.snapshotSyncObject) {
```
GalRogozinski: Use locks. See updated code

src/main/java/com/iota/iri/controllers/StateDiffViewModel.java
```diff
@@ -24,7 +24,7 @@ public StateDiffViewModel(final Map<Hash, Long> state, final Hash hash) {
         this.stateDiff.state = state;
     }
 
-    public static boolean exists(Tangle tangle, Hash hash) throws Exception {
+    public static boolean maybeExists(Tangle tangle, Hash hash) throws Exception {
```
GalRogozinski: Like

Desolatora:
Rebased with new code.

GalRogozinski:
Hey @Desolatora,
It seems like the rebase didn't roll out as smoothly as we like.
If you click on the "commits" tab or the "file changes" tab you will immediately see what I mean.

Git can be annoying sometimes, but we need to get it fixed.
If you find it too hard to fix this PR, you can close it and issue a new PR with only a link to this one in the description.

Desolatora:
Rebase fixed.

Desolatora:
It should be fixed now. If it isn't i will create a new PR.

Desolatora:
buildSnapshot() will now stop searching for candidates if inconsistent milestone is encountered.
Also i removed the log.isDebugEnabled() part and actually changed the log level to INFO. I think it is essencial for iri to report milestone initialization, because this is fairly slow process and it may appear that iri is stuck while actually it is properly initializing.

PR #487, opened 2018-01-18
Remove snapshot synchronisation objects

th0br0:
WIP


src/main/java/com/iota/iri/Snapshot.java
```diff
@@ -86,76 +87,141 @@
...
+                } else {
+                    return Optional.of(v);
+                }
+            }
+
+            Long v = ref.state.get(h);
+            if (l == null) {
+                l = v;
+            } else if (v != null) {
+                l += v;
```
alon-e: should we use safe math to avoid theoretical overflows?

src/main/java/com/iota/iri/Snapshot.java
```diff
@@ -86,76 +87,141 @@
...
+        return new Snapshot(this, diff, index);
     }
 
     boolean isConsistent() {
-        long stateValue = state.values().stream().reduce(Math::addExact).orElse(Long.MAX_VALUE);
-        if(stateValue != TransactionViewModel.SUPPLY) {
+        if (isConsistentMarker.get()) {
+            return true;
+        }
+        Map<Hash, Long> totalState;
```
alon-e: it seems wasteful that the `totalState` is calculated, but not stored in the snapshot itself.
this means that, for example, `buildSnapshot` recalculates the total state from initial snapshot each for each milestone being rebuilt - O(n^2).

src/main/java/com/iota/iri/service/API.java
```diff
@@ -806,10 +804,8 @@ private AbstractResponse getBalancesStatement(final List<String> addrss, final L
 
         final Map<Hash, Long> balances = new HashMap<>();
         final int index;
-        Snapshot referenceSnapshot;
-        synchronized (instance.milestone.latestSnapshot.snapshotSyncObject) {
-            referenceSnapshot = new Snapshot(instance.milestone.latestSnapshot);
-        }
+        Snapshot referenceSnapshot = instance.milestone.latestSnapshot.get();
```
alon-e: `getBalances` doesn't work.
it fails with `java.lang.StackOverflowError` in L814.

GalRogozinski:
We went with #489 instead

PR #489, opened 2018-01-20
Updated Snapshot usage to remove copious copies

paulhandy:
Two major changes:
* No more `new Snapshot` used after start
* `synchronized` snapshot object removed in favor of read write locks

### DIFF
--- a/src/main/java/com/iota/iri/LedgerValidator.java
+++ b/src/main/java/com/iota/iri/LedgerValidator.java
@@ -264,8 +264,9 @@ public class LedgerValidator {
         boolean isConsistent = Snapshot.isConsistent(milestone.latestSnapshot.patchedDiff(currentState));
         if (isConsistent) {
             currentState.forEach((key, value) -> {
-                diff.computeIfPresent(key, ((hash, aLong) -> value + aLong));
-                diff.putIfAbsent(key, value);
+                if(diff.computeIfPresent(key, ((hash, aLong) -> value + aLong)) == null) {
+                    diff.putIfAbsent(key, value);
+                }
             });
             approvedHashes.addAll(visitedHashes);
         }
--- a/src/main/java/com/iota/iri/LedgerValidator.java
+++ b/src/main/java/com/iota/iri/LedgerValidator.java
@@ -201,7 +201,7 @@ public class LedgerValidator {
     private MilestoneViewModel buildSnapshot() throws Exception {
         MilestoneViewModel consistentMilestone = null;
         StateDiffViewModel stateDiffViewModel;
-        milestone.latestSnapshot.rwlock.writeLock().lock();
+        long stamp = milestone.latestSnapshot.sl.readLock();
         try {
             MilestoneViewModel snapshotMilestone = MilestoneViewModel.firstWithSnapshot(tangle);
             while (snapshotMilestone != null) {
@@ -213,14 +213,14 @@ public class LedgerValidator {
                 }
             }
         } finally {
-            milestone.latestSnapshot.rwlock.writeLock().unlock();
+            milestone.latestSnapshot.sl.unlockRead(stamp);
         }
         return consistentMilestone;
     }
 
     public boolean updateSnapshot(MilestoneViewModel milestoneVM) throws Exception {
         TransactionViewModel transactionViewModel = TransactionViewModel.fromHash(tangle, milestoneVM.getHash());
-        milestone.latestSnapshot.rwlock.writeLock().lock();
+        long stamp = milestone.latestSnapshot.sl.readLock();
         try {
             final int transactionSnapshotIndex = transactionViewModel.snapshotIndex();
             boolean hasSnapshot = transactionSnapshotIndex != 0;
@@ -240,7 +240,7 @@ public class LedgerValidator {
             }
             return hasSnapshot;
         } finally {
-            milestone.latestSnapshot.rwlock.writeLock().unlock();
+            milestone.latestSnapshot.sl.unlockRead(stamp);
         }
     }
 
--- a/src/main/java/com/iota/iri/LedgerValidator.java
+++ b/src/main/java/com/iota/iri/LedgerValidator.java
@@ -43,14 +43,14 @@ public class LedgerValidator {
      * until it reaches a transaction that is marked as a "confirmed" transaction.
      * If {milestone} is false, it will search up until it reaches a confirmed transaction, or until it finds a hash that has been
      * marked as consistent since the previous milestone.
-     * @param visitedHashes         hashes that have been visited and considered as approved
-     * @param tip                   the hash of a transaction to start the search from
-     * @param latestSnapshotIndex   index of the latest snapshot to traverse to
-     * @param milestone             marker to indicate whether to stop only at confirmed transactions
-     * @return {state}              the addresses that have a balance changed since the last diff check
+     * @param visitedNonMilestoneSubtangleHashes hashes that have been visited and considered as approved
+     * @param tip                                the hash of a transaction to start the search from
+     * @param latestSnapshotIndex                index of the latest snapshot to traverse to
+     * @param milestone                          marker to indicate whether to stop only at confirmed transactions
+     * @return {state}                           the addresses that have a balance changed since the last diff check
      * @throws Exception
      */
-    public Map<Hash,Long> getLatestDiff(final Set<Hash> visitedHashes, Hash tip, int latestSnapshotIndex, boolean milestone) throws Exception {
+    public Map<Hash,Long> getLatestDiff(final Set<Hash> visitedNonMilestoneSubtangleHashes, Hash tip, int latestSnapshotIndex, boolean milestone) throws Exception {
         Map<Hash, Long> state = new HashMap<>();
         int numberOfAnalyzedTransactions = 0;
         Set<Hash> countedTx = new HashSet<>(Collections.singleton(Hash.NULL_HASH));
@@ -58,7 +58,7 @@ public class LedgerValidator {
         final Queue<Hash> nonAnalyzedTransactions = new LinkedList<>(Collections.singleton(tip));
         Hash transactionPointer;
         while ((transactionPointer = nonAnalyzedTransactions.poll()) != null) {
-            if (visitedHashes.add(transactionPointer)) {
+            if (visitedNonMilestoneSubtangleHashes.add(transactionPointer)) {
 
                 final TransactionViewModel transactionViewModel = TransactionViewModel.fromHash(tangle, transactionPointer);
                 if (transactionViewModel.snapshotIndex() == 0 || transactionViewModel.snapshotIndex() > latestSnapshotIndex) {
--- a/src/main/java/com/iota/iri/LedgerValidator.java
+++ b/src/main/java/com/iota/iri/LedgerValidator.java
@@ -43,9 +43,11 @@ public class LedgerValidator {
      * until it reaches a transaction that is marked as a "confirmed" transaction.
      * If {milestone} is false, it will search up until it reaches a confirmed transaction, or until it finds a hash that has been
      * marked as consistent since the previous milestone.
-     * @param tip       the hash of a transaction to start the search from
-     * @param milestone marker to indicate whether to stop only at confirmed transactions
-     * @return {state}  the addresses that have a balance changed since the last diff check
+     * @param visitedHashes         hashes that have been visited and considered as approved
+     * @param tip                   the hash of a transaction to start the search from
+     * @param latestSnapshotIndex   index of the latest snapshot to traverse to
+     * @param milestone             marker to indicate whether to stop only at confirmed transactions
+     * @return {state}              the addresses that have a balance changed since the last diff check
      * @throws Exception
      */
     public Map<Hash,Long> getLatestDiff(final Set<Hash> visitedHashes, Hash tip, int latestSnapshotIndex, boolean milestone) throws Exception {
--- a/src/main/java/com/iota/iri/Snapshot.java
+++ b/src/main/java/com/iota/iri/Snapshot.java
@@ -153,8 +153,9 @@ public class Snapshot {
         }
         long stamp = sl.writeLock();
         patch.entrySet().stream().forEach(hashLongEntry -> {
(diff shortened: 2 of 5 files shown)


src/main/java/com/iota/iri/LedgerValidator.java
```diff
@@ -198,103 +187,85 @@ private void updateConsistentHashes(final Snapshot snapshot, Hash tip, int index
...
+        MilestoneViewModel snapshotMilestone = MilestoneViewModel.firstWithSnapshot(tangle);
+        while (snapshotMilestone != null) {
+            stateDiffViewModel = StateDiffViewModel.load(tangle, snapshotMilestone.getHash());
+            if(Snapshot.isConsistent(milestone.latestSnapshot.patch(stateDiffViewModel.getDiff()))) {
+                milestone.latestSnapshot.merge(stateDiffViewModel.getDiff(), snapshotMilestone.index());
+                consistentMilestone = snapshotMilestone;
+                snapshotMilestone = snapshotMilestone.nextWithSnapshot(tangle);
             }
         }
+        milestone.latestSnapshot.rwlock.writeLock().unlock();
```
Awnedion: Should these unlocks be in finally blocks? If an exception is thrown during lockings, there will be deadlock.

src/main/java/com/iota/iri/LedgerValidator.java
```diff
@@ -198,103 +187,89 @@ private void updateConsistentHashes(final Snapshot snapshot, Hash tip, int index
...
-                    }
-                    StateDiffViewModel stateDiffViewModel;
-                    stateDiffViewModel = new StateDiffViewModel(currentState, milestone.getHash());
-                    if(currentState.size() != 0) {
-                        stateDiffViewModel.store(tangle);
-                    }
-                    latestSnapshot.merge(latestSnapshot.patch(stateDiffViewModel.getDiff(), milestone.index()));
+    public boolean updateSnapshot(Snapshot latestSnapshot, MilestoneViewModel milestoneVM) throws Exception {
+        TransactionViewModel transactionViewModel = TransactionViewModel.fromHash(tangle, milestoneVM.getHash());
+        this.milestone.latestSnapshot.rwlock.writeLock().lock();
```
alon-e: in `buildSnapshot` you don't use `this.*` - better to have a uniform style.

src/main/java/com/iota/iri/LedgerValidator.java
```diff
@@ -198,103 +187,89 @@ private void updateConsistentHashes(final Snapshot snapshot, Hash tip, int index
...
+        final int transactionSnapshotIndex = transactionViewModel.snapshotIndex();
+        boolean hasSnapshot = transactionSnapshotIndex != 0;
+        if(!hasSnapshot) {
+            Hash tail = transactionViewModel.getHash();
+            Set<Hash> visitedHashes = new HashSet<>();
```
alon-e: `visitedHashes` is never used in the function.
better just pass straight to `getLatestDiff`. otherwise, it's confusing: has a name -> has a purpose.

src/main/java/com/iota/iri/LedgerValidator.java
```diff
@@ -198,103 +187,89 @@ private void updateConsistentHashes(final Snapshot snapshot, Hash tip, int index
...
-                    Map<Hash, Long> currentState = getLatestDiff(snapshot, tail, latestSyncIndex, false);
-                    isConsistent = currentState != null && snapshot.patch(currentState, latestSyncIndex).isConsistent();
-                    if (isConsistent) {
-                        updateConsistentHashes(snapshot, tip, latestSyncIndex);
-                        snapshot.merge(snapshot.patch(currentState, latestSyncIndex));
-                    }
-                }
-            }
+        if (approvedHashes.contains(tip)) return true;
+        Set<Hash> visitedHashes = new HashSet<>(approvedHashes);
```
alon-e: can you explain why these are copied, and "copied" back (`approvedHashes.addAll(visitedHashes);`) at the end w/o any change?
or am I missing something?

src/main/java/com/iota/iri/LedgerValidator.java
```diff
@@ -171,17 +164,13 @@ private void updateSnapshotMilestone(MilestoneViewModel milestone) throws Except
      * @param tip
      * @throws Exception
      */
-    private void updateConsistentHashes(final Snapshot snapshot, Hash tip, int index) throws Exception {
+    private void updateConsistentHashes(final Set<Hash> visitedHashes, Hash tip, int index) throws Exception {
```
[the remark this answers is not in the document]
alon-e: this + the fact that `visitedHashes` is not updated in `updateDiff` anymore probably means you are missing a call to `updateConsistentHashes`

src/main/java/com/iota/iri/LedgerValidator.java
```diff
@@ -198,103 +187,89 @@ private void updateConsistentHashes(final Snapshot snapshot, Hash tip, int index
...
-            while (snapshotMilestone != null) {
-                stateDiffViewModel = StateDiffViewModel.load(tangle, snapshotMilestone.getHash());
-                updatedSnapshot = updatedSnapshot.patch(stateDiffViewModel.getDiff(), snapshotMilestone.index());
-                if (updatedSnapshot.isConsistent()) {
-                    consistentMilestone = snapshotMilestone;
-                    latestSnapshot.merge(updatedSnapshot);
-                    snapshotMilestone = snapshotMilestone.nextWithSnapshot(tangle);
-                }
+        StateDiffViewModel stateDiffViewModel;
+        milestone.latestSnapshot.rwlock.writeLock().lock();
```
alon-e: why not add lock/unlock read/write methods?

src/main/java/com/iota/iri/service/API.java
```diff
@@ -805,48 +807,32 @@ private AbstractResponse getBalancesStatement(final List<String> addrss, final L
...
-        index = referenceSnapshot.index();
-        for(Hash hash: hashes) {
-            if (!TransactionViewModel.exists(instance.tangle, hash)) {
-                return ErrorResponse.create("Tip not found: " + hash.toString());
-            }
-            if (!instance.ledgerValidator.isTipConsistent(referenceSnapshot, hash)) {
-                return ErrorResponse.create("Tips are not consistent");
-            }
-        }
+        instance.milestone.latestSnapshot.rwlock.readLock().lock();
```
alon-e: should be moved up to before L803? 
`hashes = Collections.singletonList(instance.milestone.latestSolidSubtangleMilestone);`

as the milestone can change between L803 and 810. so the index & milestone hash may mismatch.

src/main/java/com/iota/iri/LedgerValidator.java
paulhandy: Thanks

src/main/java/com/iota/iri/LedgerValidator.java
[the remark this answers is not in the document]
paulhandy: I was missing a call, thanks - but since this would be redundant, I refactored to just use a copy of the set fed into getLatestDiff, and update it if consistent.

src/main/java/com/iota/iri/LedgerValidator.java
paulhandy: I was missing the change, it's fixed now

src/main/java/com/iota/iri/service/API.java
```diff
@@ -797,56 +807,40 @@ private AbstractResponse getBalancesStatement(final List<String> addrss, final L
         final List<Hash> addresses = addrss.stream().map(address -> (new Hash(address)))
                 .collect(Collectors.toCollection(LinkedList::new));
         final List<Hash> hashes;
+        final Map<Hash, Long> balances = new HashMap<>();
+        final int index = instance.milestone.latestSnapshot.index();
```
alon-e: accessed w/o holding an explicit lock.
while this may be ok (the lock is obtained inside Snapshot). this is hard to reason about.
better to just move the lock above it imo.

src/main/java/com/iota/iri/LedgerValidator.java
```diff
@@ -49,24 +48,15 @@ public LedgerValidator(Tangle tangle, Milestone milestone, TransactionRequester
      * @return {state}  the addresses that have a balance changed since the last diff check
      * @throws Exception
      */
-    private Map<Hash,Long> getLatestDiff(final Snapshot snapshot, Hash tip, int latestSnapshotIndex, boolean milestone) throws Exception {
+    public Map<Hash,Long> getLatestDiff(final Set<Hash> visitedHashes, Hash tip, int latestSnapshotIndex, boolean milestone) throws Exception {
```
th0br0: Outdated docs, please fix

src/main/java/com/iota/iri/Snapshot.java
```diff
@@ -26,6 +31,7 @@
 
     public static final Map<Hash, Long> initialState = new HashMap<>();
     public static final Snapshot initialSnapshot;
+    public final ReadWriteLock rwlock = new ReentrantReadWriteLock();
```
th0br0: see https://stackoverflow.com/questions/26087738/performance-of-reentrantreadwritelock-read-lock and the follow up https://stackoverflow.com/questions/26070041/synchronizing-searches-and-modifications/26075089#26075089 for why we should rather use `StampedLock` here.

src/main/java/com/iota/iri/LedgerValidator.java
```diff
@@ -198,103 +184,90 @@ private void updateConsistentHashes(final Snapshot snapshot, Hash tip, int index
...
-                }
-            }
+        if (approvedHashes.contains(tip)) return true;
+        Set<Hash> visitedHashes = new HashSet<>(approvedHashes);
+        Map<Hash, Long> currentState = getLatestDiff(visitedHashes, tip, milestone.latestSnapshot.index(), false);
+        if (currentState == null) return false;
+        boolean isConsistent = Snapshot.isConsistent(milestone.latestSnapshot.patchedDiff(currentState));
+        if (isConsistent) {
+            currentState.forEach((key, value) -> {
+                diff.computeIfPresent(key, ((hash, aLong) -> value + aLong));
```
th0br0: This will return none if key is absent. Check for that and only all `putIfAbsent` conditionally?

src/main/java/com/iota/iri/Snapshot.java
```diff
@@ -86,75 +92,70 @@
...
-        state.clear();
-        state.putAll(snapshot.state);
-        index = snapshot.index;
+    void apply(Map<Hash, Long> patch, int newIndex) {
+        if (!patch.entrySet().stream().map(Map.Entry::getValue).reduce(Math::addExact).orElse(0L).equals(0L)) {
+            throw new RuntimeException("Diff is not consistent.");
+        }
+        rwlock.writeLock().lock();
+        patch.entrySet().stream().forEach(hashLongEntry -> {
+            state.computeIfPresent(hashLongEntry.getKey(), (hash, aLong) -> hashLongEntry.getValue() + aLong);
```
th0br0: same as above (only call putIfAbsent if `computeIfPresent` returns `null`)

src/test/java/com/iota/iri/SnapshotTest.java
```diff
@@ -21,45 +21,34 @@ public void getState() throws Exception {
...
+        Map<Hash, Long> diff = new HashMap<>();
+        diff.put(firstOne.getKey(), -firstOne.getValue());
+        diff.put(someHash, firstOne.getValue());
         Assert.assertNotEquals(0, diff.size());
-        Assert.assertTrue("The ledger should be consistent", newState.isConsistent());
+        Assert.assertTrue("The ledger should be consistent", Snapshot.isConsistent(Snapshot.initialSnapshot.patchedDiff(diff)));
     }
 
     @Test
     public void merge() throws Exception {
```
th0br0: Test should be renamed.

src/main/java/com/iota/iri/LedgerValidator.java
```diff
@@ -44,29 +43,22 @@ public LedgerValidator(Tangle tangle, Milestone milestone, TransactionRequester
      * until it reaches a transaction that is marked as a "confirmed" transaction.
      * If {milestone} is false, it will search up until it reaches a confirmed transaction, or until it finds a hash that has been
```
GalRogozinski: {milestone} -> {@code milestone}

src/test/java/com/iota/iri/SnapshotTest.java
```diff
@@ -21,45 +21,34 @@ public void getState() throws Exception {
...
         badMap.put(new Hash("PSRQPWWIECDGDDZEHGJNMEVJNSVOSMECPPVRPEVRZFVIZYNNXZNTOTJOZNGCZNQVSPXBXTYUJUOXYASLS"), 100L);
-        latestCopy.merge(latestCopy.patch(badMap, 0));
-        assertFalse("should be inconsistent", latestCopy.isConsistent());
+        badMap.put(new Hash("ESRQPWWIECDGDDZEHGJNMEVJNSVOSMECPPVRPEVRZFVIZYNNXZNTOTJOZNGCZNQVSPXBXTYUJUOXYASLS"), -100L);
+        Map<Hash, Long> patch = latestSnapshot.patchedDiff(badMap);
+        assertFalse("should be inconsistent", Snapshot.isConsistent(latestSnapshot.patchedDiff(badMap)));
     }
 
-    private Map<Hash, Long> getModifiedMap(Snapshot fromSnapshot) {
+    private Map<Hash, Long> getModifiedMap() {
```
GalRogozinski: This method is unused

src/test/java/com/iota/iri/SnapshotTest.java
```diff
@@ -21,45 +21,34 @@ public void getState() throws Exception {
...
         Hash someHash = new Hash("PSRQPWWIECDGDDZXHGJNMEVJNSVOSMECPPVRPEVRZFVIZYNNXZNTOTJOZNGCZNQVSPXBXTYUJUOXYASLS");
         Map<Hash, Long> newMap;
-        newMap = (Map<Hash, Long>) new HashMap<>(fromSnapshot.getState()).clone();
+        newMap = new HashMap<>();
```
GalRogozinski: Will we be iterating over an empty map?

paulhandy:
Merged into v1.4.1.7

GalRogozinski:
Changes were merged in separate of this PR so I am closing.

PR #498, opened 2018-01-25
TransactionsToApprove - unset reference if transaction is deeper than depth

alon-e:


src/main/java/com/iota/iri/service/API.java
```diff
@@ -531,8 +531,16 @@ public static void incEllapsedTime_getTxToApprove(long ellapsedTime) {
         Hash referenceHash = null;
         if(reference != null) {
             referenceHash = new Hash(reference);
-            if(!TransactionViewModel.exists(instance.tangle, referenceHash)) {
+            if (!TransactionViewModel.exists(instance.tangle, referenceHash)) {
+                //reference not found
                 referenceHash = null;
+            } else {
+                TransactionViewModel transactionViewModel = TransactionViewModel.fromHash(instance.tangle, referenceHash);
+                if (transactionViewModel.snapshotIndex() != 0
```
Awnedion: Does this != 0 check mean that a very large block of unconfirmed transactions can be used as a reference? Is that safe?

src/main/java/com/iota/iri/service/API.java
```diff
@@ -531,8 +531,16 @@ public static void incEllapsedTime_getTxToApprove(long ellapsedTime) {
...
+                        && transactionViewModel.snapshotIndex() < instance.milestone.latestSolidSubtangleMilestoneIndex - depth) {
```
th0br0: Depth needs to be bounded else it can be abused.

Awnedion: Depth is already bounded.

th0br0:
Closing because #501 contained these changes.

Issue #499, opened 2018-01-25
LatestSubtangleMilestone stucking after getTransactionsToApprove call


mstrewe:
The LatestSubtangleMilestone getting stucked in version 1.4.1.7. We discussed that in #fullnode discord channel. First idea was a ddos api attack but was not. I dumped every call on my node and realized, that after the first call of getTransactionsToApprove the subtangle milstone gets stucked. 

TCP dump attached. You can analyse the file with wireshark. 
[tcpdump.txt](https://github.com/iotaledger/iri/files/1662881/tcpdump.txt)

michi6383:
same here.

nuriel77:
https://github.com/iotaledger/iri/pull/498

GalRogozinski:
#501

PR #501, opened 2018-01-25
TransactionsToApprove - unset reference if transaction is deeper than depth 

GalRogozinski:
see PR #498 and issue #499

### DIFF
--- a/src/main/java/com/iota/iri/service/API.java
+++ b/src/main/java/com/iota/iri/service/API.java
@@ -523,12 +523,16 @@ public class API {
         ellapsedTime_getTxToApprove += ellapsedTime;
     }
 
-    public synchronized Hash[] getTransactionToApproveStatement(final int depth, final String reference, final int numWalks) throws Exception {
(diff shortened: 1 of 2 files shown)


src/main/java/com/iota/iri/service/API.java
```diff
@@ -523,16 +523,28 @@ public static void incEllapsedTime_getTxToApprove(long ellapsedTime) {
...
 
-    public synchronized Hash[] getTransactionToApproveStatement(final int depth, final String reference, final int numWalks) throws Exception {
+    public synchronized Hash[] getTransactionToApproveStatement(int depth, final String reference, final int numWalks) throws Exception {
         int tipsToApprove = 2;
         Hash[] tips = new Hash[tipsToApprove];
         final SecureRandom random = new SecureRandom();
         final int randomWalkCount = numWalks > maxRandomWalks || numWalks < 1 ? maxRandomWalks:numWalks;
         Hash referenceHash = null;
+        int maxDepth = instance.tipsManager.getMaxDepth();
+        if (maxDepth > depth) {
```
th0br0: Don't default to the maximum value.

PR #508, opened 2018-01-27
Feature: attachment timestamp filter

alon-e:
based on #435 ,
this allows reattaching old transactions from a previous epoch.

in addition, transactions with a timestamp in far future will also be filtered.

### DIFF
--- a/src/main/java/com/iota/iri/TransactionValidator.java
+++ b/src/main/java/com/iota/iri/TransactionValidator.java
@@ -73,19 +73,19 @@ public class TransactionValidator {
         return MIN_WEIGHT_MAGNITUDE;
     }
 
-    private static boolean invalidTimestamp(TransactionViewModel transactionViewModel) {
+    private static boolean hasInvalidTimestamp(TransactionViewModel transactionViewModel) {
         if (transactionViewModel.getAttachmentTimestamp() == 0) {
-            return transactionViewModel.getTimestamp() < MIN_TIMESTAMP && !transactionViewModel.getHash().equals(Hash.NULL_HASH)
+            return transactionViewModel.getTimestamp() < MIN_TIMESTAMP && !Objects.equals(transactionViewModel.getHash(), Hash.NULL_HASH)
                     || transactionViewModel.getTimestamp() > (System.currentTimeMillis() / 1000) + MAX_TIMESTAMP_FUTURE;
         }
         return transactionViewModel.getAttachmentTimestamp() < MIN_TIMESTAMP_MS
-                || transactionViewModel.getTimestamp() > System.currentTimeMillis() + MAX_TIMESTAMP_FUTURE_MS;
+                || transactionViewModel.getAttachmentTimestamp() > System.currentTimeMillis() + MAX_TIMESTAMP_FUTURE_MS;
     }
 
     public static void runValidation(TransactionViewModel transactionViewModel, final int minWeightMagnitude) {
         transactionViewModel.setMetadata();
         transactionViewModel.setAttachmentData();
-        if(invalidTimestamp(transactionViewModel)) {
+        if(hasInvalidTimestamp(transactionViewModel)) {
             throw new StaleTimestampException("Invalid transaction timestamp.");
         }
         for (int i = VALUE_TRINARY_OFFSET + VALUE_USABLE_TRINARY_SIZE; i < VALUE_TRINARY_OFFSET + VALUE_TRINARY_SIZE; i++) {
--- a/src/main/java/com/iota/iri/TransactionValidator.java
+++ b/src/main/java/com/iota/iri/TransactionValidator.java
@@ -84,6 +84,7 @@ public class TransactionValidator {
 
     public static void runValidation(TransactionViewModel transactionViewModel, final int minWeightMagnitude) {
         transactionViewModel.setMetadata();
+        transactionViewModel.setAttachmentData();
         if(invalidTimestamp(transactionViewModel)) {
             throw new StaleTimestampException("Invalid transaction timestamp.");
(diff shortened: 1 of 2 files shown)


src/main/java/com/iota/iri/network/Node.java
```diff
@@ -251,8 +252,9 @@ public void preProcessReceivedData(byte[] receivedData, SocketAddress senderAddr
 
                     if (!cached) {
                         //if not, then validate
-                        receivedTransactionViewModel = TransactionValidator.validate(receivedData, transactionValidator.getMinWeightMagnitude());
+                        receivedTransactionViewModel = new TransactionViewModel(receivedData, Hash.calculate(receivedData, TransactionViewModel.TRINARY_SIZE, SpongeFactory.create(SpongeFactory.Mode.CURLP81)));
```
th0br0: We shouldn't create a new instance each run here.

paulhandy: I think it's fine (better, even)

src/main/java/com/iota/iri/TransactionValidator.java
```diff
@@ -69,10 +73,20 @@ public int getMinWeightMagnitude() {
         return MIN_WEIGHT_MAGNITUDE;
     }
 
-    private static void runValidation(TransactionViewModel transactionViewModel, final int minWeightMagnitude) {
+    private static boolean invalidTimestamp(TransactionViewModel transactionViewModel) {
```
GalRogozinski: by convention it is better to call this: `hasInvalidTimeStamp`
See: http://geosoft.no/development/javastyle.html

src/main/java/com/iota/iri/TransactionValidator.java
```diff
@@ -69,10 +73,20 @@ public int getMinWeightMagnitude() {
...
+        if (transactionViewModel.getAttachmentTimestamp() == 0) {
+            return transactionViewModel.getTimestamp() < MIN_TIMESTAMP && !transactionViewModel.getHash().equals(Hash.NULL_HASH)
```
GalRogozinski: For defensive programming it is better to use `Objects.equals(transactionViewModel.getHash(), Hash.NULL_HASH)` or at least `Hash.NULL_HASH.equals(transaction.getHash())`.

Unless you WANT an exception to be thrown if `transactionViewModel.getHash() == null`, of course (then it will be nice to precede with `Objects.requireNonNull()`.

Even if you claim `transactionViewModel.getHash() == null` is an impossibility consider the following:
1. Man throughout history proved that the impossible is possible.
2. Even if it is impossible, getting into the habit of using `Objects.equals()` in general eliminates pesky null pointer exceptions without even thinking about them. So I want us to adopt that habit.

src/main/java/com/iota/iri/TransactionValidator.java
```diff
@@ -69,10 +73,20 @@ public int getMinWeightMagnitude() {
...
+                    || transactionViewModel.getTimestamp() > (System.currentTimeMillis() / 1000) + MAX_TIMESTAMP_FUTURE;
+        }
+        return transactionViewModel.getAttachmentTimestamp() < MIN_TIMESTAMP_MS
+                || transactionViewModel.getTimestamp() > System.currentTimeMillis() + MAX_TIMESTAMP_FUTURE_MS;
```
GalRogozinski: I see that two lines above we consider `transactionViewModel.getTimestamp()` as something that is measures in seconds. Now it is measured in ms...

Maybe you wanted to use `getAttachmentTimestamp()`??

I also recomment adding the words `seconds` to `getTimestamp` if we use seconds rather than ms.

src/main/java/com/iota/iri/network/Node.java
```diff
@@ -221,7 +222,7 @@ private Runnable spawnNeighborDNSRefresherThread() {
     }
 
     public void preProcessReceivedData(byte[] receivedData, SocketAddress senderAddress, String uriScheme) {
-        TransactionViewModel receivedTransactionViewModel = null;
+        TransactionViewModel receivedTransactionViewModel;
```
GalRogozinski: I actually like explicitly setting variables null...

If we want to improve the code on PRs (which I've been told to refrain from doing) this would be one thing I wouldn't be doing.

src/main/java/com/iota/iri/network/Node.java
GalRogozinski: I agree with Paul, unless you can show that creating instances causes an issue.

src/main/java/com/iota/iri/network/Node.java
```diff
@@ -265,6 +267,14 @@ public void preProcessReceivedData(byte[] receivedData, SocketAddress senderAddr
...
                 } catch (NoSuchAlgorithmException e) {
                     log.error("MessageDigest: " + e);
+                } catch (final TransactionValidator.StaleTimestampException e) {
+                    log.error(e.getMessage());
+                    try {
+                        transactionRequester.clearTransactionRequest(receivedTransactionHash);
+                    } catch (Exception e1) {
+                        log.error(e1.getMessage());
+                    }
+                    neighbor.incInvalidTransactions();
```
GalRogozinski: put in `finally` block maybe?

src/main/java/com/iota/iri/network/Node.java
alon-e: fair enough,
it was intellij's idea :)

src/main/java/com/iota/iri/network/Node.java
alon-e: this should only happen if an exception is thrown, not always.

PR #516, opened 2018-01-29
Add milestoneStartIndex to getNodeInfo

rajivshah3:
This exposes `Milestone.MILESTONE_START_INDEX` to the `getNodeInfo` API call. It can be used for checking the milestone that a snapshot was performed at. If stored, it can be used to detect whether a snapshot occurred.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -2,8 +2,9 @@
 Hints for a successful PR:
 1. It is recommended that before you submit a PR to IOTA, to open an issue first and assign yourself.
 This way you may get inputs and discover parallel PRs to the one you want to submit.
-2. In case of a big PR, consider breaking it up to smaller PRs. This will help getting it merged in an incremental process
-3. It will be helpful if you make additional comments on the code via github PR review to explain the chooices you made
+2. In case of a big PR, consider breaking it up to smaller PRs. This will help getting it merged in an incremental process.
+3. Note that a PR should have a *single* area of responsibility. If your PR does more than one thing than it should be splitted to several PRs!!!!!
+4. It will be helpful if you make additional comments on the code via github PR review to explain the choices you made
 -->
 
 # Description


dependency-reduced-pom.xml
```diff
@@ -4,7 +4,7 @@
   <groupId>com.iota</groupId>
   <artifactId>iri</artifactId>
   <name>IRI</name>
-  <version>1.4.1.7</version>
+  <version>1.4.2.1</version>
```
jserv: It is independent change.

src/main/java/com/iota/iri/service/dto/GetNodeInfoResponse.java
```diff
@@ -18,6 +18,8 @@
     private String latestSolidSubtangleMilestone;
     private int latestSolidSubtangleMilestoneIndex;
 
+		private int milestoneStartIndex;
```
mlouielu: I believe this is an unexpected indent.

dependency-reduced-pom.xml
rajivshah3: When I built it to check if my changes worked properly, I think it made this change automatically. See #523

src/main/java/com/iota/iri/service/dto/GetNodeInfoResponse.java
rajivshah3: Thanks. I think it's because I used `tab` instead of 4 spaces like the rest of the code used

GalRogozinski:
Why did you delete the dependency-reduced-pom?
If you think it should be removed from git please do it in a seperate PR with an explanation.

Thanks

rajivshah3:
@GalRogozinski it was deleted in #523 which was merged into develop so it was marked as a conflicting file

GalRogozinski:
Did you test that the libs can parse the new response correctly?

rajivshah3:
@GalRogozinski just like #567 , I don't think it should cause any problems as long as the lib can parse JSON properly

PR #518, opened 2018-01-29
add optional reference tips to get transactions to approve

paulhandy:
I think that larger bundles should reference more than just the two first tx, to be able to merge subtangles equally with small tx expanding them.

This allows to run getTxToApprove multiple times (i,e. throughout the PoW phase), each time giving the previous tips used.

### DIFF
--- a/src/test/java/com/iota/iri/NodeIntegrationTests.java
+++ b/src/test/java/com/iota/iri/NodeIntegrationTests.java
@@ -135,7 +135,7 @@ public class NodeIntegrationTests {
     }
 
     private void sendMilestone(API api, long index) throws Exception {
-        newMilestone(api, api.getTransactionToApproveStatement(10, null, 1), index);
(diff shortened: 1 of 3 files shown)


src/test/java/com/iota/iri/NodeIntegrationTests.java
```diff
@@ -135,7 +135,7 @@ Runnable spawnCoordinator(API api, long spacing) {
     }
 
     private void sendMilestone(API api, long index) throws Exception {
-        newMilestone(api, api.getTransactionToApproveStatement(10, null, 1), index);
+        newMilestone(api, api.getTransactionToApproveStatement(10, null, 1, new ArrayList<>()), index);
```
GalRogozinski: Collections.EMPTY_LIST will be better.
Saves memory and IIUC the list is supposed to be immutable

PR #520, opened 2018-01-30
v1.4.2.1-testnet

th0br0:


src/main/java/com/iota/iri/TransactionValidator.java
```diff
@@ -55,9 +55,9 @@ public void init(boolean testnet,int MAINNET_MWM, int TESTNET_MWM) {
         } else {
             MIN_WEIGHT_MAGNITUDE = MAINNET_MWM;
         }
-        //lowest allowed MWM encoded in 46 bytes.
-        if (MIN_WEIGHT_MAGNITUDE<13){
-            MIN_WEIGHT_MAGNITUDE = 13;
+        //lowest allowed MWM encoded in 49 bytes.
+        if (MIN_WEIGHT_MAGNITUDE<9){
+            MIN_WEIGHT_MAGNITUDE = 9;
```
GalRogozinski: Why is 9 hardcoded?
I see we can take it from config...

th0br0: Lower MWM bound depends on `REQUEST_HASH_SIZE` & `TRANSACTION_PACKET_SIZE`.

GalRogozinski:
Outdated

PR #522, opened 2018-01-31
Make the pause between broadcasts dynamic

paulhandy:
If we allow the pause between broadcasts to be dynamic, we could allow an ixi script ( or an API call if we decide so ) to update the broadcast delay to trigger local queue saturation if a node is able to detect a malicious flood of transactions.

An example nashorn script to use this:
```
var iri = com.iota.iri;
var node = IOTA.node;

var Callable = iri.service.CallableRequest;
var Response = iri.service.dto.IXIResponse;

function setPause(request) {
    var delay = java.lang.Long.parseLong(request.get("ns"));
    node.setPauseBetweenTransactions(delay)
    return Response.create({
        millis: node.msPauseBetweenTransactions,
        nanos: node.nsPauseBetweenTransactions
    });
}

API.put("setPause", new Callable({ call: setPause }))
```

### DIFF
--- a/src/main/java/com/iota/iri/network/Node.java
+++ b/src/main/java/com/iota/iri/network/Node.java
@@ -37,7 +37,6 @@ public class Node {
     private int BROADCAST_QUEUE_SIZE;
     private int RECV_QUEUE_SIZE;
     private int REPLY_QUEUE_SIZE;
-    private static final int PAUSE_BETWEEN_TRANSACTIONS = 1;
     public static final int REQUEST_HASH_SIZE = 46;
     private static double P_SELECT_MILESTONE;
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/iota/iri/network/Node.java
```diff
@@ -515,6 +516,11 @@ private Runnable spawnBroadcasterThread() {
         };
     }
 
+    public void setPauseBetweenTransactions(long nanoseconds) {
```
GalRogozinski: Since this setter doesn't simply assign the parameter to a field but employs logic maybe adding a unit test is a good idea.

GalRogozinski: If you add a unit test @paulhandy I just might approve this change

GalRogozinski:
when we overhaul the network implementation this will be not needed sees #583

PR #523, opened 2018-02-02
Don't store dependency-reduced-pom.xml in git.

msrb:
The file is a byproduct of the Maven Shade plugin. No need to keep it in git.


rajivshah3:
Ideally I think all of these should be added: https://www.gitignore.io/api/maven

msrb:
> Ideally I think all of these should be added: https://www.gitignore.io/api/maven

Nice! I didn't know about gitignore.io. Thanks :wink:
I can add the rest, probably in a separate PR, if people agree.

th0br0:
Feel free to create the other PR!

PR #528, opened 2018-02-07
TipSelection - reduce "inconsistent tips pair selected" by having second walker aware of first walker's PoV

### LINKED ISSUE #527: bug: TipSelection - "inconsistent tips pair selected" appears too often
Due to changes in the Snapshot immutability structure - `visitedHashes` no longer reflects the "what-if" analysis done by each random walk.

this means that the 2 walks (for `branch` and `trunk`) are totally independent and may often conflict.

I'll propose a PR to remedy this.

alon-e:
update visitedHashes & diff with selected tip's view - resolves #527

### DIFF
--- a/src/main/java/com/iota/iri/service/TipsManager.java
+++ b/src/main/java/com/iota/iri/service/TipsManager.java
@@ -161,7 +161,7 @@ public class TipsManager {
                 monteCarloIntegrations.put(tail,1);
             }
         }
-        return monteCarloIntegrations.entrySet().stream().reduce((a, b) -> {
+        Hash resultTip = monteCarloIntegrations.entrySet().stream().reduce((a, b) -> {
             if (a.getValue() > b.getValue()) {
                 return a;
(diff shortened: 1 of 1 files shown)


src/main/java/com/iota/iri/service/TipsManager.java
```diff
@@ -161,7 +161,7 @@ Hash markovChainMonteCarlo(final Set<Hash> visitedHashes, final Map<Hash, Long>
                 monteCarloIntegrations.put(tail,1);
             }
         }
-        return monteCarloIntegrations.entrySet().stream().reduce((a, b) -> {
+        Hash resultTip = monteCarloIntegrations.entrySet().stream().reduce((a, b) -> {
```
GalRogozinski: Did you leave out the case where a.getValue() == b.getValue() on purpose?

paulhandy: `seed.nextBoolean()` a couple lines below it

PR #535, opened 2018-02-10
Accelerate rating calculation

olaf-2:
I know it's not a hot spot, but with increasing number of tips and incoming transactions even milliseconds add up. I propose to replace the call to the general and quite expensive `Math.pow()` method with simple multiplication and division which are faster by a factor of >20.

### DIFF
--- a/src/main/java/com/iota/iri/service/TipsManager.java
+++ b/src/main/java/com/iota/iri/service/TipsManager.java
@@ -243,8 +243,8 @@ public class TipsManager {
                 long tipRating = ratings.get(tip);
                 for (int i = 0; i < tips.length; i++) {
                     //transition probability = ((Hx-Hy)^-3)/maxRating
-                	long rating = tipRating - ratings.getOrDefault(tips[i],0L);
-                	double divisior = (double) rating * rating * rating;
+                    long rating = tipRating - ratings.getOrDefault(tips[i],0L);
+                    double divisior = (double) rating * rating * rating;
                     walkRatings[i] = 1.0 / divisior;
                     maxRating += walkRatings[i];
                 }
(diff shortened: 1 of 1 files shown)


olaf-2:
```JAVA
package com.iota.iri.service;

import java.util.Random;

public class AccelRatingCalc {

    public static void main(String[] args) {
        int length = 10_000_000;
        long[] t = new long[length];
        double resultPow[] = new double[length];
        double resultMul[] = new double[length];
        Random r = new Random();
        for (int i = 0; i < 10; i++) {
            for (int j = 0; j < t.length; j++) {
                t[j] = (long) r.nextInt(1000);
            }
            long start = System.currentTimeMillis();
            for (int j = 0; j < length; j++) {
                resultPow[j] = Math.pow(t[j], -3);
            }
            System.out.println("pow: " + (System.currentTimeMillis() - start));
            start = System.currentTimeMillis();
            for (int j = 0; j < length; j++) {
                long rating = t[j];
                double div = (double) rating * rating * rating;
                resultMul[j] = 1.0 / div;
            }
            System.out.println("mul: " + (System.currentTimeMillis() - start));
            start = System.currentTimeMillis();
            for (int j = 0; j < length; j++) {
                if (Math.abs(resultMul[j] - resultPow[j]) > 1e-6) {
                    System.out.println("not equal for " + j);
                }
            }
        }
    }
}
```

"On my machine"(TM) Math.pow need 550 ms vs 20 ms mul&div.

GalRogozinski:
Thanks for your contribution.
We are rewriting that calculation to be more similar to the white paper.
So for the meanwhile we are not merging.

GalRogozinski:
@olaf-2 
Thanks for your contribution.
I want to be clear that the reason this is rejected is not because it is not good. It is because we want that the next change will use the whitepaper formula: `walkRatings[i] = Math.exp(-alpha*(tipRating - ratings.getOrDefault(tips[i], 0L)`. Where `alpha` should be a *configurable* value between 0 and 1.

Notice that due to the nature of how computers deal with small `double` numbers one needs to change the way the formula is written using a normalization trick. Also some optimizations may be added.

If you are up to the task of implementing this please open an issue and assign yourself. Thanks!

PR #540, opened 2018-02-18
Added getTagValue to ZMQ tx topic

SteveFromTheOffice:
It's about time, anyways!

### DIFF
--- a/src/main/java/com/iota/iri/network/TCPNeighbor.java
+++ b/src/main/java/com/iota/iri/network/TCPNeighbor.java
@@ -77,12 +77,18 @@ public class TCPNeighbor extends Neighbor {
         this.sink = sink;
     }
 
+    /**
+     * This is a non-blocking write and that makes it necessary to make a defensive copy of the sent data.
+     *
+     * @param packet the data to be queued for sending.
+     */
     @Override
     public void send(DatagramPacket packet) {
-        if ( sendQueue.remainingCapacity() == 0 ) {
+        if (sendQueue.remainingCapacity() == 0) {
             sendQueue.poll();
         }
-        sendQueue.add(ByteBuffer.wrap(packet.getData()));
+        byte[] bytes = packet.getData().clone();
+        sendQueue.add(ByteBuffer.wrap(bytes));
     }
 
     @Override
--- a/src/main/java/com/iota/iri/network/UDPNeighbor.java
+++ b/src/main/java/com/iota/iri/network/UDPNeighbor.java
@@ -16,11 +16,16 @@ public class UDPNeighbor extends Neighbor {
 
     private final DatagramSocket socket;
 
-    public UDPNeighbor(final InetSocketAddress address, final DatagramSocket socket, final boolean isConfigured) {
+    UDPNeighbor(final InetSocketAddress address, final DatagramSocket socket, final boolean isConfigured) {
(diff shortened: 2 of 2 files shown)


th0br0:
Don't replace the field, please.
I'm happy to accept a PR to append the tag to the ZMQ output as necessary though. (in the right position of course)

cwarner818:
Why would you want to keep `obsoleteTag`? Isn't it obsolete?

SteveFromTheOffice:
Appended it instead, based it on the tryte order (what I assumed was the right position). If not, let me know!

cwarner818:
It doesn't hold any data AFAIK. It's just a copy of the `tag` that gets overwritten during hashing. Am I misunderstanding it?

cwarner818:
What if we expanded the transaction output via zeromq to contain all transaction fields, and change as necessary as IRI evolves? That way we include all the data and its up to the foundation to not keep switching up the protocol :) 

But for now I think it makes a lot of sense to switch out `obosleteTag` for `tag` unless someone has a specific use case for keeping `obosleteTag`.

src/main/java/com/iota/iri/storage/ZmqPublishProvider.java
```diff
@@ -148,4 +149,4 @@ public void clearMetadata(Class<?> column) throws Exception {
 
     }
 
-}
\ No newline at end of file
+}
```
GalRogozinski: Why did you add an extra line?

GalRogozinski:
Please append the new value at the very end. This is so that people who parse the ZMQ tx topic according to positions won't be affected by movement ArrivalTime's position.

Also, maybe you are willing to edit `com/iota/iri/zmq/README.md` to describe the ZMQ topic structure.

cwarner818:
Isn't obsoleteTag going to be removed entirely in the future? Why not rip that bandaid off now instead of pushing it off until later? Everything I've seen that uses obsoleteTag is just using it to display the tag field as if it was still valid (ie http://tangle.glumb.de/) anyways.

SteveFromTheOffice:
Apologies for the delay, was MIA for a few weeks. I'm also willing to look at the readme and add details to it, but I think we should PR that separately.

GalRogozinski:
@SteveFromTheOffice If you can open a separate PR like you suggested it will be great!

alon-e:
This PR is ready for merging, but requires informing ZMQ users,
even though ZMQ is still experimental and non-default - there is code dependant on the format.
@geminoz @viossat @alon-e @tsvi -- please test this PR against your existing implementations and report back.

viossat:
@alon-e It's ok here. (thanks for the ping)
Please also consider #447

geminoz:
Saw this now. 

Heaps of changes in the dev branch in the past week. Is a rebase of this PR required to test against the newer code?

alon-e:
@gem for testing your infra no, just run this PR and test the `tx` topic parser.

But in order to merge it, a small rework is needed, to match #447 merged changes.

tsvisabo:
No problems on my side

alon-e:
works on my implementation also. 
-- @SteveFromTheOffice while reworking PR to match latest changes the PR lost context to upstream, hence I lost the ability to push the rework to your fork. sorry about that.

I'll create a new PR with the correct changes and merge.

PR #557, opened 2018-02-25
Improve log message for upd send errors.

georgmittendorfer:
This is only a minor change. The current message did cost me half an hour to find my firewall problem, though (typo in port number). It might help someone else, too. 

Changes:

- log target address in case of udp send errors
- log message with toString() instead of getMessage(). This gives better information as it additionally logs the exceptions name. Some exceptions don't have a message, e.g. NPE.
- made socket final as it is never changed

Feel free to simply discard or to alter (log wording?) the changes if you don't like them without any comment.

log before:

`02/25 18:28:58.898 [pool-2-thread-1] ERROR com.iota.iri.network.UDPNeighbor - UDP send error: Operation not permitted (sendto failed)`

log after:

`02/25 21:19:15.947 [pool-2-thread-1] ERROR com.iota.iri.network.UDPNeighbor - Error sending UDP packet to [foo.bar/91.194.91.220:1460]: java.io.IOException: Operation not permitted (sendto failed)`


src/test/java/com/iota/iri/network/UDPNeighborTest.java
```diff
@@ -0,0 +1,52 @@
...
+
+    @Test
+    public void whenSendThenSendPacketToSocket() throws IOException {
+        UDPNeighbor neighbor = new UDPNeighbor(new InetSocketAddress(HOST, PORT), udpSocket, true);
+        neighbor.send(packet);
+        verify(udpSocket).send(packet);
+    }
+
+    @Test
+    public void whenSendExceptionThenCatchAndContinue() throws IOException {
```
GalRogozinski: So when does the test fail?

Consider a solution like this:
https://stackoverflow.com/questions/2469911/how-do-i-assert-my-exception-message-with-junit-test-annotation

georgmittendorfer: It will never fail as there is a try catch in the code. I just used this method to check the log message. I am going to remove the tests or feel free to do so.

georgmittendorfer:
Fixed

PR #558, opened 2018-02-26
Tip Selection - memory efficient algorithm for computing comulative weights

GalRogozinski:
Context: The whitepaper describes a need to perform cumulative weight calculations on the transactions in order to perform the MCMC algorithm.

Problem: There are two versions of the algorithm implemented in the code:

In use - a time and memory efficient algorithm that performs a different calculation than the one described in the WP. Each transaction sums the weight of its direct ancestors and adds itself. This way indirect ancestors may be counted more than once. This may cause weight to increase exponentially.

Solution: A space time efficient algorithm similar to algo (2). If you traverse the subtangle in topological order, you can dispose of sets you have used. See https://github.com/alongalky/iota-docs/blob/master/cumulative.md.

### DIFF
--- a/src/main/java/com/iota/iri/service/TipsManager.java
+++ b/src/main/java/com/iota/iri/service/TipsManager.java
@@ -3,18 +3,25 @@ package com.iota.iri.service;
 import java.util.*;
 
 import com.iota.iri.LedgerValidator;
-import com.iota.iri.Snapshot;
+import com.iota.iri.Milestone;
 import com.iota.iri.TransactionValidator;
+import com.iota.iri.controllers.ApproveeViewModel;
+import com.iota.iri.controllers.MilestoneViewModel;
+import com.iota.iri.controllers.TipsViewModel;
+import com.iota.iri.controllers.TransactionViewModel;
 import com.iota.iri.model.Hash;
-import com.iota.iri.controllers.*;
 import com.iota.iri.storage.Tangle;
+import com.iota.iri.utils.collections.BoundedSetValuedHashMap;
 import com.iota.iri.zmq.MessageQ;
+import org.apache.commons.collections4.CollectionUtils;
+import org.apache.commons.collections4.multimap.AbstractSetValuedMap;
+import org.apache.commons.lang3.ObjectUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import com.iota.iri.Milestone;
-
 public class TipsManager {
+
+    public static final int MAX_ANCESTORS_SIZE = 10000;
 
     private final Logger log = LoggerFactory.getLogger(TipsManager.class);
     private final Tangle tangle;
@@ -109,15 +116,15 @@ public class TipsManager {
         if(milestone.latestSolidSubtangleMilestoneIndex > Milestone.MILESTONE_START_INDEX ||
                 milestone.latestMilestoneIndex == Milestone.MILESTONE_START_INDEX) {
 
-            Map<Hash, Long> ratings = new HashMap<>();
             Set<Hash> analyzedTips = new HashSet<>();
             Set<Hash> maxDepthOk = new HashSet<>();
             try {
                 Hash tip = entryPoint(reference, extraTip, depth);
-                serialUpdateRatings(visitedHashes, tip, ratings, analyzedTips, extraTip);
+                Map<Hash, Integer> cumulativeWeights = calculateCumulativeWeight(visitedHashes, tip,
+                        extraTip != null, new HashSet<>());
                 analyzedTips.clear();
                 if (ledgerValidator.updateDiff(visitedHashes, diff, tip)) {
-                    return markovChainMonteCarlo(visitedHashes, diff, tip, extraTip, ratings, iterations, milestone.latestSolidSubtangleMilestoneIndex - depth * 2, maxDepthOk, seed);
+                    return markovChainMonteCarlo(visitedHashes, diff, tip, extraTip, cumulativeWeights, iterations, milestone.latestSolidSubtangleMilestoneIndex - depth * 2, maxDepthOk, seed);
                 } else {
                     throw new RuntimeException("starting tip failed consistency check: " + tip.toString());
                 }
@@ -149,11 +156,13 @@ public class TipsManager {
         return milestone.latestSolidSubtangleMilestone;
     }
 
-    Hash markovChainMonteCarlo(final Set<Hash> visitedHashes, final Map<Hash, Long> diff, final Hash tip, final Hash extraTip, final Map<Hash, Long> ratings, final int iterations, final int maxDepth, final Set<Hash> maxDepthOk, final Random seed) throws Exception {
+    Hash markovChainMonteCarlo(final Set<Hash> visitedHashes, final Map<Hash, Long> diff, Hash tip, Hash extraTip, Map<Hash, Integer> cumulativeWeight,
+            int iterations, int maxDepth, Set<Hash> maxDepthOk, Random seed) throws Exception {
         Map<Hash, Integer> monteCarloIntegrations = new HashMap<>();
         Hash tail;
         for(int i = iterations; i-- > 0; ) {
-            tail = randomWalk(visitedHashes, diff, tip, extraTip, ratings, maxDepth, maxDepthOk, seed);
+            tail = randomWalk(visitedHashes, diff, tip, extraTip, cumulativeWeight,
+                    maxDepth, maxDepthOk, seed);
             if(monteCarloIntegrations.containsKey(tail)) {
                 monteCarloIntegrations.put(tail, monteCarloIntegrations.get(tail) + 1);
             } else {
@@ -173,7 +182,25 @@ public class TipsManager {
         }).map(Map.Entry::getKey).orElse(null);
     }
 
-    Hash randomWalk(final Set<Hash> visitedHashes, final Map<Hash, Long> diff, final Hash start, final Hash extraTip, final Map<Hash, Long> ratings, final int maxDepth, final Set<Hash> maxDepthOk, Random rnd) throws Exception {
+    /**
+     * Performs a walk from {@code start} until you reach a tip or {@code extraTip}. The path depends of the values
+     * of transaction weights given in {@code cumulativeWeights}. If a tx weight is missing, then calculate it on
+     * the fly.
+     *
+     * @param visitedHashes hashes of transactions that were validated and their weights can be disregarded when we have
+     *                      {@code extraTip} is not {@code null}.
+     * @param diff map of address to change in balance since last snapshot.
+     * @param start hash of the transaction that starts the walk.
+     * @param extraTip an extra ending point for the walk. If not null the walk will ignore the weights of
+     * {@code visitedHashes}.
+     * @param cumulativeWeights maps transaction hashes to weights. Missing data is computed by this method.
+     * @param maxDepth the transactions we are traversing may not be below this depth measured in number of snapshots.
+     * @param maxDepthOk transaction hashes that we know are not below {@code maxDepth}
+     * @param rnd generates random doubles to make the walk less deterministic
+     * @return a tip's hash
+     * @throws Exception
+     */
+    Hash randomWalk(final Set<Hash> visitedHashes, final Map<Hash, Long> diff, final Hash start, final Hash extraTip, final Map<Hash, Integer> cumulativeWeights, final int maxDepth, final Set<Hash> maxDepthOk, Random rnd) throws Exception {
         Hash tip = start, tail = tip;
         Hash[] tips;
         Set<Hash> tipSet;
@@ -232,17 +259,18 @@ public class TipsManager {
             } else {
                 // walk to the next approver
                 tips = tipSet.toArray(new Hash[tipSet.size()]);
-                if (!ratings.containsKey(tip)) {
-                    serialUpdateRatings(myApprovedHashes, tip, ratings, analyzedTips, extraTip);
+                if (!cumulativeWeights.containsKey(tip)) {
+                    cumulativeWeights.putAll(calculateCumulativeWeight(myApprovedHashes, tip, extraTip != null,
+                            analyzedTips));
                     analyzedTips.clear();
                 }
 
                 walkRatings = new double[tips.length];
                 double maxRating = 0;
-                long tipRating = ratings.get(tip);
+                long tipRating = cumulativeWeights.get(tip);
                 for (int i = 0; i < tips.length; i++) {
                     //transition probability = ((Hx-Hy)^-3)/maxRating
-                    walkRatings[i] = Math.pow(tipRating - ratings.getOrDefault(tips[i],0L), -3);
+                    walkRatings[i] = Math.pow(tipRating - cumulativeWeights.getOrDefault(tips[i],0), -3);
                     maxRating += walkRatings[i];
                 }
                 ratingWeight = rnd.nextDouble() * maxRating;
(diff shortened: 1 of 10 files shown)


src/main/java/com/iota/iri/service/TipsManager.java
```diff
@@ -272,70 +302,161 @@ static long capSum(long a, long b, long max) {
...
+            Set<Hash> analyzedTips) throws Exception {
+        log.info("Start calculating cw starting with tx hash {}", currentTxHash);
+        log.debug("Start topological sort");
+        long start = System.currentTimeMillis();
+        LinkedHashSet<Hash> txHashesToRate = sortTransactionsInTopologicalOrder(currentTxHash);
+        log.debug("Subtangle size: {}", txHashesToRate.size());
+        log.debug("Topological sort done. Start traversing on txs in order and calculate weight");
+        Map<Buffer, Integer> cumulativeWeights = calculateCwInOrder(txHashesToRate, myApprovedHashes, confirmLeftBehind,
+                analyzedTips);
+        log.debug("Cumulative weights calculation done in {} ms", System.currentTimeMillis() - start);
```
alongalky: Is there a better way to measure times in Java? Maybe logs come with built in timestamps anyway? You probably want a profiling of every function call, not only the cumulative weight calculation

src/main/java/com/iota/iri/service/TipsManager.java
```diff
@@ -272,70 +302,161 @@ static long capSum(long a, long b, long max) {
...
+    }
+
+    private LinkedHashSet<Hash>  sortTransactionsInTopologicalOrder(Hash startTx) throws Exception {
+        LinkedHashSet<Hash> sortedTxs = new LinkedHashSet<>();
+        Set<Hash> temporary = new HashSet<>();
+        Deque<Hash> stack = new ArrayDeque<>();
+        Map<Hash, Collection<Hash>> txToDirectApprovers = new HashMap<>();
+
+        stack.push(startTx);
+        while (CollectionUtils.isNotEmpty(stack)) {
```
alongalky: I think `!stack.isEmpty()` is cleaner, and more consistent with the rest of the calls

src/main/java/com/iota/iri/service/TipsManager.java
```diff
@@ -272,70 +302,161 @@ static long capSum(long a, long b, long max) {
...
+        LinkedHashSet<Hash> txHashesToRate = sortTransactionsInTopologicalOrder(currentTxHash);
+        log.debug("Subtangle size: {}", txHashesToRate.size());
+        log.debug("Topological sort done. Start traversing on txs in order and calculate weight");
+        Map<Buffer, Integer> cumulativeWeights = calculateCwInOrder(txHashesToRate, myApprovedHashes, confirmLeftBehind,
+                analyzedTips);
+        log.debug("Cumulative weights calculation done in {} ms", System.currentTimeMillis() - start);
+        return cumulativeWeights;
+    }
+
+    private LinkedHashSet<Hash>  sortTransactionsInTopologicalOrder(Hash startTx) throws Exception {
```
alongalky: What algorithm are you using? Can you add a comment like `// based on DFS algorithm, taken from: https://en.wikipedia.org/wiki/Topological_sorting#Depth-first_search`?

src/main/java/com/iota/iri/service/TipsManager.java
```diff
@@ -272,70 +302,161 @@ static long capSum(long a, long b, long max) {
...
+        Set<Hash> temporary = new HashSet<>();
+        Deque<Hash> stack = new ArrayDeque<>();
+        Map<Hash, Collection<Hash>> txToDirectApprovers = new HashMap<>();
+
+        stack.push(startTx);
+        while (CollectionUtils.isNotEmpty(stack)) {
+            Hash txHash = stack.peek();
+            if (!sortedTxs.contains(txHash)) {
+                Collection<Hash> appHashes = getTxDirectApproversHashes(txHash, txToDirectApprovers);
+                if (CollectionUtils.isNotEmpty(appHashes)) {
```
alongalky: `!appHashes.isEmpty()`

src/test/java/com/iota/iri/service/TipsManagerTest.java
```diff
@@ -65,111 +75,269 @@ public static void tearDown() throws Exception {
     }
 
     @Test
-    public void updateLinearRatingsTestWorks() throws Exception {
-        TransactionViewModel transaction, transaction1, transaction2;
+    public void testCalculateCumulativeWeight() throws Exception {
```
alongalky: Can you make the name a bit more descriptive?

src/test/java/com/iota/iri/service/TipsManagerTest.java
```diff
@@ -65,111 +75,269 @@ public static void tearDown() throws Exception {
...
+                transaction.getHash()), getRandomTransactionHash());
+        transaction2 = new TransactionViewModel(getRandomTransactionWithTrunkAndBranch(transaction.getHash(),
+                transaction.getHash()), getRandomTransactionHash());
+        transaction3 = new TransactionViewModel(getRandomTransactionWithTrunkAndBranch(transaction1.getHash(),
+                transaction2.getHash()), getRandomTransactionHash());
+        transaction.store(tangle);
+        transaction1.store(tangle);
+        transaction2.store(tangle);
+        transaction3.store(tangle);
+        log.debug("printing transaction in diamond shape \n                      {} \n{}  {}\n                      {}",
```
alongalky: Do we need this log in a UT?

src/test/java/com/iota/iri/service/TipsManagerTest.java
```diff
@@ -65,111 +75,269 @@ public static void tearDown() throws Exception {
...
         TransactionViewModel transaction, transaction1, transaction2, transaction3, transaction4;
         transaction = new TransactionViewModel(getRandomTransactionTrits(), getRandomTransactionHash());
-        transaction1 = new TransactionViewModel(getRandomTransactionWithTrunkAndBranch(transaction.getHash(), transaction.getHash()), getRandomTransactionHash());
-        transaction2 = new TransactionViewModel(getRandomTransactionWithTrunkAndBranch(transaction1.getHash(), transaction1.getHash()), getRandomTransactionHash());
-        transaction3 = new TransactionViewModel(getRandomTransactionWithTrunkAndBranch(transaction2.getHash(), transaction1.getHash()), getRandomTransactionHash());
-        transaction4 = new TransactionViewModel(getRandomTransactionWithTrunkAndBranch(transaction2.getHash(), transaction3.getHash()), getRandomTransactionHash());
+        transaction1 = new TransactionViewModel(getRandomTransactionWithTrunkAndBranch(
+                transaction.getHash(), transaction.getHash()), getRandomTransactionHash());
+        transaction2 = new TransactionViewModel(getRandomTransactionWithTrunkAndBranch(
```
alongalky: Maybe it makes sense to factor these out to a function, something like:
```
generateTangle([[1, 0], [2, 1], [3,2])
```
which returns a store with the three different edges and vertices. The code appears in every test and is hard to read, which makes it difficult to understand the scenario that is being tested

src/test/java/com/iota/iri/service/TipsManagerTest.java
```diff
@@ -65,111 +75,269 @@ public static void tearDown() throws Exception {
...
         transaction2.store(tangle);
         transaction3.store(tangle);
         transaction4.store(tangle);
-        Map<Hash, Set<Hash>> ratings = new HashMap<>();
-        tipsManager.updateHashRatings(transaction.getHash(), ratings, new HashSet<>());
-        Assert.assertEquals(ratings.get(transaction.getHash()).size(), 5);
-        Assert.assertEquals(ratings.get(transaction1.getHash()).size(),4);
-        Assert.assertEquals(ratings.get(transaction2.getHash()).size(), 3);
+
+        log.info(String.format("Linear ordered hashes from tip %.4s, %.4s, %.4s, %.4s, %.4s", transaction4.getHash(),
```
alongalky: log in test?

src/test/java/com/iota/iri/service/TipsManagerTest.java
```diff
@@ -65,111 +75,269 @@ public static void tearDown() throws Exception {
...
+        Assert.assertEquals(String.format(TX_CUMULATIVE_WEIGHT_IS_NOT_AS_EXPECTED_FORMAT, 1),
+                4, txToCw.get(transaction1.getHash().getSubHash()).intValue());
+        Assert.assertEquals(String.format(TX_CUMULATIVE_WEIGHT_IS_NOT_AS_EXPECTED_FORMAT, 0),
+                5, txToCw.get(transaction.getHash().getSubHash()).intValue());
     }
 
     @Test
-    public void updateRatings2TestWorks() throws Exception {
-        TransactionViewModel transaction, transaction1, transaction2, transaction3, transaction4;
+    public void testCalculateCumulativeWeightAlon() throws Exception {
```
alongalky: Which Alon? :)
I think the name needs to say a bit more

src/test/java/com/iota/iri/service/TipsManagerTest.java
```diff
@@ -65,111 +75,269 @@ public static void tearDown() throws Exception {
...
+        String msg = "Cumulative weight is wrong for tx";
+        Assert.assertEquals(msg + 4, 1, cumulativeWeight.get(transaction4.getHash().getSubHash()).intValue());
+        Assert.assertEquals(msg + 3, 1, cumulativeWeight.get(transaction3.getHash().getSubHash()).intValue());
+        Assert.assertEquals(msg + 2, 1, cumulativeWeight.get(transaction2.getHash().getSubHash()).intValue());
+        Assert.assertEquals(msg + 1, 2, cumulativeWeight.get(transaction1.getHash().getSubHash()).intValue());
+        Assert.assertEquals(msg + 0, 3, cumulativeWeight.get(transaction.getHash().getSubHash()).intValue());
     }
 
-    //@Test
+    //    @Test
```
alongalky: I recommend removing changes to this function, if it's only formatting and comments

src/main/java/com/iota/iri/service/TipsManager.java
alongalky: Also, maybe it makes sense to take this function out of the module so you can write unit tests for it

src/main/java/com/iota/iri/service/TipsManager.java
```diff
@@ -1,21 +1,32 @@
...
 import org.slf4j.LoggerFactory;
 
-import com.iota.iri.Milestone;
+import java.nio.Buffer;
+import java.nio.ByteBuffer;
+import java.util.*;
 
 public class TipsManager {
 
+    public static final int MAX_ANCESTORS_SIZE = 1000;
```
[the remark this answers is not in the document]
GalRogozinski: This is the max size of the ancestor set. It is limited to reduce the chance of `OutOfMemory` exception.
@alon-e and I agreed on it.

Is it the correct number to use?
Frankly I don't know.
If @paulhandy or @th0br0 have something to say about this figure then I am all ears.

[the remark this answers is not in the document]
alongalky: Is there a memory consumption estimation? This will tell us at least if we are in the ball park or being overly permissive or restrictive.
It could even be a back-of-the-envelope type calculation: 
1000 transactions *
size of the subhash *
expected walk depth *
reasonable number of transactions between milestones (say with 100 tx/sec)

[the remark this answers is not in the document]
alongalky: In addition, what happens when there the lists are full? It seems silly to keep allocating ancestor sets, you should just somehow pass "MAX" to save memory and time.

src/main/java/com/iota/iri/service/TipsManager.java
GalRogozinski: The idea be`CollectionUtils` is to be defensive and fall for `NullPointerException`.

You can say that in this specific code segment `stack` will never be `null`, and even if it can the above line (`stack.push()`) will throw an exception.

However, the reason I use it is purely out of habit, which I believe to be a good one.
If wherever it is fine to treat a `null` collection like an empty colllection one will use a null safe method then we will get less pesky null pointer exceptions.

src/main/java/com/iota/iri/service/TipsManager.java
GalRogozinski: I can add a comment.

A unit test may be a good idea, but currently it is hard to write.
This is because Topological Order is not deterministic...

Thinking about this a little more now,
I can do a diamond shaped graph with 4 vertices. It will have only two possible orders and I can test whether one of them occurs.
If you have an idea of how to better test this please tell.

Another caveat (unrelated to your comment) to think about is the use of `LinkedHashMap`.
There was an earlier version of the code where the the `continue` in L#356 was erroneously omitted. The result was that there were multiple calls to `add`. In that early version `List` and `Set` were used instead of `LinkedHashMap`, so the bug was easily found. However if that `continue` will be deleted now that bug will not be so easy to see. The method will return the correct output but will be slower due to excessive `add` calls.

The advantage of using `LinkedHashMap` is that it will be more efficient on memory, than 2 objects. This method by the way *is* a bottle-neck memory-wise.

src/main/java/com/iota/iri/service/TipsManager.java
GalRogozinski: defensive programming

src/test/java/com/iota/iri/service/TipsManagerTest.java
GalRogozinski: I think it helps when the test fails

src/test/java/com/iota/iri/service/TipsManagerTest.java
GalRogozinski: It is a unit test with a tangle @alon-e made up to understand the PR better. I have no idea what name to give it, honestly.

Any name you can recommend will be fine.

alongalky: Something to describe what it does.

src/main/java/com/iota/iri/service/TipsManager.java
```diff
@@ -232,17 +259,20 @@ Hash randomWalk(final Set<Hash> visitedHashes, final Map<Hash, Long> diff, final
...
                 walkRatings = new double[tips.length];
                 double maxRating = 0;
-                long tipRating = ratings.get(tip);
+                ByteBuffer subHash = tip.getSubHash();
+                long tipRating = cumulativeWeights.get(subHash);
                 for (int i = 0; i < tips.length; i++) {
+                    subHash = tips[i].getSubHash();
                     //transition probability = ((Hx-Hy)^-3)/maxRating
-                    walkRatings[i] = Math.pow(tipRating - ratings.getOrDefault(tips[i],0L), -3);
+                    walkRatings[i] = Math.pow(tipRating - cumulativeWeights.getOrDefault(subHash,0), -3);
```
olaf-2: Performance could be improved by a factor of 20+. See https://github.com/iotaledger/iri/pull/535

src/main/java/com/iota/iri/service/TipsManager.java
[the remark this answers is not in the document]
GalRogozinski: hmmm

Let say the size of a Subhash is 32 bytes. 1000 txs => ~32kb per set.
The total amount of memory 32kb * num_of_unreleased_sets.
This depends on how wide the tangle is...

src/main/java/com/iota/iri/service/TipsManager.java
alongalky: I recommend leaving this change out of this PR, it's big enough as it is. It's an unrelated issue.

src/test/java/com/iota/iri/service/TipsManagerTest.java
GalRogozinski: In some point in the future I will rewrite the test to not use random hashes (It is now this way because this is how it was before). Then I will also add this fix.

src/test/java/com/iota/iri/service/TipsManagerTest.java
GalRogozinski: I think it is helpful to log. If others will agree with you I am fine with removing

src/main/java/com/iota/iri/service/TipsManager.java
GalRogozinski: We are supposed to switch to the formula described in the whitepaper (with the alpha). If it won't happen soon then we will merge your PR.

src/main/java/com/iota/iri/service/TipsManager.java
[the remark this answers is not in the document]
GalRogozinski: Will check on testnet how often we hit the hard limit + make it configurable

PR #562, opened 2018-02-28
simplified code when adding neighbor + removal of strange condition in replyToRequest + changed FifoHashCache datastructure

gjeee:
2 small changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ It allows to connect easily using java directly to a local or a remote [[IOTA no
 
 # How to get started
 
-Obviously, because this is its own, independent network, you have to go through the same process as in the main network: **find neighbors**. You can find neighbors in the `#nodesharing` Slack channel[[Slack Invite]](http://slack.iota.org), or on our forum. Community members are usually very happy to help you out and get you connected. If you want to get tokens for your testcase, please just ask in one of the communication channels as well.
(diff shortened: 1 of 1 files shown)


src/main/java/com/iota/iri/network/Node.java
```diff
@@ -410,7 +410,7 @@ public void replyToRequest(Hash requestedHash, Neighbor neighbor) {
         if (requestedHash.equals(Hash.NULL_HASH)) {
             //Random Tip Request
             try {
-                if (transactionRequester.numberOfTransactionsToRequest() > 0 && rnd.nextDouble() < P_REPLY_RANDOM_TIP) {
+                if (rnd.nextDouble() < P_REPLY_RANDOM_TIP) {
```
GalRogozinski: This change requires a little explanation

gjeee: well as far as i know the transactionRequester is responsible for fetching tx which are unknown to the node. this piece of code just needs to reply to the neighbor with a random tip (from tipsViewModel). so i do not see the need of a positive numberOfTransactionsToRequest when replying.

gjeee:
it can not compile due to missing[ apache lib](https://commons.apache.org/proper/commons-collections/apidocs/org/apache/commons/collections4/set/ListOrderedSet.html)...maybe you should pull this as well?

Issue #566, opened 2018-03-04
Add IP address to getNeighbors response


cwarner818:
There are some systems out there that have a single dns entry pointed at multiple nodes. It would be beneficial to add the actual IP address of the connected node to the `getNeighbors` response so that users will know specifically which node they are connected to. I've also thought that instead of having an `address` field, we should perhaps return `hostname`, `ipAddress`, and `port` to really break down the information, although just adding an `ipAddress` field would be great for now. 

I looked in to doing this myself but Java is not in my toolbox and I was quickly lost. I don't think this would be too hard for someone that actually knew what they were doing though.

PR #567, opened 2018-03-04
Add neighbor IP to GetNeighbors

cwarner818:
This is a PR for #566 with code from microhash on discord (thank you!) 

This adds an IP address field to the `getNeighbors` response so that systems that use round-robin DNS can still be debugged on a per-node level. Without this patch its impossible to know which specific node you are peered with because `getNeighbors` just returns the hostname. 

There are a few of these systems out there already, andI have seen then with up to 12 nodes being served under one DNS entry.


GalRogozinski:
Did you test how the different libraries parse the new output?

cwarner818:
I have tested it against giota on [my own node](https://github.com/tangleengineering/iri/commit/7dc5d840d0b98016dac0e47bd3fec2df9e4d41db) as that is what I use, but in reality it shouldn't cause any issues with any other libraries as long as they dont panic on new JSON fields.

src/main/java/com/iota/iri/service/dto/GetNeighborsResponse.java
```diff
@@ -44,6 +44,7 @@ public static Neighbor createFrom(com.iota.iri.network.Neighbor n) {
             Neighbor ne = new Neighbor();
             int port = n.getPort();
             ne.address = n.getAddress().getHostString() + ":" + port;
+            ne.ip = n.getAddress().getAddress().toString().split("/")[1];
```
GalRogozinski: Please put `n.getAddress()` in a local variable, and use for the `ne.address`, `ne.ip` assignment.
This is because:
1. I don't like the concatenated double `getAddress()`.
2. In general, concatenated gets make it a bit harder to pinpoint null pointer exceptions.

mikrohash:
@cwarner818 please change it to this:

    public static Neighbor createFrom(com.iota.iri.network.Neighbor n) {
        Neighbor ne = new Neighbor();
        java.net.InetSocketAddress address = n.getAddress();
        int port = n.getPort();
        ne.address = address.getHostString() + ":" + port;
        ne.ip = address.getAddress().getHostAddress();
        ...
    }

cwarner818:
Will do, just haven't been on a computer much this week. Expect to get to
it this weekend.

mikrohash:
@cwarner818 you forgot to change this line:

    ne.ip = address.getAddress().getHostAddress();

cwarner818:
Fixed, thanks for pointing that at. No more early morning commits for me. @GalRogozinski change has been made.

src/main/java/com/iota/iri/service/dto/GetNeighborsResponse.java
```diff
@@ -42,8 +42,10 @@ public String getConnectionType() {
 
         public static Neighbor createFrom(com.iota.iri.network.Neighbor n) {
             Neighbor ne = new Neighbor();
+            java.net.InetSocketAddress address = n.getAddress();
```
alon-e: there is a `Neighbor` field also named `address` which is a string in the same file above - better to change the name to avoid confusion

mikrohash: @cwarner818 

    public static Neighbor createFrom(com.iota.iri.network.Neighbor n) {
        Neighbor ne = new Neighbor();
        java.net.InetSocketAddress socketAddress = n.getAddress();
        int port = n.getPort();
        ne.address = socketAddress.getHostString() + ":" + port;
        ne.ip = socketAddress.getAddress().getHostAddress();
        ...
    }

src/main/java/com/iota/iri/service/dto/GetNeighborsResponse.java
```diff
@@ -42,8 +42,10 @@ public String getConnectionType() {
 
         public static Neighbor createFrom(com.iota.iri.network.Neighbor n) {
             Neighbor ne = new Neighbor();
+            java.net.InetSocketAddress socketAddress = n.getAddress();
```
GalRogozinski: could you please use an `import` statement instead of using the fully qualified name?

mikrohash: @cwarner818 please add `import java.net.InetSocketAddress;` at the head of the file and change line 45 from `java.net.InetSocketAddress socketAddress = n.getAddress();` to `InetSocketAddress socketAddress = n.getAddress();`

cwarner818:
@GalRogozinski requested change has been made, sorry for the delay!

GalRogozinski:
Too old, became irrelevant now

PR #568, opened 2018-03-05
Added link for nelson

payal-kothari:


GalRogozinski:
Nelson is a prominent community project, but it is not part of the foundation's development of the core protocol. 
So we cannot accept this PR.

PR #569, opened 2018-03-05
zmq: only write to socket from one thread

th0br0:
zeromq is not thread safe to use.
Hence, the Socket should only be used from within one thread.

### DIFF
--- a/src/main/java/com/iota/iri/zmq/MessageQ.java
+++ b/src/main/java/com/iota/iri/zmq/MessageQ.java
@@ -2,8 +2,9 @@ package com.iota.iri.zmq;
 
 import org.zeromq.ZMQ;
 
-import java.text.MessageFormat;
-import java.util.Objects;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
 
 /**
  * Created by paul on 6/20/17.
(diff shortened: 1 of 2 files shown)


th0br0:
Both internal & community users reported no more ZMQ outages.

src/main/java/com/iota/iri/zmq/MessageQ.java
```diff
@@ -29,12 +32,19 @@ public MessageQ(int port, String ipc, int nthreads, boolean enabled) {
...
         }
     }
 
-    public void shutdown () {
+    public void shutdown() {
+        publisherService.shutdown();
+        try {
+            publisherService.awaitTermination(5, TimeUnit.SECONDS);
+        } catch (InterruptedException e) {
+            e.printStackTrace();
```
GalRogozinski: Shouldn't this be printed with a logger?

src/main/java/com/iota/iri/zmq/MessageQ.java
```diff
@@ -29,12 +32,19 @@ public MessageQ(int port, String ipc, int nthreads, boolean enabled) {
...
+        }
         publisher.close();
```
GalRogozinski: Since there is a `try` I find it more readable to put the last two lines in a `finally` block.

PR #570, opened 2018-03-05
Checkstyle will fail build

GalRogozinski:
This will cause Travis CI to fail due to checkstyle errors enforced by the rules in `iota-checks.xml`.
Currently only braces around code blocks are enforced (see http://checkstyle.sourceforge.net/config_blocks.html#NeedBraces).

Rules will be added incrementally.

Note that `mvn site` will still create a checkstyle report using the default `sun-checks.xml`. This is because the report with `iota-checks.xml` will  always be empty on successful builds. If we think the report is redundant we can remove it.

### DIFF
--- a/iota-checks.xml
+++ b/iota-checks.xml
@@ -1,0 +1,22 @@
+<?xml version="1.0"?>
+<!DOCTYPE module PUBLIC
+          "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
+          "http://checkstyle.sourceforge.net/dtds/configuration_1_3.dtd">
+<!--
+
+IOTA checkstyle draft
+
+-->
+<module name="Checker">
+    <!--
+        If you set the basedir property below, then all reported file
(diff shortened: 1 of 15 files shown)


iota-checks.xml
```diff
@@ -0,0 +1,22 @@
...
+    <!--
+        If you set the basedir property below, then all reported file
+        names will be relative to the specified directory. See
+        http://checkstyle.sourceforge.net/5.x/config.html#Checker
+
+        <property name="basedir" value="${basedir}"/>
+    -->
+    <property name="fileExtensions" value="java, properties, xml"/>
+    <module name="TreeWalker">
+        <module name="NeedBraces"/>
```
th0br0: We should also use `<module name="AvoidInlineConditionals"/>`

GalRogozinski: @th0br0 I can't naively make this change without producing checkstyle errors.
I opened GalRogozinski#3 to make sure this change will happen.

Issue #565, opened 2018-03-02
Capacity FifoHashCache is breached!


gjeee:
TipsViewModel  uses 2 FifoHashCache objects called tips and solidTips with a capacity of 5000 each. So the total tipcount should be never more than 10K. However my own node shows a tipcount (solid+non-solid) of more than 10000. A lot of other nodes as well as can be seen on https://iotanode.host

So somehow the add-routine of [FifoHashCache](https://github.com/iotaledger/iri/blob/6433963b4b59e61fce093e308e5271b50d110469/src/main/java/com/iota/iri/controllers/TipsViewModel.java) is not working as intended.

GalRogozinski:
Also `com.iota.iri.controllers.TipsViewModel.FifoHashCache#addAll` is no good since it ignore `capacity`. This isn't the problem because no one is calling it, but it also should be fixed in this issue.

gjeee:
Indeed i also noticed that #addAll issue. Apart from this i have been experimenting with ListOrderedSet (from org.apache.commons.collections4) instead of LinkedHashSet. Using this datastructure did not breach the capacity. Moreover you can randomly access elements since it has a get(index) method which is very useful for  the getRandomSolidTipHash() routine. However this routine does throw now and then IndexOutOfBounds exceptions (caught in "Tip Solidity Rescan" thread) but i have not been able yet to track down why. There are locks in place so it is not a thread thing (????).

GalRogozinski:
Did you understand why the change from LinkedHashSet solves the capacity issue?
It does seem like all methods are synchronized as far as I can tell.

GalRogozinski:
It is apparent that Hash#Equals() and Hash#HashCode() are broken...

gjeee:
if the Hash#Equals() is broken basically means the hashset can contain duplicates....but how does that explain that capacity is breached? it should still contain  <=5000

gjeee:
somehow i am now unable to reproduce the findings. maybe it only happens in a relatively high-load environment.

gjeee:
![image](https://user-images.githubusercontent.com/13155183/37003653-1ca9954e-20ce-11e8-8d0b-695d4a2a2743.png)

i was able to reproduce findings again!!! see attached image: the left-number is size non-solid tips, right number is total tips. 

it happens with dev branch + patch 470 . Probably patch 470 triggers the assertion in Hash#Equals() .

i will try once more with just the dev branch.

GalRogozinski:
[...]

PR #575, opened 2018-03-10
Thread-safe Hash implementation: fix for issue/565

gjeee:
I changed the Hash implementation. I made sure that all the constructors read all the state vars in 1 go in a synchronised fashion. It fixes issue 565 https://github.com/iotaledger/iri/issues/565

### DIFF
--- a/src/main/java/com/iota/iri/model/Hash.java
+++ b/src/main/java/com/iota/iri/model/Hash.java
@@ -8,6 +8,8 @@ import com.iota.iri.utils.Converter;
 
 import java.io.Serializable;
 import java.util.Arrays;
+import java.util.Objects;
+ 
 
 public class Hash implements Serializable, Indexable {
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/iota/iri/model/Hash.java
```diff
@@ -19,33 +21,56 @@
     private byte[] bytes;
     private int[] trits;
```
gubatron: why not just make these 2 guys final, and use new instances whenever you need a Hash, nothing to put synchronized sections on, just make a new copy.

In my experience, synchronized should be a last resource. Try to work it functional, no mutability.
All those synchronized blocks most likely will end up creating deadlock headaches.

src/main/java/com/iota/iri/model/Hash.java
```diff
@@ -19,33 +21,56 @@
...
     private int hashCode;
+
```
gubatron: have you tried making these 3 properties final, calculated when the object is created.
The best way to make code thread safe is to make it immutable. If you need to change it in any way, just create a new instance, the functional approach is way more robust and easier to parallelize.

In my experience, `synchronized` should be a last resource and usually a sign of something bad in design. 
Try to work it functionally, with no mutability.
All those synchronized blocks most likely will end up creating deadlock headaches.

gjeee: hi tx for your suggestion.....

like u said maybe setting all properties in the constructor and make them final can make this whole class thread-safe.....of course there is no lazy initialization but at least coding-wise a lot simpler....

but there is already another PR as a possible fix https://github.com/iotaledger/iri/pull/577
..... it features lazy initialization but it still needs some sync blocks though....so all in all a bit more complex

gubatron: wish I had more time to try and code this myself immutably, perhaps a few cascading refactors might show weaknesses in the architecture

GalRogozinski:
Close this PR in favor of #577

PR #577, opened 2018-03-14
Made `Hash` thread-safe, fixed `compareTo` with an identity check, and ...

footloosejava:
… some minor edits with regards to layout

(1) Made Hash thread-safe without needing full synchronization.

All access to `bytes`, `trits` and `hashcode` done via a null check against an inner class object that is instantiated in a synchronized block with a double check. `ByteSafe` and `TritSafe` are inner classes and the data hidden in them is not available until the classes are fully instantiated (classloading ensures this).

(2)  The (1) above fixes the race condition to `hashcode()` and `equals()` and `bytes()` that was proven to be happening and might also have cause the FIFO cache overrun.

(3) Fixed `compareTo` to do an identity check and not need to make a new `Hash` every time - because Hash extends Indexable and the the indexible provided could already be a Hash.

(4) Commented on bounds check. This is easy to fix but so many tests are abusing it and they won't pass until the tests are also fixed.

(5) Added a null check in `fullread' because the RocksDB does provide null byte[] to read methods sometimes and they can be safely passed over.

(6) Added a 2nd time `read` persistence check. The class should be effectively immutable (not possible) but at least check to see if it has already been initialized and reject a second initialization.

(7) No other classes extend `Hash` . I have made it `final` and that can be changed in the future if it ever needs to be extended.

All tests pass.
Live run passed.

### DIFF
--- a/src/main/java/com/iota/iri/model/Hash.java
+++ b/src/main/java/com/iota/iri/model/Hash.java
@@ -8,51 +8,39 @@ import com.iota.iri.utils.Converter;
 
 import java.io.Serializable;
 import java.util.Arrays;
+import java.util.Objects;
 
-public class Hash implements Serializable, Indexable {
+public final class Hash implements Serializable, Indexable {
 
     public static final int SIZE_IN_TRITS = 243;
     public static final int SIZE_IN_BYTES = 49;
 
     public static final Hash NULL_HASH = new Hash(new int[Curl.HASH_LENGTH]);
 
-    private byte[] bytes;
-    private int[] trits;
-    private int hashCode;
-    
-    // constructors' bill
+    private final Object lock = new Object();
+    private ByteSafe byteSafe;
+    private TritSafe tritSafe;
 
-    public Hash(final byte[] bytes, final int offset, final int size) {
-        fullRead(bytes, offset, size);
+    private final class ByteSafe {
+        private final byte[] bytes;
+        private final Integer hashcode;
+
+        private ByteSafe(byte[] bytes) {
+            Objects.requireNonNull(bytes, "ByteSafe is attempted to be initialized with a null byte array");
+            this.bytes = bytes;
+            this.hashcode = Arrays.hashCode(bytes);
+        }
     }
 
-    public Hash(){}
+    private final class TritSafe {
+        private final int[] trits;
 
-    public Hash(final byte[] bytes) {
-        this(bytes, 0, SIZE_IN_BYTES);
+        private TritSafe(int[] trits) {
+            this.trits = Objects.requireNonNull(trits, "TritSafe is attempted to be initialized with a null int array");
+        }
     }
 
-    public Hash(final int[] trits, final int offset) {
-        this.trits = new int[SIZE_IN_TRITS];
-        System.arraycopy(trits, offset, this.trits, 0, SIZE_IN_TRITS);
-        //this(Converter.bytes(trits, offset, trits.length));
-    }
 
-    public Hash(final int[] trits) {
-        this(trits, 0);
-    }
-
-    public Hash(final String trytes) {
-        this.trits = new int[SIZE_IN_TRITS];
-        Converter.trits(trytes, this.trits, 0);
-    }
-
-    //
-    /*
-    public static Hash calculate(byte[] bytes) {
-        return calculate(bytes, SIZE_IN_TRITS, new Curl());
-    }
-    */
     public static Hash calculate(SpongeFactory.Mode mode, int[] trits) {
         return calculate(trits, 0, trits.length, SpongeFactory.create(mode));
     }
@@ -62,6 +50,7 @@ public class Hash implements Serializable, Indexable {
         Converter.getTrits(bytes, trits);
         return calculate(trits, 0, tritsLength, curl);
     }
+
     public static Hash calculate(final int[] tritsToCalculate, int offset, int length, final Sponge curl) {
         int[] hashTrits = new int[SIZE_IN_TRITS];
         curl.reset();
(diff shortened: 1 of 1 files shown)


src/main/java/com/iota/iri/model/Hash.java
```diff
@@ -10,85 +10,125 @@
...
         return zeros;
     }
 
     public int[] trits() {
-        if(trits == null) {
-            trits = new int[Curl.HASH_LENGTH];
-            Converter.getTrits(bytes, trits);
+        TritSafe safe = tritSafe;
+        if (safe == null) {
+            synchronized (lock) {
```
GalRogozinski: Good implementation double-checked locking!
Here is a link for others who may inspecting the code:
https://en.wikipedia.org/wiki/Double-checked_locking#Usage_in_Java

src/main/java/com/iota/iri/model/Hash.java
```diff
@@ -10,85 +10,125 @@
...
+                if (tritSafe == null) {
+                    Objects.requireNonNull(byteSafe, "I need my bytes to be initialized in order to construct trits.");
```
GalRogozinski: I like the error message ;-)

src/main/java/com/iota/iri/zmq/MessageQ.java
```diff
@@ -51,7 +51,7 @@ public void shutdown() {
             LOG.error("Publisher service shutdown failed.", e);
         }
 
-        publisher.close();
-        context.term();
+        if(publisher != null) publisher.close();
```
GalRogozinski: Please use brackets and spaces:
```
if (publisher != null) {
    publisher.close();
}
```
This is because #570 that should be merged soon.

src/main/java/com/iota/iri/zmq/MessageQ.java
```diff
@@ -51,7 +51,7 @@ public void shutdown() {
...
+        if(context != null) context.term();
```
GalRogozinski: ditto

src/main/java/com/iota/iri/model/Hash.java
```diff
@@ -10,85 +10,125 @@
...
+        private final int[] trits;
+
+        private TritSafe(int[] trits) {
+            this.trits = Objects.requireNonNull(trits);
+        }
     }
 
-    public Hash(final int[] trits) {
-        this(trits, 0);
+    public Hash() {
```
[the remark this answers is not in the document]
footloosejava: A default no-args constructor is required for instantiation in the RocksDB marshalling where classes of Persistable and Indexable (of which Hash is) are instantiated using clazz.newInstance(). They are then populated with data from a byte[]. I think that is why.

src/main/java/com/iota/iri/model/Hash.java
```diff
@@ -10,85 +10,125 @@
...
+    public Hash(final byte[] bytes) {
+        this(bytes, 0, SIZE_IN_BYTES);
     }
-    */
-    public static Hash calculate(SpongeFactory.Mode mode, int[] trits) {
-        return calculate(trits, 0, trits.length, SpongeFactory.create(mode));
+
+    public Hash(final int[] trits, final int offset) {
+        int[] dest = new int[SIZE_IN_TRITS];
+        System.arraycopy(trits, offset, dest, 0, dest.length);
```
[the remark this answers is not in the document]
footloosejava: Agreed!

src/main/java/com/iota/iri/model/Hash.java
```diff
@@ -105,35 +145,38 @@ public boolean equals(Object o) {
 
     @Override
     public int hashCode() {
-        if(bytes == null) {
+        if (byteSafe == null) {
```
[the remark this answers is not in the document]
footloosejava: I suppose. But it does save a theoretical method call and unneeded byte[] return value. Purely speculative. It may make no difference either way.

footloosejava:
All feedback incorporated into the codebase. Thanks for your comments everyone!

src/main/java/com/iota/iri/model/Hash.java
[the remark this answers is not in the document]
footloosejava: The good news is ... the System.arraycopy does bounds checks and will fail fast with some obscure message. It is bad style ... but it will survive thanks to downstream code.

footloosejava:
Travis CI reports a problem with 'Objects' variable. Huh? I think Travis needs coffee... There is no bug.

GalRogozinski:
@footloosejava 
It was a tricky one.
Rebase your local branch on `dev`, you will notice you have a missing import statement.
Fix it quickly and push and we will approve.

footloosejava:
Did I do that right? I see Green check marks :)

src/main/java/com/iota/iri/model/Hash.java
```diff
@@ -147,7 +187,8 @@ public Indexable decremented() {
 
     @Override
     public int compareTo(Indexable indexable) {
-        Hash hash = new Hash(indexable.bytes());
+        Objects.requireNonNull(indexable);
```
GalRogozinski: Just one last thing (didn't notice it before)...
Unlike the other places where you rightfully added an early null fail check, here it seems extraneous.
The next line will throw the exact same exception and it will be very clear that `indexable` will be the one to blame.

Java is a rather verbose language, but I don't think there is a need to over do it.

footloosejava:
It is needed because of the following line that creates a new Hash from the Indexable byte[] - if the `indexable` provided is found to not be a Hash instance. I assume that anywhere that an actual null is provided to a `Hash compareTo(...)` would be considered a programming error. 

After the instanceof check, the indexable has a method called. I can make this accept a null `Indexable' if you want. Not advised though because collections with null `Hash` references are probably bugs.

```
    @Override
    public int compareTo(Indexable indexable) {
        Objects.requireNonNull(indexable, "parameter 'indexable' cannot be null");
        Hash hash = (indexable instanceof Hash) ? (Hash) indexable : new Hash(indexable.bytes());
        if (this.equals(hash)) {
            return 0;
        }
        long diff = Converter.longValue(hash.trits(), 0, SIZE_IN_TRITS) - Converter.longValue(trits(), 0, SIZE_IN_TRITS);
        if (Math.abs(diff) > Integer.MAX_VALUE) {
            return diff > 0L ? Integer.MAX_VALUE : Integer.MIN_VALUE + 1;
        }
        return (int) diff;
    }
```

GalRogozinski:
Hey @footloosejava,

1. Your change is now running in the testnet (with the last change you agreed on, but didn't yet push). In the not so far future we will have a process that will let us merge PRs far more quickly to dev.

2. We would like to speak to you in private. Please contact us at alon@iota.org, or gal.rogozinski@iota.org.

footloosejava:
@GalRogozinski  @alon-e I saw your message and just got your email. I'll be in touch shortly.

footloosejava:
@th0br0 @GalRogozinski and the rest of the team. I have added a series of new individual commits in a new PR to clean up some of the code base and fix a few minor issues. Each commit has its own accompanying comments describing the changes and reasons. I hope you find everything in order.  I look forward to your comments.

GalRogozinski:
@footloosejava 
Thanks for trying to improve IOTA.

It is a bit of problem that you are pushing unrelated changes to an already reviewed PR.
Please perform the following steps:
1. For each change that has a defined area of responsibility create a new git branch.
2. Cherry pick the relevant commits to their corresponding branches.
3. On the PR's original branch do a `git reset --hard` to the commit with the latest change of the PR's original purpose. Push with `git push -f`.
4. Create new PRs with the new branches.

Thanks

footloosejava:
No problem - I'll try that.

footloosejava:
Alright! New PR for the rest as requested.

GalRogozinski:
Thanks,

When I look at the Files Changed list of this PR I still see changes from #593 and vice versa.
I see that your other PRs also have commits that came from previous PRs.

This makes reviewing and merging them a bit hard.
There are 2 options:
1. If the changes in each PR is dependent on the changes of the previous PR (as far as I can tell they are mostly not, but I didn't look close enough yet), explain why they are in a comment here.
2. If they are independent (as I believe they are) please perform the following for each PR branch:

1. `git rebase -i HEAD~[# of commits you want to look back at]`
2.  In the editor that opens up mark the unneeded commits with `d` to delete them.
3. `git push -f`.

If you have any troubles with git (many experienced people do) feel free to contact me via email and I will help you out.
If you are a git master just ignore the above line :-)

footloosejava:
@GalRogozinski I think I may have done something right :-)
Thanks for your offer for help - I almost took you up on that offer!

PR #590, opened 2018-03-26
Node.java cleanup

footloosejava:
- moved permanent settings from init() to constructor and made final.
- Removed static modifier for instance variables and moved all static vars to top
- re-timed threads for ongoing tasks
- used direct Threads with Daemon status to jobs (no need for executor)
- added locking on separate FINAL mutex fields.
- cleaned up parsing of neighbor's configuration and allowed for multiple spaces and also handling for commas if present.
- added P_DROP check to see if property warrants doing a call to rnd.nextDouble().
- added some commenting to datagram copy and send routines.
- and a few minor things.
-removed static modifier on Node.rejectedAddresses since neighbors is instance level, so should rejected addresses. *required one change in Replicator to make a ref use instance of Node instead of static method.

### DIFF
--- a/src/main/java/com/iota/iri/network/Node.java
+++ b/src/main/java/com/iota/iri/network/Node.java
@@ -21,40 +21,47 @@ import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.security.SecureRandom;
 import java.util.*;
-import java.util.concurrent.*;
+import java.util.concurrent.ConcurrentSkipListSet;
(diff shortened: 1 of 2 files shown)


footloosejava:
This needs the changes from the ReplicatorSourceProcessor PR to build properly.

GalRogozinski:
I am reopening this PR so when we refactor this class we won't forget about these changes

GalRogozinski:
Not relevant due to #1072

PR #591, opened 2018-03-26
TipsViewModel.java minor edits

footloosejava:
- Fixed Generic Type to remove infraction warning during  build.
- Tightened access modifiers.
- Made cache private class with final properties
- Removed unused returns in 2 methods.
(access can be broadened in the future if needed again)

### DIFF
--- a/src/main/java/com/iota/iri/controllers/TipsViewModel.java
+++ b/src/main/java/com/iota/iri/controllers/TipsViewModel.java
@@ -7,11 +7,7 @@ import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.Set;
-import java.util.concurrent.ExecutionException;
 
-/**
- * Created by paul on 3/14/17 for iri-testnet.
- */
 public class TipsViewModel {
 
     // THIS NEEDS TO BE PACKAGE LEVEL ACCESS FOR TESTS ONLY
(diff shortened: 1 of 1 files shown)


src/main/java/com/iota/iri/controllers/TipsViewModel.java
```diff
@@ -1,44 +1,45 @@
...
-    private SecureRandom seed = new SecureRandom();
-    public final Object sync = new Object();
+    private final FifoHashCache<Hash> tips = new FifoHashCache<>(TipsViewModel.MAX_TIPS);
+    private final FifoHashCache<Hash> solidTips = new FifoHashCache<>(TipsViewModel.MAX_TIPS);
 
-    public boolean addTipHash (Hash hash) throws ExecutionException, InterruptedException {
+    private final SecureRandom seed = new SecureRandom();
+    private final Object sync = new Object();
+
+    public void addTipHash(Hash hash) throws ExecutionException, InterruptedException {
```
GalRogozinski: What are the advantages of returning `void` rather than `boolean`?
In my opinion if you can find something to return in a method other than void you should always do it.

Maybe in the future you will use it maybe not.

src/main/java/com/iota/iri/controllers/TipsViewModel.java
```diff
@@ -1,44 +1,45 @@
...
         synchronized (sync) {
-            return tips.add(hash);
+            tips.add(hash);
         }
     }
 
-    public boolean removeTipHash (Hash hash) throws ExecutionException, InterruptedException {
+    public void removeTipHash(Hash hash) throws ExecutionException, InterruptedException {
```
GalRogozinski: ditto

src/main/java/com/iota/iri/controllers/TipsViewModel.java
footloosejava: What should the return value should mean for these two methods. In `addTipHash`, it will be added but may return false it it was already present. In the last method, does true mean that the operation succeeded or that it was removed and added and was not already added?

```
 /**
     * @param hash The tip to add.
     * @return true If the {hash} was added to {@code tips} and {@code tips} did not already contain it.
     */
    public boolean addTipHash(Hash hash) {
        synchronized (sync) {
            return tips.add(hash);
        }
    }

    /**
     * This removes the tip from {@code tips} and adds it to {@code solidTips}.
     *
     * @param tip The tip to set solid.
     * @return true If the tip was removed from {@code tips} and added to {@code solidTips}.
     */
    public boolean setSolid(Hash tip) {
        synchronized (sync) {
            return tips.remove(tip) && solidTips.add(tip);
        }
    }
```

GalRogozinski: ok you convinced me it is confusing

GalRogozinski: I now noticed that you also pushed with javadocs. It is fine, but if you feel that your original change is justified next time you can explain, and I may accept.

footloosejava: You just gave me an idea ...

footloosejava:
I reverted to previous and also removed unneeded exceptions from throws list in methods.
Void is probably the best return for those methods. I had an idea to replace both FIFOCaches with one Map<Hash,Boolean> but then it would require tracking for size() reports and probably not make anything clearer.

PR #592, opened 2018-03-26
TCP BUG FIX and UDP minor fixup

footloosejava:
TCP
===
Added defensive copy of packet data. DatagramPacket data sent from Node.java is a reusable and the internal array is reused - resulting in errant/bad data in TCPNeighbor ArrayBlockingQueue.

UDP
===
Added comment to UDPNeighbor to explain why there is no defensive copy like with the TCPNeighbor.


src/main/java/com/iota/iri/network/TCPNeighbor.java
```diff
@@ -77,12 +77,18 @@ public void setSink(Socket sink) {
         this.sink = sink;
     }
 
+    /**
+     * This is a non-blocking write and that makes it necessary to make a defensive copy of the sent data.
+     *
```
alon-e: remove newline (or add one in the UDP doc)

footloosejava: done

gjeee:
i am trying a build with dev branch + pr591/592/593/596/597/599 
i do get some error statements error when trying:

git apply --check 592.patch
git apply --check 599.patch

is that harmful or innocent?

PR #593, opened 2018-03-26
Replicator source processor - major edit

footloosejava:
Major rewrite of TCP neighboring code.
- Fixed error handling and unclosed 'connection'
- Changed to buffered input
- Re-structured flow
- added 1 hour SO_TIMEOUT instead of FOREVER (or change to whatever is deemed reasonable).

### DIFF
--- a/src/main/java/com/iota/iri/network/replicator/ReplicatorSourceProcessor.java
+++ b/src/main/java/com/iota/iri/network/replicator/ReplicatorSourceProcessor.java
@@ -1,46 +1,38 @@
 package com.iota.iri.network.replicator;
 
+import com.iota.iri.network.Neighbor;
+import com.iota.iri.network.Node;
+import com.iota.iri.network.TCPNeighbor;
+import org.apache.commons.lang3.StringUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.BufferedInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.InetSocketAddress;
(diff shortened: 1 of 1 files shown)


src/main/java/com/iota/iri/network/replicator/ReplicatorSourceProcessor.java
```diff
@@ -50,149 +42,140 @@ public ReplicatorSourceProcessor(final ReplicatorSinkPool replicatorSinkPool,
...
-                }
-                
-                offset = 0;
+            try (InputStream inputStream = new BufferedInputStream(connection.getInputStream())) {
 
-                try {
-                    CRC32 crc32 = new CRC32();
-                    for (int i=0; i<TRANSACTION_PACKET_SIZE; i++) {
-                        crc32.update(data[i]);
+                {   // STAGE 1 - Read neighbors tcp listener port number and set it for the Sink
```
GalRogozinski: Instead of blocks it is preferable to extract code to a private method. Actually this entire `run` method should be broken down to smaller methods.

footloosejava:
@alon-e @GalRogozinski  Can we move on this? Any changes needed? Even if you plan to change it in the future, why not implement this now. This is a big improvement over the current one.

footloosejava:
I'll have to leave this with you now. I am too busy now with some other work.
Good luck!

GalRogozinski:
I am reopening this PR so when we refactor this class we won't forget about these changes

GalRogozinski:
Not needed due to #1072

PR #599, opened 2018-03-29
RocksDBPersistenceProvider - a few fixes

footloosejava:
RocksDBPersistenceProvider - lots of edits - no breaking changes.

Potential bug fixes:
* Made sure that all database iterators are closed in final methods ... or that they are closed in try-with-resources blocks.
* Also close all db connections after use or init failures.
* Added log warning for failed mkdir.
* Ensure that all sets returned are unmodifiable.

Fixes for general editing and clarity:
* Removed unneeded AtomicReference for maps.
* Centralized repeat code
* Marked unused code.
* Some misc cleaning.

Also edited:
- Pair.java - made all values final.

Added file:
- Quietly.java - can be used anywhere to quietly close resources without interrupting further closings or burying already thrown exceptions.


src/main/java/com/iota/iri/storage/rocksDB/RocksDBPersistenceProvider.java
```diff
@@ -5,38 +5,47 @@
...
 
-    private static final org.slf4j.Logger log = LoggerFactory.getLogger(RocksDBPersistenceProvider.class);
+    private static final Logger log = LoggerFactory.getLogger(RocksDBPersistenceProvider.class);
     private static final int BLOOM_FILTER_BITS_PER_KEY = 10;
 
+    private static final Pair<Indexable, Persistable> PAIR_OF_NULLS = new Pair<>(null, null);
+    private static final Set<Indexable> EMPTY_SET = Collections.emptySet();
+
     private final List<String> columnFamilyNames = Arrays.asList(
-            new String(RocksDB.DEFAULT_COLUMN_FAMILY),
```
GalRogozinski: Btw,
Seperate commits for style changes will be much appreciated in future contributions.
Github lets me select which commits to look at for review so I can cancel noise.

src/main/java/com/iota/iri/storage/rocksDB/RocksDBPersistenceProvider.java
```diff
@@ -83,438 +88,373 @@ public boolean isAvailable() {
...
             }
         }
-        return keys;
+        return keys == null ? EMPTY_SET : Collections.unmodifiableSet(keys);
+    }
+
+    /**
+     * @param key   What we are looking for.
+     * @param value The bytes we are searching in.
+     * @return true If the value byte[] starts with the key byte[].
```
GalRogozinski: @return true If the {@code value} starts with the {@code key}.

src/main/java/com/iota/iri/storage/rocksDB/RocksDBPersistenceProvider.java
```diff
@@ -83,438 +88,373 @@ public boolean isAvailable() {
...
+     */
+    private boolean compareStartsWithKey(byte[] key, byte[] value) {
```
GalRogozinski: When I see a method starts with `compare` it confuses me (I expect it to return an int).
Also it is a private method that is called with the parameters `(value, iterator.key())`, however it recieves `(key, value)`,
I find it also confusing.

Since code is written once but read many times please change to `keyStartsWithValue(byte[] value, byte[] key)`.

Also, this method can (and thus should)  be static.

src/main/java/com/iota/iri/utils/Pair.java
```diff
@@ -2,10 +2,12 @@
 
 /**
  * Created by paul on 4/15/17.
+ * Edited by footloosejava 03/28/2018
```
GalRogozinski: You can remove this and Paul's line above.
Every time you see a line like this it is a leftover or a mistake.

By the way this class will probably be replaced by `Pair` from Apache soon

src/main/java/com/iota/iri/utils/Quietly.java
```diff
@@ -0,0 +1,52 @@
...
+    public static void close(Closeable closeable, Consumer<Throwable> consumer) {
+        if (closeable == null) {
+            return;
+        }
+        try {
+            closeable.close();
+        } catch (Throwable closeException) {
+            if (consumer != null) {
+                try {
+                    consumer.accept(closeException);
```
GalRogozinski: I suppose this is the main reason you don't use `IOUtils.closeQuietly`?
Give an example of what you would with consumer please.

src/main/java/com/iota/iri/storage/rocksDB/RocksDBPersistenceProvider.java
GalRogozinski: and ofc change to fit below comment

src/main/java/com/iota/iri/storage/rocksDB/RocksDBPersistenceProvider.java
```diff
@@ -83,438 +88,373 @@ public boolean isAvailable() {
...
-        if( !db.keyMayExist(handle, index.bytes(), new StringBuffer()) ) {
-            counts.put(thing.getClass(), counts.get(thing.getClass()) + 1);
-        }
-        */
+        ColumnFamilyHandle handle = classTreeMap.get(thing.getClass());
         db.put(handle, index.bytes(), thing.bytes());
-        ColumnFamilyHandle referenceHandle = metadataReference.get().get(thing.getClass());
-        if(referenceHandle != null) {
+
+        ColumnFamilyHandle referenceHandle = metadataReference.get(thing.getClass());
```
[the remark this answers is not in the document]
footloosejava: :-)

src/main/java/com/iota/iri/storage/rocksDB/RocksDBPersistenceProvider.java
```diff
@@ -5,38 +5,47 @@
...
+ * All Sets returned are unmodifiable.
  */
 public class RocksDBPersistenceProvider implements PersistenceProvider {
 
-    private static final org.slf4j.Logger log = LoggerFactory.getLogger(RocksDBPersistenceProvider.class);
+    private static final Logger log = LoggerFactory.getLogger(RocksDBPersistenceProvider.class);
     private static final int BLOOM_FILTER_BITS_PER_KEY = 10;
 
+    private static final Pair<Indexable, Persistable> PAIR_OF_NULLS = new Pair<>(null, null);
+    private static final Set<Indexable> EMPTY_SET = Collections.emptySet();
```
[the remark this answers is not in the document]
footloosejava: Good point!

src/main/java/com/iota/iri/utils/Quietly.java
footloosejava: IOUtils works mostly too.
For consumer ... could do something like:
Quietly.close(closeable, e -> log.warn("You need to know this: {}",e));
Your call.

src/main/java/com/iota/iri/utils/Pair.java
footloosejava: I noticed 2 different Pair in the project ... Apache makes sense.

src/main/java/com/iota/iri/utils/Quietly.java
footloosejava: In any case - I subbed in IOUtils.

footloosejava:
Back to you 🦅

PR #605, opened 2018-04-02
Project-wide fix needed - InterruptExceptions ignored and not reset

footloosejava:
… sleeps, waits, and joins (and interrupts from blocking IO operations) need to be attended instead of ignored, and the thread Interrupt status needs to be reset to prevent further blockages.

Unless the interrupt is accidental (99% never the case), then all InterruptedExceptions need to be handled by resetting the interrupt status and then issuing a break/return/shutdown.

If this is not done, than any further blocking operations will continue to block and never return.
If an operation is 'shutdown' and may be in a blocking state, then a 'shutdown' flag change should also be accompanied by 'Thread.currentThread().interrupt();'

Reference: https://www.javaspecialists.eu/archive/Issue056.html

)

```
try {
   blocking op that can throw interruptExceptionm
} catch (final InterruptedException e) {
   Thread.currentThread().interrupt();
   (break or return or whatever)
}
```

### DIFF
--- a/src/main/java/com/iota/iri/IXI.java
+++ b/src/main/java/com/iota/iri/IXI.java
@@ -2,7 +2,6 @@ package com.iota.iri;
 
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
-import com.iota.iri.hash.ISS;
 import com.iota.iri.service.CallableRequest;
 import com.iota.iri.service.dto.AbstractResponse;
 import com.sun.nio.file.SensitivityWatchEventModifier;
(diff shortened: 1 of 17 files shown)


footloosejava:
@GalRogozinski  Feel free to close this if this is something that will be done in a project-wide refactoring. It involves pretty much every time a sleep, join, wait or other Interruption is thrown. A lot of classes will need a minor edit to bring this up to spec.

GalRogozinski:
The main reason this was not tended to is because of the size (again).
If you will break this up to formatting PR and PR with changes I will look at it.

GalRogozinski:
Just wondering @footloosejava,
will you do a new PR?

footloosejava:
No.

GalRogozinski:
@footloosejava, I Saw your other comments that you are busy with other work only after I answered.
Good luck!

footloosejava:
Thanks!

GalRogozinski:
Since there are many conflicts and the PR is too large, it is better to redo this.
See #1583

Issue #609, opened 2018-04-03
Create an ixi for testnet store messages api call


GalRogozinski:
### Description
For IRI version that can run as a testnet/mainnet hybrid, the "store messages" call should be outsourced to an ixi.
See https://github.com/iotaledger/iri/commit/db42893ca54165c6c8eda22ea12a0eeea0b6f2b1

### Motivation
This call should not be part of IRI

PR #610, opened 2018-04-03
IRI can be now deployed as testnet or mainnet compatible

GalRogozinski:
**Note**
Since the object of this PR is to help getting PRs get merged quickly, without creating too many conflicts with the PRs that are now waiting in queue. This fix is quick and dirty.
When `Configuration` will be refactored we will no longer have to pass values via constructors and methods. In the meanwhile when you review the code try to concentrate on things that break functionality.
# Description
A prerequisite for merging a PR, no matter how small and benign it seems, is to let it mature on the testnet. This hinders the mergers of PRs to this branch.

We are now creating a new process that will allow PRs to be merged quickly after they passed a review and basic sanity tests. In order to make this process work IRI code should be able to take the roll of a testnet deployment. Even though there is a `--testnet` flag, manual changes were required.

This change involves 3 commits:
1. Added an API call that is only functional on the testnet. This is actually not the correct way to do this and should be part of an ixi module. See #609.
2. Give the `--testnet flag` the ability to deploy iri as testnet. New flags were also introduced that will enable to run IRI in a customized private testnet.
3. Integrated a global snapshot of the public *testnet* (calm down mainnet is unaffected)
4. Version is updated because of testnet global snapshot.

**List of new flags:**
--testnet: [bool] flag to trigger further non-mainnet config
--testnet-coordinator: [string] coo address (no checksum)
--testnet-no-coo-validation: [bool] signature validation on/off
--snapshot: [file path] [default snapshotMainnet.txt/snapshotTestnet.txt]
--snapshot-sig: [file path] [default snapshotMainnet.sig/snapshotTestnet.sig] (although on testnet the signature is not verified)
--mwm: [int] [default 14/9] [min 13/0] [max ~/9]
--milestone-start [int] (subtangle first milestone, usually it's the latest snapshot milestone and/or most recent bootstrap+1 milestone) [default mainnet/testnet]
--milestone-keys: (advanced option: defines the depth of the merkle tree used to sign milestones by the coo) [int] [defaults 20/22]
--snapshot-timestamp: [epoch] [defaults epoch-mainnet/epoch-testnet]

## Type of change
- [x] Enhancement (non-breaking change which adds functionality)

# How Has This Been Tested?
Community testing is required

# Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have made corresponding changes to the documentation  - TBD


src/main/java/com/iota/iri/IRI.java
```diff
@@ -143,7 +143,6 @@ private static void validateParams(final Configuration configuration, final Stri
             configuration.put(DefaultConfSettings.COORDINATOR, Configuration.TESTNET_COORDINATOR_ADDRESS);
             configuration.put(DefaultConfSettings.SNAPSHOT_FILE, Configuration.TESTNET_SNAPSHOT_FILE);
             configuration.put(DefaultConfSettings.MILESTONE_START_INDEX, Configuration.TESTNET_MILESTONE_START_INDEX);
-            configuration.put(DefaultConfSettings.SNAPSHOT_SIGNATURE_FILE, Configuration.TESTNET_SNAPSHOT_SIG_FILE);
```
alon-e: I understand the reasoning here - that `snapshotTestnet.sig` is not validated so there is no need to point to it,
However, his means that `snapshotMainnet.sig` will be read also for testnet, even if the signature is not validated it still mixes file names.

better to set as `null` / `""`

src/main/java/com/iota/iri/SignedFiles.java
```diff
@@ -22,21 +20,29 @@
...
-                }
-            } catch (IOException e) {
-                log.error("Can't read file " + filename, e);
-                return false;
+        BufferedReader reader = null;
+        //digest file
+        try {
+            InputStream inputStream = SignedFiles.class.getResourceAsStream(filename);
+            if (inputStream == null) {
+                inputStream = new FileInputStream(filename);
```
alon-e: add a comment to explicitly state that if a path is not in resources, then reading the file from OS.

src/main/java/com/iota/iri/Snapshot.java
```diff
@@ -29,46 +23,78 @@
...
+            checkInitialSnapshotIsConsistent(initialState);
+
+        }
+        return initialSnapshot;
+    }
+
+    private static InputStream getSnapshotStream(String snapshotPath) throws FileNotFoundException {
+        InputStream inputStream = Snapshot.class.getResourceAsStream(snapshotPath);
+        if (inputStream == null) {
+            inputStream = new FileInputStream(snapshotPath);
```
alon-e: same comment as in signedFiles

src/main/java/com/iota/iri/IRI.java
GalRogozinski: set to "".
I think `null` will raise an exception when inserted as a value to `ConcurrentHashMap<>`

src/main/java/com/iota/iri/SignedFiles.java
GalRogozinski: Added similar comment in `Snapshot.java` as well

GalRogozinski:
Sanity tests were performed

footloosejava:
Nice work! Let me know if you need help with `Configuration`. I might have a few ideas - see [https://github.com/footloosejava/annoconf](https://github.com/footloosejava/annoconf ).  I'm pretty handy with Guice too ;-)

PR #613, opened 2018-04-05
Kerl optimizations: 30%+ faster Kerl.absorb and 15% to 20% faster Kerl.Squeeze

footloosejava:
Across the board 30%+ improvement in Kerl speed during absorb.
Up to 15% to 20% improvement in Kerl speed in squeeze.

Kerl performance optimizations:
- using a cache of function powers
- avoiding unneeded array range copy when range is all the bytes
- misc optimizations
 
Converter.java - Added exception instead of null return value.
SpongeFactory.java - Added exception if unrecognized Mode.


src/main/java/com/iota/iri/hash/Curl.java
```diff
@@ -8,9 +8,8 @@
 
 /**
  * (c) 2016 Come-from-Beyond and Paul Handy
- *
+ * <p>
```
footloosejava: No changes in this Curl.java class except paragraph symbol that the editor added - unbeknownst to me!

src/main/java/com/iota/iri/hash/Kerl.java
```diff
@@ -11,8 +11,15 @@
...
 
-    private byte[] byte_state;
-    private int[] trit_state;
+    // radix constant conversion for fast math
+    public static final BigInteger RADIX = BigInteger.valueOf(Converter.RADIX);
+    // lookups for fast math
+    private static final BigInteger[] RADIX_POWERS = makeRadixPowers();
+    private static final BigInteger[] RADIX_POSITIVE_ONE_FUNCTION = makeRadixFunctionPowers(true);
+    private static final BigInteger[] RADIX_NEGATIVE_ONE_FUNCTION = makeRadixFunctionPowers(false);
+
```
footloosejava: The radix positive one function is a prepopulated lookup so that repeat trits of 1 or -1 are easily converted without multiple round-trips of `value = value.multiply(RADIX).add(ONE)`

In the first index is 1.
The second is B + 1
The third is B^2 + B + 1
The fourth is B^3 + B^2 + B + 1
etc.

The negative one is:

- AB^2 - B - 1
- AB^3 - B^2 - B - 1 
etc.

There are many instances of repeat number in the trits array and this saves both the round tripping and a lot of instances of BigInteger.

src/main/java/com/iota/iri/hash/Kerl.java
```diff
@@ -85,63 +91,108 @@ public void squeeze(final int[] trits, int offset, int length) {
...
+                    while (i - 1 >= offset && trits[i - 1] == -1) {
+                        i--;
+                        sequence++;
+                    }
+                    value = value.multiply(RADIX_POWERS[power + 1 + sequence]).add(RADIX_NEGATIVE_ONE_FUNCTION[sequence]);
+                    power = 0;
+                }
+                break;
+                default:
+                    power++;
```
footloosejava: If it is a ZERO trit, we increment the power so that the next round the power of the multiply is fetched from a lookup.
At the end, if all we have is ZEROS, then the POWER will need to be applied after the loop finishes.

src/main/java/com/iota/iri/hash/Kerl.java
```diff
@@ -85,63 +91,108 @@ public void squeeze(final int[] trits, int offset, int length) {
...
-
         return value;
     }
 
     public static BigInteger bigIntFromBytes(final byte[] bytes, final int offset, final int size) {
-
-        return new BigInteger(Arrays.copyOfRange(bytes, offset, offset + size));
+        return (bytes.length == offset + size) ? new BigInteger(bytes)
+            : new BigInteger(Arrays.copyOfRange(bytes, offset, offset + size));
     }
```
footloosejava: Almost 100% of the time, the size is the array given. So this is an easy optimization to not do a copy.

src/main/java/com/iota/iri/hash/Kerl.java
```diff
@@ -85,63 +91,108 @@ public void squeeze(final int[] trits, int offset, int length) {
...
+        if (destination.length - offset < size) {
             throw new IllegalArgumentException("Destination array has invalid size");
         }
 
-        BigInteger absoluteValue = value.compareTo(BigInteger.ZERO) < 0 ? value.negate() : value;
+        final int signum = value.signum();
+        if (signum == 0) {
+            Arrays.fill(destination, offset, size, 0);
+            return;
+        }
```
footloosejava: We can shortcut the whole thing with the signum and an empty array.

src/main/java/com/iota/iri/utils/Converter.java
```diff
@@ -232,9 +231,9 @@ public static String asciiToTrytes(String input) {
         StringBuilder sb = new StringBuilder(80);
         for (int i = 0; i < input.length(); i++) {
             int asciiValue = input.charAt(i);
-            // If not recognizable ASCII character, return null
             if (asciiValue > 255) {
-                return null;
+                throw new IllegalArgumentException("The ascii trytes String contains un illegal tryte character:" +
+                    " char= " + ((char)asciiValue) + ", (" + asciiValue +")");
             }
```
footloosejava: returning null will only cause a null pointer exception - and a hard to find one! Either throw a null pointer exception here , or else a more useful message such as this suggested message.

PR #614, opened 2018-04-05
edit SignedFiles and Snapshot

footloosejava:
Major/minor edit of SignedFiles.java and Snapshot.java

Major edit of SignedFiles.isFilesSignatureValid
- created two sub methods: digestFile(...) and validateSignature(...)

Major/minor edits of Snapshot.java 
- removed unneeded (unused) 'stateIterator.remove()'' from Snapshot.isConsistent.
This previously only applied to diffs that were immediately disposed and never used in application.
- moved pruning of zero values to Snapshot.init (where it also might not be necessary but that is where it does something)
- made init() method threadsafe synced on the Snapshot.class
- centralized System.exit() to one catch block.
-renamed clone() to copySnapshot (clone() is reserved for Clonable, with a true clone)
- encapsulated state to getState()
- ensured that all operations with locks are enclosed in try-finally blocks.
- streamlined the map value concatenation to make it faster and read clearer.

ALSO:
- removed redundant bufferedInputStreams. Buffered reader already buffers a full IO buffer on each read.
- converted read function to a lambda method, eliminating a null check and loop.
- removed REGEX parsing of line and replaced with a quick index of ';' function
- converted all file operations to try with resources (automatically adds nested exceptions if on close)

Signed-off-by: footloosejava <32090281+footloosejava@users.noreply.github.com>

(cherry picked from commit d3bf92e)


src/main/java/com/iota/iri/Snapshot.java
```diff
@@ -2,175 +2,177 @@
...
+            });
+            return state;
         } catch (IOException e) {
-            //syso is left until logback is fixed
-            System.out.println("Failed to load snapshot.");
-            log.error("Failed to load snapshot.", e);
-            System.exit(-1);
-        }
-        finally {
-            IOUtils.closeQuietly(reader);
```
footloosejava: If using try-with-resources, there is no need to close quietly. It will automatically add any extra exceptions as `addSuppressed` in the original exception.

src/main/java/com/iota/iri/Snapshot.java
```diff
@@ -2,175 +2,177 @@
...
-
-            final Map.Entry<Hash, Long> entry = stateIterator.next();
-            if (entry.getValue() <= 0) {
-
-                if (entry.getValue() < 0) {
-                    log.info("Skipping negative value for address: " + entry.getKey() + ": " + entry.getValue());
-                    return false;
-                }
-
-                stateIterator.remove();
```
footloosejava: In all instances, this method is called on maps that are never used in the end. So the call to remove does nothing. The only map it could apply to have any effect on is the initial snapshot. It has been eliminated and added into init().

src/main/java/com/iota/iri/Snapshot.java
```diff
@@ -2,175 +2,177 @@
...
+                // of ZERO values in original Snapshot.isConsistent(...)
+                initialState.entrySet().removeIf(entry -> entry.getValue() == 0);
+
+                return initialSnapshot = new Snapshot(initialState, 0);
+
+            } catch (Exception e) {
+                log.error("Quitting - Unable to verify snapshot: " + e, e);
+                System.exit(-1);
+                throw new Error("unreachable");
             }
```
footloosejava: System.exit(-1) should really not be here .. or anywhere outside the IRI.main method. 
As part of initialization, it should be possible to throw an exception that can be captured at a strategic point so that the application can exit. This also limits testing because if a test was designed to make this fail, then the application could not properly test it since it exits.

src/main/java/com/iota/iri/Snapshot.java
```diff
@@ -2,175 +2,177 @@
...
+                // Prune zero values - to honour original intent of 'remove'
+                // of ZERO values in original Snapshot.isConsistent(...)
+                initialState.entrySet().removeIf(entry -> entry.getValue() == 0);
+
+                return initialSnapshot = new Snapshot(initialState, 0);
+
+            } catch (Exception e) {
+                log.error("Quitting - Unable to verify snapshot: " + e, e);
+                System.exit(-1);
+                throw new Error("unreachable");
```
footloosejava: System.exit(-1) should really not be here .. or anywhere outside the IRI.main method.
As part of initialization, it should be possible to throw an exception that can be captured at a strategic point so that the application can exit. This also limits testing because if a test was designed to make this fail, then the application could not properly test it since it exits.
The Error exception is there to prevent a 'missing return' statement warning because System.exit() does not throw.

footloosejava:
I have an update for this ... just waiting to git it up and change the PR.

src/main/java/com/iota/iri/Snapshot.java
```diff
@@ -2,175 +2,177 @@
...
-        int i;
         rwlock.readLock().lock();
-        i = index;
-        rwlock.readLock().unlock();
-        return i;
+        try {
+            return index;
+        } finally {
+            rwlock.readLock().unlock();
+        }
```
GalRogozinski: Like on putting the unlock in finally

src/main/java/com/iota/iri/Snapshot.java
```diff
@@ -2,175 +2,177 @@
...
-        patch = diff.entrySet().stream().map(hashLongEntry ->
-            new HashMap.SimpleEntry<>(hashLongEntry.getKey(), state.getOrDefault(hashLongEntry.getKey(), 0L) + hashLongEntry.getValue())
-        ).collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
-        rwlock.readLock().unlock();
-        return patch;
+        try {
+            Map<Hash, Long> map = new HashMap<>(diff.size());
+            diff.forEach((key, value) -> {
+                Long previousValue = state.getOrDefault(key, 0L);
+                map.put(key, Math.addExact(previousValue, value));
```
GalRogozinski: much more readable
Thanks

footloosejava:
@GalRogozinski 
See https://github.com/iotaledger/iri/pull/729 and https://github.com/iotaledger/iri/pull/730

Because of the changes in 729, a lot of dependencies are effected. I hope it isn't too confusing.

PR #623, opened 2018-04-11
IRI-326 ⁃ rename checkstyle xml to match convention - codacy integration

alon-e:
# Description

rename checkstyle xml to match convention - codacy integration

## Type of change

# How Has This Been Tested?

`mvn test` to see that `checkstyle` still runs.


GalRogozinski:
FYI,
Running `mvn validate` will be faster.

Do *NOT* use `mvn checkstyle:check` because it will override the configuration with default checkstyle rules

PR #624, opened 2018-04-11
allow enabling ZMQ via CLI flags

alon-e:
# Description

adds a `--zmq-enable` flag to enable ZMQ via CLI

Fixes # https://iotaledger.atlassian.net/browse/IRI-27

## Type of change

# How Has This Been Tested?

- manual test of launching with and without the flag, and reading the ZMQ feed.


geminoz:
The cmd line flag is --zmq-enabled which is inline with the iri config options ZMQ_ENABLED. Looks good to me.

PR #628, opened 2018-04-13
Kerl.java absorb now 😈 666% faster 😈 - plus 🍦 cleaner and 🍦 simplified code

footloosejava:
`Kerl.absorb(...)` now 666% faster, plus cleaner and simplified code ! I'm exaggerating a little, it is actually only 665.55% faster. 
Snapshot.initial loading is now 👍 315% faster. 👍 

Big time hog was **bigIntFromTrits(...)**. Everything else in Kerl.java was very fast already.
- Created a fast track for Radix power multiplication.
- Created a native math routine to munch 40 trits at a time into the main math routine.
- Added test to ensure that there is no math overflow.

Reusable instance state removed since it complicates reading the code and offers no performance advantages. All state now method-local.

As per @th0br0  suggestion, I compared to alternate implementation [jota...kerl](https://github.com/iotaledger/iota.lib.java/blob/master/jota/src/main/java/jota/pow/Kerl.java) 
- My 'absorb' is **468% faster** - 4.34 seconds faster each round.
- jota-kerl 'squeeze' is **first-round 220% faster** than mine. However, it is insignificant since the difference from 'squeeze' on digesting a 20 MB file is less than 2/5th of a millisecond. After 20 rounds, my squeeze (and the original) are still faster.

**Tested digesting 20 complete loads of the 20MB+ snapshot. All times milliseconds.**

FIRST RUN | OLD | MINE | jota-kerl
-----------  | ----- | ------ | -------
SNAPSHOT | 9693 | **3074** | 7467
ABSORB     | 7920 |  **1190** | 5570
SQUEEZE    | 0.80  |      0.77    | 0.35

AVG of 20 runs |   OLD |   MINE |  jota-kerl
-------- | -------- | -----| -------
SNAPSHOT | 8760 |	**2569**	| 6784
ABSORB     | 7195 |	**1029**  | 5223
SQUEEZE	  |  0.19  |	0.18	| 0.32

Fixes # (issue)

## Type of change
- Enhancement - Speeds up Kerl considerably

# How Has This Been Tested?
- Test written
- Tested on Snapshot loading
- Tested via existing Kerl and Curl and Bundle Validator tests

# Checklist:
 [X] My code follows the style guidelines for this project
 [X] I have performed a self-review of my own code
 [X] I have added tests that prove my fix is effective or that my feature works
 [X] New and existing unit tests pass locally with my changes

### DIFF
--- a/src/main/java/com/iota/iri/hash/Kerl.java
+++ b/src/main/java/com/iota/iri/hash/Kerl.java
@@ -13,19 +13,14 @@ public final class Kerl implements Sponge {
     public static final int BIT_HASH_LENGTH = 384;
     public static final int BYTE_HASH_LENGTH = BIT_HASH_LENGTH / 8;
 
-    // radix constant conversion for fast math
     public static final BigInteger RADIX = BigInteger.valueOf(Converter.RADIX);
     private static final int MAX_POWERS_LONG = 40;
     private static final BigInteger[] RADIX_POWERS = IntStream.range(0, MAX_POWERS_LONG + 1).mapToObj(RADIX::pow).toArray(BigInteger[]::new);
 
-    private final byte[] byte_state;
-    private final int[] trit_state;
     private final Keccak.Digest384 keccak;
 
     protected Kerl() {
         this.keccak = new Keccak.Digest384();
-        this.byte_state = new byte[BYTE_HASH_LENGTH];
-        this.trit_state = new int[Sponge.HASH_LENGTH];
     }
 
     @Override
@@ -39,15 +34,11 @@ public final class Kerl implements Sponge {
             throw new RuntimeException("Illegal length: " + length);
         }
         for (int pos = offset; pos < offset + length; pos += HASH_LENGTH) {
-            //copy trits[offset:offset+length]
-            System.arraycopy(trits, pos, trit_state, 0, HASH_LENGTH);
-
-            //convert to bits
-            trit_state[HASH_LENGTH - 1] = 0;
-            bytesFromBigInt(bigIntFromTrits(trit_state, 0, HASH_LENGTH), byte_state);
-
-            //run keccak
-            keccak.update(byte_state);
+            //convert to bytes && update
+            byte[] state = new byte[BYTE_HASH_LENGTH];
+            trits[pos + HASH_LENGTH - 1] = 0;
+            bytesFromBigInt(bigIntFromTrits(trits, pos, HASH_LENGTH), state);
+            keccak.update(state);
         }
     }
 
@@ -58,19 +49,20 @@ public final class Kerl implements Sponge {
         }
         try {
             for (int pos = offset; pos < offset + length; pos += HASH_LENGTH) {
-                this.keccak.digest(byte_state, 0, BYTE_HASH_LENGTH);
-                //convert to trits
-                tritsFromBigInt(bigIntFromBytes(byte_state, 0, BYTE_HASH_LENGTH), trit_state, 0, Sponge.HASH_LENGTH);
 
-                //copy with offset
-                trit_state[HASH_LENGTH - 1] = 0;
-                System.arraycopy(trit_state, 0, trits, pos, HASH_LENGTH);
+                byte[] state = new byte[BYTE_HASH_LENGTH];
+                keccak.digest(state, 0, BYTE_HASH_LENGTH);
+
+                //convert into trits
+                BigInteger value = new BigInteger(state);
+                tritsFromBigInt(value, trits, pos, Sponge.HASH_LENGTH);
+                trits[pos + HASH_LENGTH - 1] = 0;
 
                 //calculate hash again
-                for (int i = byte_state.length; i-- > 0; ) {
-                    byte_state[i] = (byte) (byte_state[i] ^ 0xFF);
+                for (int i = state.length; i-- > 0; ) {
+                    state[i] = (byte) (state[i] ^ 0xFF);
                 }
-                keccak.update(byte_state);
+                keccak.update(state);
             }
         } catch (DigestException e) {
             e.printStackTrace(System.err);
@@ -86,20 +78,15 @@ public final class Kerl implements Sponge {
         }
         BigInteger value = BigInteger.ZERO;
         for (int n = offset + size - 1; n >= offset; ) {
+            int count = 0;
             long num = 0L;
-            int power = 0;
-            while (n >= offset && power < MAX_POWERS_LONG) {
+            while (n >= offset && count < MAX_POWERS_LONG) {
                 num = 3 * num + trits[n--];
-                power++;
+                count++;
             }
-            value = value.multiply(RADIX_POWERS[power]).add(BigInteger.valueOf(num));
+            value = value.multiply(RADIX_POWERS[count]).add(BigInteger.valueOf(num));
         }
         return value;
-    }
-
-    public static BigInteger bigIntFromBytes(final byte[] bytes, final int offset, final int size) {
-        return (bytes.length == offset + size) ? new BigInteger(bytes)
-            : new BigInteger(Arrays.copyOfRange(bytes, offset, offset + size));
     }
 
     public static void tritsFromBigInt(final BigInteger value, final int[] destination, final int offset, final int size) {
@@ -130,10 +117,11 @@ public final class Kerl implements Sponge {
         if (destination.length < BYTE_HASH_LENGTH) {
             throw new IllegalArgumentException("Destination array has invalid size.");
         }
-        int start = BYTE_HASH_LENGTH - (value.bitLength() / 8 + 1);
+        byte[] bytes = value.toByteArray();
+        int start = BYTE_HASH_LENGTH - bytes.length;
         Arrays.fill(destination, 0, start, (byte) (value.signum() < 0 ? -1 : 0));
-        for (byte aByte : value.toByteArray()) {
-            destination[start++] = aByte;
+        for (int i = 0; i < bytes.length; i++) {
+            destination[start++] = bytes[i];
         }
     }
 }
--- a/src/main/java/com/iota/iri/hash/Kerl.java
+++ b/src/main/java/com/iota/iri/hash/Kerl.java
@@ -6,17 +6,22 @@ import org.bouncycastle.jcajce.provider.digest.Keccak;
 import java.math.BigInteger;
 import java.security.DigestException;
 import java.util.Arrays;
+import java.util.stream.IntStream;
 
-public class Kerl implements Sponge {
+public final class Kerl implements Sponge {
+
     public static final int BIT_HASH_LENGTH = 384;
     public static final int BYTE_HASH_LENGTH = BIT_HASH_LENGTH / 8;
 
-    private byte[] byte_state;
-    private int[] trit_state;
+    // radix constant conversion for fast math
+    public static final BigInteger RADIX = BigInteger.valueOf(Converter.RADIX);
+    private static final BigInteger[] RADIX_POWERS = IntStream.range(0, Sponge.HASH_LENGTH + 1).mapToObj(RADIX::pow).toArray(BigInteger[]::new);
+
+    private final byte[] byte_state;
+    private final int[] trit_state;
     private final Keccak.Digest384 keccak;
 
     protected Kerl() {
-
         this.keccak = new Keccak.Digest384();
         this.byte_state = new byte[BYTE_HASH_LENGTH];
         this.trit_state = new int[Sponge.HASH_LENGTH];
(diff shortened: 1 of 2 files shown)


src/main/java/com/iota/iri/hash/Kerl.java
```diff
@@ -6,142 +6,174 @@
...
         }
     }
-    public static void bytesFromBigInt(final BigInteger value, byte[] destination, int offset) {
-        if(destination.length - offset < BYTE_HASH_LENGTH) {
-            throw new IllegalArgumentException("Destination array has invalid size.");
-        }
-
-        final byte[] bytes = value.toByteArray();
-        int i = 0;
-        while (i + bytes.length < BYTE_HASH_LENGTH) {
```
footloosejava: offset was never used - so I removed.
Also, Think it is clearer with Arrays.fill .. and also use some math to calc length without creating a variable for array.

src/main/java/com/iota/iri/hash/Kerl.java
```diff
@@ -6,142 +6,174 @@
...
-
         return value;
     }
 
     public static BigInteger bigIntFromBytes(final byte[] bytes, final int offset, final int size) {
-
-        return new BigInteger(Arrays.copyOfRange(bytes, offset, offset + size));
+        return (bytes.length == offset + size) ? new BigInteger(bytes)
+            : new BigInteger(Arrays.copyOfRange(bytes, offset, offset + size));
     }
```
footloosejava: bigIntFromBytes - most times array is proper size and no copy is needed. Added conditional copy.

src/main/java/com/iota/iri/hash/Kerl.java
```diff
@@ -6,142 +6,174 @@
...
-
-            value = value.multiply(BigInteger.valueOf(Converter.RADIX)).add(BigInteger.valueOf(trits[offset + i]));
+        long num;
+        int power = 0;
+        int n = offset + size - 1;
+        for (; n > offset + 1; n--) {
+            switch (trits[n]) {
+                case 0: // a * b
+                    power++;
+                    break;
```
footloosejava: powers are accumulated

src/main/java/com/iota/iri/hash/Kerl.java
```diff
@@ -6,142 +6,174 @@
...
+                case 1: // a * b + 1
+                    num = 9 + (trits[--n] * 3) + trits[--n];
+                    for (int count = 0; n > offset && count <= 36; count++, power++) {
+                        num = 3 * num + trits[--n];
+                    }
+                    value = value.multiply(RADIX_POWERS[power + 3]).add(BigInteger.valueOf(num));
+                    power = 0;
+                    break;
```
footloosejava: Up to 36 trits can be safely processed without using a BigInteger for the math. 
This speeds things up considerably.

src/main/java/com/iota/iri/hash/Kerl.java
```diff
@@ -6,142 +6,174 @@
...
-            bytesFromBigInt(bigIntFromTrits(trit_state, 0, HASH_LENGTH), byte_state, 0);
+            bytesFromBigInt(bigIntFromTrits(trit_state, 0, HASH_LENGTH), byte_state);
 
             //run keccak
             keccak.update(byte_state);
-            offset += HASH_LENGTH;
-
-        } while ((length -= HASH_LENGTH) > 0);
+        }
     }
```
footloosejava: I just thought it reads clearer with a traditional for loop than tracking both offset and length in the old version. Now the inputs are also final.

src/main/java/com/iota/iri/hash/Kerl.java
```diff
@@ -6,142 +6,174 @@
...
+                //copy with offset
+                trit_state[HASH_LENGTH - 1] = 0;
+                System.arraycopy(trit_state, 0, trits, pos, HASH_LENGTH);
+
+                //calculate hash again
+                for (int i = byte_state.length; i-- > 0; ) {
+                    byte_state[i] = (byte) (byte_state[i] ^ 0xFF);
+                }
+                keccak.update(byte_state);
+            }
```
footloosejava: Same here - traditional for loop instead of tracking both offset and length. Inputs also final now.

src/main/java/com/iota/iri/hash/Kerl.java
```diff
@@ -6,142 +6,174 @@
...
+                    power = 0;
+                    break;
+                case -1:
+                    value = value.multiply(RADIX_POWERS[power + 1]).subtract(BigInteger.ONE);
+                    power = 0;
+                    break;
+                default:
+                    throw new RuntimeException("unreachable");
+            }
+        }
```
footloosejava: There is a leftover 1 or 2 trits that need  processing .. .and also any un multiplies powers that accumulated.

src/test/java/com/iota/iri/hash/KerlTest.java
```diff
@@ -70,26 +70,36 @@ public void loopRandTritsFromBigInt() throws Exception {
...
+    @Test
+    public void limitBigIntFromTrits() {
+        // this confirms that the limit we set using long math works for any size input
+        int[] trits = new int[2048];
+        Arrays.fill(trits, 1);
+        Kerl.bigIntFromTrits(trits, 0, trits.length);
+        Arrays.fill(trits, -1);
+        Kerl.bigIntFromTrits(trits, 0, trits.length);
+    }
+
```
footloosejava: This is the one added test. Nothing fancy - just proves that with a big input, the overflow does not happen.

th0br0:
Did you happen to also compare the old implementation (which is known to be slow) with the optimised variant that all libraries are using?
See here for a Java, BigInt-less implementation:
https://github.com/iotaledger/iota.lib.java/blob/master/jota/src/main/java/jota/pow/Kerl.java

footloosejava:
@th0br0 Didn't know about the others.
I just took a look. **Mine is a lot simpler and more straightforward.** I'll have to see how they compare ....

th0br0:
It definitely is simpler and easier to parse. 
The reason for the current IRI codebase (still) doing things slowly mainly was prudence.

footloosejava:
@th0br0 test done. Mine 'absorb' is is **at least 430% faster** than the other one. Over 5 seconds faster each round.
JCURL 'squeeze' is is **178% faster** than mine. However, the difference from 'squeeze' on digesting a 20 MB file is less than 1/3 of 1/1000th of a second.

Tested digesting 20 complete load of of the 20MB+ snapshot.

FIRST RUN | OLD | MINE | JCURL
-----------  | ----- | ------ | -------
SNAPSHOT |  9693	| **2879** | 7487 millis
ABSORB | 7920 |  **1306** | 5572 millis
SQUEEZE|	    800828|	832151|	367134 nanos (1/3 of a millis)

AVG of 20 |   OLD |   MINE |  JCURL
-------- | -------- | -----| -------
SNAPSHOT|	8760 |	**2378**	|6754
ABSORB|		7195 |	**1197**|	5158
SQUEEZE	 |   187030|	186910	|332769 nanos (1/3 of a millis)

src/main/java/com/iota/iri/hash/Kerl.java
```diff
@@ -6,142 +6,174 @@
...
     }
 
     public static BigInteger bigIntFromTrits(final int[] trits, final int offset, final int size) {
-
+        for (int i = offset; i < offset + size; i++) {
+            if (trits[i] < -1 || trits[i] > 1) {
+                throw new IllegalArgumentException("not a trit: " + trits[i]);
+            }
+        }
         BigInteger value = BigInteger.ZERO;
```
footloosejava: This check really isn't required from trusted data. There isn't checking anywhere else ... but I thought it was good form to include it. Ideally, all the trit[] should be checked on first input. Tests show that the check doesn't really slow anything down.

footloosejava:
Just realized that simpler is even better. Removed all the complicated goop, and kept the essential goodness .. and pretty much as fast as before but 1000% easier to read.

src/main/java/com/iota/iri/hash/Kerl.java
```diff
@@ -6,142 +6,132 @@
...
-        for (int i = size; i-- > 0; ) {
-
-            value = value.multiply(BigInteger.valueOf(Converter.RADIX)).add(BigInteger.valueOf(trits[offset + i]));
+        for (int n = offset + size - 1, count = 0; n >= offset; count = 0) {
+            long num = 0L;
+            for (; n >= offset && count < MAX_POWERS_LONG; count++, n--) {
+                num = 3 * num + trits[n];
+            }
+            value = value.multiply(RADIX_POWERS[count]).add(BigInteger.valueOf(num));
         }
```
footloosejava:
```
BigInteger value = BigInteger.ZERO;
for (int n = offset + size - 1, count = 0; n >= offset; count = 0) {
    long num = 0L;
    for (; n >= offset && count < MAX_POWERS_LONG; count++, n--) {
        num = 3 * num + trits[n];
    }
    value = value.multiply(RADIX_POWERS[count]).add(BigInteger.valueOf(num));
}
```

It all comes down to this routine. Do the math the native way until the power is too big, then merge it into the BigInteger and do it all again. The Radix powers from 0 to 40 are pre-calculated.

footloosejava:
@GalRogozinski not sure why this build failed. Works fine on my end. Looks like a Travis error.

src/main/java/com/iota/iri/hash/Kerl.java
```diff
@@ -6,142 +6,134 @@
...
-            value = value.multiply(BigInteger.valueOf(Converter.RADIX)).add(BigInteger.valueOf(trits[offset + i]));
+        for (int n = offset + size - 1; n >= offset; ) {
+            long num = 0L;
+            int power = 0;
+            while (n >= offset && power < MAX_POWERS_LONG) {
+                num = 3 * num + trits[n--];
+                power++;
+            }
+            value = value.multiply(RADIX_POWERS[power]).add(BigInteger.valueOf(num));
         }
```
footloosejava:
```
BigInteger value = BigInteger.ZERO;
for (int n = offset + size - 1; n >= offset; ) {
    long num = 0L;
    int power = 0;
    while (n >= offset && power < MAX_POWERS_LONG) {
        num = 3 * num + trits[n--];
        power++;
    }
    value = value.multiply(RADIX_POWERS[power]).add(BigInteger.valueOf(num));
}
```

It all comes down to this routine. Do the math the native way ... until the power is too big. Then merge it into the BigInteger ... and do it all again. The radix powers from 0 to 40 are pre-calculated.

footloosejava:
@pinpong @jpkrohling In case you are interested.

gjeee:
nice job again footloose! i used your Kerl class and my CPU sure seems to like it......

footloosejava:
Commit incoming - I just found a way to make it faster again ... and **even simpler** with **no instance state**. There will be lots of ice-cream with cappuccinos tonight ☕️  + 🍨  :)

src/main/java/com/iota/iri/hash/Kerl.java
```diff
@@ -6,142 +6,122 @@
...
+import java.util.stream.IntStream;
+
+public final class Kerl implements Sponge {
 
-public class Kerl implements Sponge {
     public static final int BIT_HASH_LENGTH = 384;
     public static final int BYTE_HASH_LENGTH = BIT_HASH_LENGTH / 8;
 
-    private byte[] byte_state;
-    private int[] trit_state;
```
footloosejava: The instance state has been removed and replaced with method-local state. It make reading the processing much easier and there is no performance hit.

src/main/java/com/iota/iri/hash/Kerl.java
```diff
@@ -6,142 +6,122 @@
...
-            value = value.multiply(BigInteger.valueOf(Converter.RADIX)).add(BigInteger.valueOf(trits[offset + i]));
+        for (int n = offset + size - 1; n >= offset; ) {
+            int count = 0;
+            long num = 0L;
+            while (n >= offset && count < MAX_POWERS_LONG) {
+                num = 3 * num + trits[n--];
+                count++;
+            }
+            value = value.multiply(RADIX_POWERS[count]).add(BigInteger.valueOf(num));
         }
```
footloosejava:
```
BigInteger value = BigInteger.ZERO;
for (int n = offset + size - 1; n >= offset; ) {
    long num = 0L;
    int power = 0;
    while (n >= offset && power < MAX_POWERS_LONG) {
        num = 3 * num + trits[n--];
        power++;
    }
    value = value.multiply(RADIX_POWERS[power]).add(BigInteger.valueOf(num));
}
```
It all comes down to this routine. Do the math the native way ... until the power is too big. Then merge it into the BigInteger ... and do it all again. The radix powers from 0 to 40 are pre-calculated.

alon-e:
Hey @footloosejava - thanks for this contribution!
reviewing now.
did you also run the `test_vectors` in https://github.com/iotaledger/kerl ?

gjeee:
@footloosejava i applied your latest Kerl.java commit as well but i think this one has issues....my CPU spikes as never before.....the one before was like baby smooth......

????

footloosejava:
@gjeee That';s weird. Try again.

footloosejava:
@alon-e I'll take a look. Thanks.

src/test/java/com/iota/iri/hash/KerlTest.java
GalRogozinski: Hmmm, looking again at this test, I am uncertain if it is good for anything.
BigInteger doesn't produce overflow exceptions as far as I know.

The only way this test can fail is with `OutOfMemoryException`.
Am I missing anything?
If so do tell, else remove it please.

footloosejava: I will have to revise the test to be more explicit. Basically, this has a long math routine in it that needs to be checked for overflow:

```
BigInteger value = BigInteger.ZERO;
        for (int n = offset + size - 1; n >= offset; ) {
            int count = 0;
            long num = 0L;
            while (n >= offset && count < MAX_POWERS_LONG) {
                num = 3 * num + trits[n--];
                count++;
            }
            value = value.multiply(RADIX_POWERS[count]).add(BigInteger.valueOf(num));
        }
        return value;
```

The correct test would prove that trit[MAX_POWERS_LONG] produces no overflow in `num = 3 * num + trits[n--]`

The current test proves that in a not-so-clear fashion. As the boundary between MAX_POWERS_LONG is crossed, the  BigInteger result should always be the same regardless of the type of math used.

The trit[2048] test proves that ... but it just needs to be changed to trit[MAX_POWERS_LONG];

footloosejava:
@GalRogozinski see last commit and comments. Thanks!

gjeee:
ok i applied commit again and all seems fine. sorry for the wrong diagnosis!

footloosejava:
@gjeee No problem. Your feedback and second set of eyes is always appreciated!

src/test/java/com/iota/iri/hash/KerlTest.java
GalRogozinski: With the assertion this test makes more sense since `int`s don't throw overflow exceptions ;-)

alon-e:
@footloosejava @galrogo I ran the test_vectors locally from: https://github.com/iotaledger/kerl 
all pass as expected :+1:

footloosejava:
yeah 👍

PR #630, opened 2018-04-16
IRI-329 ⁃ `readPreviousEpochsSpentAddresses` should only run on mainnet

GalRogozinski:
# Description

When creating private testnets we don't want to read the previous spent addresses file from the mainnet.
So when we are in testnet mode `previousEpochsSpentAddresses.txt` file is ignored.

Fixes https://iotaledger.atlassian.net/browse/IRI-26

## Type of change

# How Has This Been Tested?

IRI was ran with `--testnet` flag. The file was skipped.

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/service/API.java
+++ b/src/main/java/com/iota/iri/service/API.java
@@ -110,7 +110,7 @@ public class API {
     }
 
     public void init() throws IOException {
-        readPreviousEpochsSpentAddresses(instance.configuration.booling(DefaultConfSettings.TESTNET));
+        readPreviousEpochsSpentAddresses(testNet);
 
         final int apiPort = instance.configuration.integer(DefaultConfSettings.PORT);
         final String apiHost = instance.configuration.string(DefaultConfSettings.API_HOST);
--- a/src/main/java/com/iota/iri/service/API.java
+++ b/src/main/java/com/iota/iri/service/API.java
@@ -110,7 +110,7 @@ public class API {
     }
 
     public void init() throws IOException {
-        readPreviousEpochsSpentAddresses();
+        readPreviousEpochsSpentAddresses(instance.configuration.booling(DefaultConfSettings.TESTNET));
 
         final int apiPort = instance.configuration.integer(DefaultConfSettings.PORT);
         final String apiHost = instance.configuration.string(DefaultConfSettings.API_HOST);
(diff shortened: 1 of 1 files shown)


src/main/java/com/iota/iri/service/API.java
```diff
@@ -110,7 +110,7 @@ public API(Iota instance, IXI ixi) {
     }
 
     public void init() throws IOException {
-        readPreviousEpochsSpentAddresses();
+        readPreviousEpochsSpentAddresses(instance.configuration.booling(DefaultConfSettings.TESTNET));
```
alon-e: why not use `testNet` field initialized in the constructor?

alon-e:
duplicate of #619 :p -- I'll close mine.

src/main/java/com/iota/iri/service/API.java
GalRogozinski: fixed

alon-e: so the reason to pass it as an argument is so it's explicit that `readPreviousEpochsSpentAddresses` depends on `testNet`? because this can also be accessed from within the method itself.

GalRogozinski: It makes it clear for the reader of `init()` that `readPreviousEpochsSpentAddresses` depends on the value of testnet. Like the `if` you had in your closed PR.

I did it this way automatically because if this method will be called again from another place in `API`, no one will forget that it depends on testnet.

You can rebut and say that this method will never be called in `API` more than once (which makes sense). In this case the way you did it in the closed PR was preferable.

footloosejava:
How long does it take?

GalRogozinski:
@footloosejava 
It took time because we are busy with some organizational reorg in IF.
Soon it will be done and we hope to achieve fast work pace.

alon-e:
I think he means how long does it take for `readPreviousEpochsSpentAddresses` to run :)

footloosejava:
Re: I think he means how long does it take for readPreviousEpochsSpentAddresses to run :)
Correct!

GalRogozinski:
hmmm,

I am not sure why testnet users should query the testnet for mainnet events. I personally think it is confusing and unexpected behavior.

Why should network A give you info on network B? If you want to get info on network B just connect to it imho.

footloosejava:
> I am not sure why testnet users should query the testnet for mainnet events. I personally think it is confusing and unexpected behavior.

I agree. It is probably unnecessary.

Issue #611, opened 2018-04-03
Incorrect tag size in TransactionViewmodel - specifies 27 bytes, whilst tagBytes is always 17 bytes


footloosejava:
Bug in code  - perhaps a typo?

Code from lines 292 to 298 in `TransactionViewModel.java`

TAG_SIZE specifies 27 bytes, whilst tagBytes is always 17 bytes.

All tests still pass when changed to 17.

```
public Hash getTagValue() {
        if(transaction.tag == null) {
            byte[] tagBytes = Converter.allocateBytesForTrits(TAG_TRINARY_SIZE);
            Converter.bytes(trits(), TAG_TRINARY_OFFSET, tagBytes, 0, TAG_TRINARY_SIZE);
            transaction.tag = new Hash(tagBytes, 0, TAG_SIZE);
        }
        return transaction.tag;
    }
```

The error is here in `TransactionViewModel.java`:

```
private static final int TAG_SIZE = 27;
```

Both `OBSOLETE_TAG_TRINARY_SIZE` and `TAG_TRINARY_SIZE` are both 81 - resulting in `TAG SIZE` = **17** (ie. Not 27).

This is also probably why this strange bounds check/adjustment was made in `Hash.java` - adjusting the bytes size to equivalent of `Math.min( size, bytes.length-offset )`:

```
private void fullRead(byte[] bytes, int offset, int size) {
        this.bytes = new byte[SIZE_IN_BYTES];
        System.arraycopy(bytes, offset, this.bytes, 0, size - offset > bytes.length ? bytes.length-offset: size);
        hashCode = Arrays.hashCode(this.bytes);
    }
```

alon-e:
@footloosejava - yeah, you're right it's a regression bug due to poor variable naming.
`TAG_SIZE` vs. `TAG_TRINARY_SIZE`.

Issue #606, opened 2018-04-02
Cannot find all the related transaction with findTransaction with Tag


yillkid:
Reproduce:

The transaction in IOTA explorer: [here](https://thetangle.org/transaction/PJMQHYNXLERTVTOWBLPKVDJ9HSTUACPURBCPJEWQDQAWGOPWZWFJMOVIXPALFW9XXSJVBZQVTFG9A9999)
* Transaction hash: PJMQHYNXLERTVTOWBLPKVDJ9HSTUACPURBCPJEWQDQAWGOPWZWFJMOVIXPALFW9XXSJVBZQVTFG9A9999
* Transaction Tag: 9JVXCXMFAMKLUQQCDACSWJVDLHC

and the transaction is issued from a full-node **http://node.deviceproof.org:14265**

Find transaction by Tag with a full-node:
`curl http://node.deviceproof.org:14265 \
   -X POST \
   -H 'Content-Type: application/json' \
   -H 'X-IOTA-API-VERSION: 1.4.1' \
   -d '{"command": "findTransactions", "tags": ["9JVXCXMFAMKLUQQCDACSWJVDLHC"]}'
`

Output:
[...]

PR #728, opened 2018-04-24
Use the Tag field when finding transaction via API.

### LINKED ISSUE #726: Use tag (not obsoleteTag) field to find transactions
### Description
Use the Tag field when finding transaction via API.

### Motivation

1. The `obsoleteTag` field will be removed in the future, `Tag` field can be set per attachment and is not part of the signature.
2. https://github.com/iotaledger/iri/issues/606#issuecomment-380700886

### Requirements
- [x] Index transactions based on `tag`, as well as existing `obsoleteTag`.

- [x] When `findTransaction` is called with `tags` fetch using `tag`, if not transactions found fetch using `obsoleteTag`.

- [x] resolve #611 

Yes
redoing #340

alon-e:
# Description

Use the Tag field when finding transaction via API.

Fixes #726 (also #340, #611, #606)

## Type of change

**This PR requires a DB migration to function properly, and is planned to be added to the next version**

# How Has This Been Tested?

- Added regression test to APIIntegrationTests that fetches using both Tag versions

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/controllers/TagViewModel.java
+++ b/src/main/java/com/iota/iri/controllers/TagViewModel.java
@@ -27,7 +27,7 @@ public class TagViewModel implements HashesViewModel {
         this.hash = hash;
     }
 
-    public static TagViewModel load(Tangle tangle, Indexable hash, Class<?> model) throws Exception {
+    private static TagViewModel load(Tangle tangle, Indexable hash, Class<?> model) throws Exception {
         return new TagViewModel((Tag) tangle.load(model, hash), hash);
     }
 
--- a/src/main/java/com/iota/iri/controllers/TagViewModel.java
+++ b/src/main/java/com/iota/iri/controllers/TagViewModel.java
@@ -27,7 +27,7 @@ public class TagViewModel implements HashesViewModel {
         this.hash = hash;
     }
 
-    private static TagViewModel load(Tangle tangle, Indexable hash, Class<?> model) throws Exception {
+    private static TagViewModel load(Tangle tangle, Indexable hash, Class<? extends Tag> model) throws Exception {
         return new TagViewModel((Tag) tangle.load(model, hash), hash);
     }
 
--- a/src/main/java/com/iota/iri/controllers/TagViewModel.java
+++ b/src/main/java/com/iota/iri/controllers/TagViewModel.java
@@ -1,9 +1,8 @@
 package com.iota.iri.controllers;
 
-import com.iota.iri.model.Address;
-import com.iota.iri.model.Bundle;
 import com.iota.iri.model.Hash;
 import com.iota.iri.model.Tag;
+import com.iota.iri.model.ObsoleteTag;
 import com.iota.iri.storage.Indexable;
 import com.iota.iri.storage.Persistable;
 import com.iota.iri.storage.Tangle;
@@ -28,8 +27,16 @@ public class TagViewModel implements HashesViewModel {
         this.hash = hash;
     }
 
+    public static TagViewModel load(Tangle tangle, Indexable hash, Class<?> model) throws Exception {
+        return new TagViewModel((Tag) tangle.load(model, hash), hash);
+    }
+
     public static TagViewModel load(Tangle tangle, Indexable hash) throws Exception {
-        return new TagViewModel((Tag) tangle.load(Tag.class, hash), hash);
+        return load(tangle, hash, Tag.class);
+    }
+
+    public static TagViewModel loadObsolete(Tangle tangle, Indexable hash) throws Exception {
+        return load(tangle, hash, ObsoleteTag.class);
     }
 
     public static Map.Entry<Indexable, Persistable> getEntry(Hash hash, Hash hashToMerge) throws Exception {
--- a/src/main/java/com/iota/iri/controllers/TransactionViewModel.java
+++ b/src/main/java/com/iota/iri/controllers/TransactionViewModel.java
@@ -166,7 +166,8 @@ public class TransactionViewModel {
         hashesList.add(new Pair<>(getBundleHash(), new Bundle(hash)));
         hashesList.add(new Pair<>(getBranchTransactionHash(), new Approvee(hash)));
         hashesList.add(new Pair<>(getTrunkTransactionHash(), new Approvee(hash)));
-        hashesList.add(new Pair<>(getObsoleteTagValue(), new Tag(hash)));
+        hashesList.add(new Pair<>(getObsoleteTagValue(), new ObsoleteTag(hash)));
+        hashesList.add(new Pair<>(getTagValue(), new Tag(hash)));
         setAttachmentData();
         setMetadata();
         return hashesList;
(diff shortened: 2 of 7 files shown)


alon-e:
@viossat this is a version of #340 that is also backward compatible with users fetching by Obsolete tag.

src/main/java/com/iota/iri/controllers/TagViewModel.java
```diff
@@ -28,8 +27,16 @@ private TagViewModel(Tag hashes, Indexable hash) {
         this.hash = hash;
     }
 
+    public static TagViewModel load(Tangle tangle, Indexable hash, Class<?> model) throws Exception {
```
GalRogozinski: why do we need `model`? 
Will it not always be `Tag.class`?

src/main/java/com/iota/iri/controllers/TransactionViewModel.java
```diff
@@ -14,7 +14,7 @@
     private final com.iota.iri.model.Transaction transaction;
 
     public static final int SIZE = 1604;
-    private static final int TAG_SIZE = 27;
+    private static final int TAG_SIZE_IN_BYTES = 17; // = ceil(81 TRITS / 5 TRITS_PER_BYTE)
```
GalRogozinski: Like on being explicit about how size is quanitified

src/main/java/com/iota/iri/model/ObsoleteTag.java
```diff
@@ -0,0 +1,12 @@
+package com.iota.iri.model;
+
+public class ObsoleteTag extends Tag {
+
+    public ObsoleteTag(Hash hash) {
+        super(hash);
+    }
+
+    public ObsoleteTag() {
```
GalRogozinski: do we need this constructor?

src/main/java/com/iota/iri/service/API.java
```diff
@@ -796,6 +797,12 @@ private synchronized AbstractResponse findTransactionStatement(final Map<String,
                 tag = padTag(tag);
                 tagsTransactions.addAll(TagViewModel.load(instance.tangle, new Hash(tag)).getHashes());
             }
+            if (tagsTransactions.isEmpty()) {
+                for (String tag : tags) {
+                    tag = padTag(tag);
+                    tagsTransactions.addAll(TagViewModel.loadObsolete(instance.tangle, new Hash(tag)).getHashes());
+                }
```
GalRogozinski: It will be nice if the new block will be encapsulated in a method

src/test/java/com/iota/iri/integration/APIIntegrationTests.java
```diff
@@ -374,14 +375,17 @@ public void shouldTestattachToTangle() {
                 body(containsString("trytes"));
     }
 
-    @Test
-    public void shouldSendTransactionAndFetch() throws InterruptedException {
+    private List<Object> sendTransfer(String[] Trytes) {
+        return sendTransfer(Trytes, NULL_HASH, NULL_HASH);
+    }
+
+    private List<Object> sendTransfer(String[] Trytes, String branch, String trunk) {
```
GalRogozinski: Trytes -> trytes or trytesArray if you don't want name conflicts

src/test/java/com/iota/iri/integration/APIIntegrationTests.java
```diff
@@ -396,27 +400,60 @@ public void shouldSendTransactionAndFetch() throws InterruptedException {
...
+    public void shouldSendTransactionAndFetchByTag() {
+
+        List<Object> trytes = sendTransfer(TRYTES);
+
+        //Tag
+        String temp = (String) trytes.get(0);
+        String[] tags = {temp.substring(TransactionViewModel.TAG_TRINARY_OFFSET / 3,
+                (TransactionViewModel.TAG_TRINARY_OFFSET + TransactionViewModel.TAG_TRINARY_SIZE) / 3)}; //extract address from trytes
+        List<Object> hashes = findTransactions("tags", tags);
+        Assert.assertTrue(!hashes.isEmpty());
```
GalRogozinski: is it possible to use `assertEquals` since you know what the tag will be?

src/test/java/com/iota/iri/integration/APIIntegrationTests.java
```diff
@@ -396,27 +400,60 @@ public void shouldSendTransactionAndFetch() throws InterruptedException {
...
+
+        //ObsoleteTag
+        String[] obsoleteTags = {temp.substring(TransactionViewModel.OBSOLETE_TAG_TRINARY_OFFSET / 3,
+                (TransactionViewModel.OBSOLETE_TAG_TRINARY_OFFSET + TransactionViewModel.OBSOLETE_TAG_TRINARY_SIZE) / 3)}; //extract address from trytes
+        hashes = findTransactions("tags", obsoleteTags);
+        Assert.assertTrue(!hashes.isEmpty());
```
GalRogozinski: ditto on the `assertEquals`

src/main/java/com/iota/iri/controllers/TagViewModel.java
alon-e: see `loadObsolete()` - loading is done with both `Tag` and `ObsoleteTag` model.
so to not repeat the intialization code.

I can either make this `private` to be more explicit or just copy the `new` call to each.

src/main/java/com/iota/iri/model/ObsoleteTag.java
alon-e: yes,
it throws `java.lang.InstantiationException: com.iota.iri.model.ObsoleteTag` if not.

src/test/java/com/iota/iri/integration/APIIntegrationTests.java
alon-e: well, the result of "findTransactions" is the hash of the transaction.
which we don't know a priori, the `trytes` can be hashed to find the transaction's hash and then compared -- adding.

src/main/java/com/iota/iri/service/API.java
alon-e: yeah, it would e nice :) , but not going to do that - not without rewriting all `findTransactionStatement`.

src/main/java/com/iota/iri/model/ObsoleteTag.java
GalRogozinski: perhaps a comment like `// used by the persistence layer to instantiate the object` is in place.
We will add that similar comment to the other `Persistable` objects in another chance.

src/main/java/com/iota/iri/controllers/TagViewModel.java
```diff
@@ -28,8 +27,16 @@ private TagViewModel(Tag hashes, Indexable hash) {
         this.hash = hash;
     }
 
+    private static TagViewModel load(Tangle tangle, Indexable hash, Class<?> model) throws Exception {
```
GalRogozinski: Class<? extends Tag>

PR #729, opened 2018-04-25
Edit snapshot - 

footloosejava:
## SNAPSHOT
(0) Replaced public lock with a private lock and lamda methods to enable access control.

(1) Removed initialSnapshot static state and replaced with a returned unmodifiable state that can be copied as needed.

(2) removed unneeded (unused) 'stateIterator.remove()'' from Snapshot.isConsistent.
This previously only applied to diffs that were immediately disposed and never used in application.

(3) moved pruning of zero values to Snapshot.init (where it also might not be necessary but that is where it does something)

(4) throws IOException if Snapshot IO occurs. This will be propagated out of the startup routine.

(5) renamed clone() to copySnapshot (clone() is reserved for Clonable, with a true clone)

(6) encapsulated state to copyState(), with readlock (copy state is only used in tests)

(7) ensured that all operations with locks are enclosed in try-finally blocks.
streamlined the map value concatenation to make it faster and read clearer.

(8) Added instance `isConsistent()` method that does not require a copy of the state.

(9) Added sum to log message if patchedDiff fails.

Some methods kept package-private because they are only used by unit tests.

ALSO:
-removed redundant bufferedInputStreams. Buffered reader already buffers a full IO buffer on each read.
-converted read function to a lambda method, eliminating a null check and loop.
-removed REGEX parsing of line and replaced with a quick index of ';' function
-converted all file operations to try with resources (automatically adds nested exceptions if on close)

## LEDGERVALIDATOR
- Adapted to new Snapshot with lambda methods instead of exposing lock.

## API
- Adapted API to new Snapshot with lambda methods instead of exposing lock.
- Changed tryte validation to use a predicate.
- Changed method-level SecureRandom to an instance level.
- Re-ordered variables in header to static at top and non-final instance vars at bottom.

## IOTA
-Adapt to new Snapshot code
- made copy during Snapshot.init(...) result and added comments.

## SNAPSHOTTEST
- Adapted to new Snapshot code.
- removed unused test.
- added @BeforeClass to create reusable snapshot instance.

## TIPSMANAGERTEST
- Adapted to new Snapshot code.
- Changed a few tests to use Assert.equals

### DIFF
--- a/src/main/java/com/iota/iri/Iota.java
+++ b/src/main/java/com/iota/iri/Iota.java
@@ -55,10 +55,13 @@ public class Iota {
         udpPort = configuration.integer(Configuration.DefaultConfSettings.UDP_RECEIVER_PORT);
         tcpPort = configuration.integer(Configuration.DefaultConfSettings.TCP_RECEIVER_PORT);
 
+        // MAKE INITIAL SNAPSHOT
         String snapshotFile = configuration.string(Configuration.DefaultConfSettings.SNAPSHOT_FILE);
         String snapshotSigFile = configuration.string(Configuration.DefaultConfSettings.SNAPSHOT_SIGNATURE_FILE);
-        Snapshot initialSnapshot = Snapshot.init(snapshotFile, snapshotSigFile, testnet).clone();
+        Snapshot initialSnapshot = Snapshot.init(snapshotFile, snapshotSigFile, testnet)
+            .copySnapshot(); // make a copy to make it mutable
         long snapshotTimestamp = configuration.longNum(Configuration.DefaultConfSettings.SNAPSHOT_TIME);
+
         int milestoneStartIndex = configuration.integer(Configuration.DefaultConfSettings.MILESTONE_START_INDEX);
         int numKeysMilestone = configuration.integer(Configuration.DefaultConfSettings.NUMBER_OF_KEYS_IN_A_MILESTONE);
         boolean dontValidateMilestoneSig = configuration.booling(Configuration.DefaultConfSettings
--- a/src/main/java/com/iota/iri/Iota.java
+++ b/src/main/java/com/iota/iri/Iota.java
@@ -1,7 +1,8 @@
 package com.iota.iri;
 
 import com.iota.iri.conf.Configuration;
-import com.iota.iri.controllers.*;
+import com.iota.iri.controllers.TipsViewModel;
+import com.iota.iri.controllers.TransactionViewModel;
 import com.iota.iri.hash.SpongeFactory;
 import com.iota.iri.model.Hash;
 import com.iota.iri.network.Node;
@@ -61,17 +62,17 @@ public class Iota {
         int milestoneStartIndex = configuration.integer(Configuration.DefaultConfSettings.MILESTONE_START_INDEX);
         int numKeysMilestone = configuration.integer(Configuration.DefaultConfSettings.NUMBER_OF_KEYS_IN_A_MILESTONE);
         boolean dontValidateMilestoneSig = configuration.booling(Configuration.DefaultConfSettings
-                .DONT_VALIDATE_TESTNET_MILESTONE_SIG);
+            .DONT_VALIDATE_TESTNET_MILESTONE_SIG);
         int transactionPacketSize = configuration.integer(Configuration.DefaultConfSettings.TRANSACTION_PACKET_SIZE);
 
         maxTipSearchDepth = configuration.integer(Configuration.DefaultConfSettings.MAX_DEPTH);
-        if(testnet) {
+        if (testnet) {
             String coordinatorTrytes = configuration.string(Configuration.DefaultConfSettings.COORDINATOR);
-            if(StringUtils.isNotEmpty(coordinatorTrytes)) {
+            if (StringUtils.isNotEmpty(coordinatorTrytes)) {
                 coordinator = new Hash(coordinatorTrytes);
             } else {
                 log.warn("No coordinator address given for testnet. Defaulting to "
-                        + Configuration.TESTNET_COORDINATOR_ADDRESS);
+                    + Configuration.TESTNET_COORDINATOR_ADDRESS);
                 coordinator = new Hash(Configuration.TESTNET_COORDINATOR_ADDRESS);
             }
         } else {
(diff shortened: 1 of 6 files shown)


src/main/java/com/iota/iri/Snapshot.java
```diff
@@ -2,175 +2,181 @@
...
-        });
-        index = newIndex;
-        rwlock.writeLock().unlock();
-    }
 
-    public static boolean isConsistent(Map<Hash, Long> state) {
-        final Iterator<Map.Entry<Hash, Long>> stateIterator = state.entrySet().iterator();
-        while (stateIterator.hasNext()) {
+            patch.forEach((key, val) ->
+                state.compute(key, (k, previous) -> Math.addExact(val, previous != null ? previous : 0L)));
```
footloosejava: For this one, I changed to `compute` because it is faster since internally `compute` keeps a hold of the Map.Entry and just changes the value.

src/main/java/com/iota/iri/Iota.java
```diff
@@ -50,61 +44,64 @@
...
         testnet = configuration.booling(Configuration.DefaultConfSettings.TESTNET);
         maxPeers = configuration.integer(Configuration.DefaultConfSettings.MAX_PEERS);
         udpPort = configuration.integer(Configuration.DefaultConfSettings.UDP_RECEIVER_PORT);
         tcpPort = configuration.integer(Configuration.DefaultConfSettings.TCP_RECEIVER_PORT);
 
+        // MAKE INITIAL SNAPSHOT
         String snapshotFile = configuration.string(Configuration.DefaultConfSettings.SNAPSHOT_FILE);
         String snapshotSigFile = configuration.string(Configuration.DefaultConfSettings.SNAPSHOT_SIGNATURE_FILE);
-        Snapshot initialSnapshot = Snapshot.init(snapshotFile, snapshotSigFile, testnet).clone();
+        Snapshot initialSnapshot = Snapshot.init(snapshotFile, snapshotSigFile, testnet)
```
footloosejava: The real change here is that since the IOException is propagated, there will no longer be a hidden RuntimeException (previous) after the `false` flag on IOException in SignedFiles and Snaphot. Effectively, the exception would have been propagated anyways, albeit invisible to callers.

src/main/java/com/iota/iri/LedgerValidator.java
```diff
@@ -223,7 +229,7 @@ private MilestoneViewModel buildSnapshot() throws Exception {
                 if (StateDiffViewModel.maybeExists(tangle, candidateMilestone.getHash())) {
                     StateDiffViewModel stateDiffViewModel = StateDiffViewModel.load(tangle, candidateMilestone.getHash());
 
-                    if (stateDiffViewModel != null && !stateDiffViewModel.isEmpty()) {
```
footloosejava: Removed because - always true.

src/main/java/com/iota/iri/LedgerValidator.java
```diff
@@ -261,9 +264,7 @@ public boolean updateSnapshot(MilestoneViewModel milestoneVM) throws Exception {
                 }
             }
             return hasSnapshot;
-        } finally {
-            milestone.latestSnapshot.rwlock.writeLock().unlock();
-        }
+        });
```
footloosejava: Changed to use new method in Snapshot `withWriteLock`.

src/main/java/com/iota/iri/Snapshot.java
```diff
@@ -2,175 +2,181 @@
...
-            checkStateHasCorrectSupply(initialState);
-            checkInitialSnapshotIsConsistent(initialState);
 
-        }
-        return initialSnapshot;
-    }
+    private final ReadWriteLock rwlock = new ReentrantReadWriteLock();
+    private final boolean immutable;
+    private final Map<Hash, Long> state;
+    private int index;
```
footloosejava: Most of the changes to Snapshot ensure that (1) state objects are never exposed (copied if needed elsewhere) and (2) that locking is internalized with access provided via lambda accepting methods.
Also, immutable flag is used to make Snapshot from static `init` method immutable.

src/main/java/com/iota/iri/Snapshot.java
```diff
@@ -2,175 +2,181 @@
...
-            new HashMap.SimpleEntry<>(hashLongEntry.getKey(), state.getOrDefault(hashLongEntry.getKey(), 0L) + hashLongEntry.getValue())
-        ).collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
-        rwlock.readLock().unlock();
-        return patch;
+    Map<Hash, Long> patchedDiff(Map<Hash, Long> diff) {
+        return withReadLockSupply(() -> {
+            Map<Hash, Long> map = new HashMap<>(diff.size());
+            diff.forEach((key, value) -> {
+                Long previous = state.getOrDefault(key, 0L);
+                map.put(key, Math.addExact(previous, value));
```
GalRogozinski: :grinning: 
looks much better than before

src/main/java/com/iota/iri/Snapshot.java
```diff
@@ -2,175 +2,181 @@
...
+            throw new UnsupportedOperationException("This snapshot is marked immutable. Make a copy if you want the 'apply' function;");
         }
         rwlock.writeLock().lock();
-        patch.entrySet().stream().forEach(hashLongEntry -> {
-            if (state.computeIfPresent(hashLongEntry.getKey(), (hash, aLong) -> hashLongEntry.getValue() + aLong) == null) {
-                state.putIfAbsent(hashLongEntry.getKey(), hashLongEntry.getValue());
+        try {
+            Long sum = patch.entrySet().stream().map(Map.Entry::getValue).reduce(Math::addExact).orElse(0L);
+            if (sum != 0L) {
+                throw new RuntimeException("Diff is not consistent. Sum is: " + sum);
```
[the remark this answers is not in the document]
GalRogozinski: @alon-e, you fine that we remove this warning from codacy?
I think it is fine to throw `RuntimeException`

src/main/java/com/iota/iri/Snapshot.java
```diff
@@ -2,175 +2,181 @@
...
-        final Iterator<Map.Entry<Hash, Long>> stateIterator = state.entrySet().iterator();
-        while (stateIterator.hasNext()) {
+            patch.forEach((key, val) ->
+                state.compute(key, (k, previous) -> Math.addExact(val, previous != null ? previous : 0L)));
 
-            final Map.Entry<Hash, Long> entry = stateIterator.next();
-            if (entry.getValue() <= 0) {
+            index = newIndex;
+        } finally {
+            rwlock.writeLock().unlock();
```
GalRogozinski: Why didn't you use `withWriteLock()`?

src/main/java/com/iota/iri/service/API.java
```diff
@@ -587,17 +582,15 @@ public static void incEllapsedTime_getTxToApprove(long ellapsedTime) {
...
-        Hash[] tips = new Hash[tipsToApprove];
-        final SecureRandom random = new SecureRandom();
-        final int randomWalkCount = numWalks > maxRandomWalks || numWalks < 1 ? maxRandomWalks:numWalks;
-        Hash referenceHash = null;
-        int maxDepth = instance.tipsManager.getMaxDepth();
-        if (depth > maxDepth) {
-            depth = maxDepth;
-        }
-        if(reference != null) {
+    public synchronized Hash[] getTransactionToApproveStatement(int theDepth, final String reference, final int numWalks) throws Exception {
```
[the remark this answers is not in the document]
GalRogozinski: Yup, this problem is not your fault

src/main/java/com/iota/iri/Snapshot.java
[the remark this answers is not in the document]
footloosejava: @GalRogozinski re: @alon-e, you fine that we remove this warning from codacy?
I kind of like it .... and it should be changed from RuntimeException to somethign else.
But ... It is no crime to throw a RuntimeException and if a better one can be substituted, then perhaps that is something to look for in a code-cleanup, not in a codacy policing file.
I vote to remove it, but also encourage users to find a better one if available.

src/main/java/com/iota/iri/Snapshot.java
footloosejava: This is the first time I see u use an emoji 😀

src/main/java/com/iota/iri/Snapshot.java
footloosejava: @GalRogozinski  If I use `withWriteLock`, then I can't access `index` and need to assign it after the lock is released.

`index = withWriteLock(....)` this will fail sync under load because someone else could read the index after waiting on the lock and before it is reassigned.

src/main/java/com/iota/iri/Snapshot.java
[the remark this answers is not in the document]
GalRogozinski: I agree that illegalArgumentException is better. If we decide that we care about those things then we shouldn't remove it from codacy. Instead when we decide that throwing `RuntimeException` is fine, we can just use the @SuppressWarning annotation in the code.

footloosejava:
@GalRogozinski Please let me know if everything is in order.
- I've re-based to the post-snapshot dev branch at the end of April.
- Merged in changes from editSnapshot previous work on PR.

GalRogozinski:
Hey,

The reason this didn't get approved is because it is a rather big PR (meaning it must have two approvers). Recent change of priorities didn't give us much time to tend to PRs unfortunately.

Anyhows since @alon-e is still being kept busy. In order to get this PR merged I would break it down to smaller patches that solve only 1 issue and are slim as possible.

If it is small enough 1 review may suffice and I will be able to merge your changes!

footloosejava:
I'll have to leave this with you now. I am too busy now with some other work.
Good luck!

GalRogozinski:
Due to refactoring during local snapshots this is no longer relevant

PR #739, opened 2018-05-01
ZMQ: Send transaction hash with trytes

viossat:
# Description

(Re)Calculating the transaction hash can be slow for the app using the ZMQ topic `tx_trytes`, I propose to add it in the ZMQ message.

## Type of change

# How Has This Been Tested?

- Custom JS script subscribing to ZMQ topic `tx_trytes`

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/storage/ZmqPublishProvider.java
+++ b/src/main/java/com/iota/iri/storage/ZmqPublishProvider.java
@@ -95,7 +95,8 @@ public class ZmqPublishProvider implements PersistenceProvider {
 
         try {
             txTrytesStringBuilder.append("tx_trytes ");
-            txTrytesStringBuilder.append(Converter.trytes(transactionViewModel.trits()));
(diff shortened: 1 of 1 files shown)


GalRogozinski:
@viossat 
We already publish the hash under the "tx" topic.
I think it is enough.

Do you have any reservations to this statement?

viossat:
You suggest to subscribe to both topics `tx` and `tx_trytes` and to implement something to match the messages received from one and from the other topic. Including the hash in `tx_trytes` would be better IMHO.

alon-e:
I agree that having the transaction hash together w/ the transaction trytes is very useful, for 2 reasons:
1. recalculating hashes takes time
2. recalculating requires the subscribing client to have the hashing logic/code

@galrogo while the `tx_trytes` topic itself might be "abusing ZMQ", the overhead of this change is not large and would make this topic more useful, so the change should be evaluated on its own merits imo.

Issue #615, opened 2018-04-06
TangleNet - Descentralized webpages


brunoamancio:
### Suggestion description
I've had the idea of allowing an IXI module to return html content to the user, so that there could be something like a "TangleNet", which is decentralized and cannot be tampered with.

Example: The transaction [BUUFWGXQYVHQNEDRHWLMRYWWKRGIJ9FDGQL9AA9XNSKYYPHJDIRBMJNQNGNUJXHMB9UYSQCDDJEJA9999 ](https://thetangle.org/transaction/BUUFWGXQYVHQNEDRHWLMRYWWKRGIJ9FDGQL9AA9XNSKYYPHJDIRBMJNQNGNUJXHMB9UYSQCDDJEJA9999) would be sent from any node with "TangleNet IXI Module" as response to any node client.

The issue here is that, currently, IRI Nodes only send JSON format as response (See methods processRequest and sendResponse in https://github.com/iotaledger/iri/blob/d59de9fbaa2432a7123d9a4d65a1555241fcd934/src/main/java/com/iota/iri/service/API.java). 

My suggestion is to allow sending HTML content as response, as well as JSON. In sum, IXI modules would dictate the response formats, in case an IXI Module was called.

### Am I planning to do it myself with a PR?
Yes, if the idea is found useful to more people.

brunoamancio:
I've been doing some research/development with https://github.com/huhn511 in this respect. :) 

We're able to have a node load web content from the Tangle and send it to the client in the correct format. Some adjustments are still necessary.

Here's the hello world of the transaction https://thetangle.org/transaction/BUUFWGXQYVHQNEDRHWLMRYWWKRGIJ9FDGQL9AA9XNSKYYPHJDIRBMJNQNGNUJXHMB9UYSQCDDJEJA9999

![hello_world](https://user-images.githubusercontent.com/3150473/39403714-8d1b8448-4b82-11e8-91ce-05aa4584385c.png)

agentpx:
like this? http://neocities.org/ --> the geocities of today (DLT)

jrdutra:
Interessante!!! Parabéns!

nullpilot:
I recently made some changes to the API to allow sending binary data from IXI: https://github.com/iotaledger/iri/issues/615 - example [here](https://github.com/nullpilot/oyster.ixi/blob/23b7fb60cbd35e585b916b201fa53a125ebf12d3/index.js)
Maybe it's interesting to someone. (I was referred to this issue from Discord)

PR #743, opened 2018-05-04
Allow IXI Modules to dictate response's content-type.

brunoamancio:
# Description

This PR allows IXI Modules to decide the content-type they wish to send to the client. The idea is that clients might want, for example, load web content directly on their browser, or get data in XML format, images, or whatever the IXI module serves.

This PR was created for issue https://github.com/iotaledger/iri/issues/615

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/service/API.java
+++ b/src/main/java/com/iota/iri/service/API.java
@@ -1201,7 +1201,7 @@ public class API {
         final HeaderMap headerMap = exchange.getResponseHeaders();
         headerMap.add(new HttpString("Access-Control-Allow-Origin"),"*");
         headerMap.add(new HttpString("Keep-Alive"), "timeout=500, max=100");
-        headerMap.put(Headers.CONTENT_TYPE, getResponseContentType(response));
+        headerMap.put(Headers.CONTENT_TYPE, getResponseContentType(res));
     }
 
     private String getResponseContentType(AbstractResponse response) {
--- a/src/main/java/com/iota/iri/service/API.java
+++ b/src/main/java/com/iota/iri/service/API.java
@@ -64,7 +64,7 @@ public class API {
 
     private Undertow server;
 
-    private final Gson gson = new GsonBuilder().create();
+    private final Gson gson = new GsonBuilder().disableHtmlEscaping().create();
     private volatile PearlDiver pearlDiver = new PearlDiver();
 
     private final AtomicInteger counter = new AtomicInteger(0);
@@ -169,24 +169,49 @@ public class API {
     }
 
     private void processRequest(final HttpServerExchange exchange) throws IOException {
-        final ChannelInputStream cis = new ChannelInputStream(exchange.getRequestChannel());
-        exchange.getResponseHeaders().put(Headers.CONTENT_TYPE, "application/json");
-
         final long beginningTime = System.currentTimeMillis();
-        final String body = IOUtils.toString(cis, StandardCharsets.UTF_8);
+        final String body = getRequestBody(exchange);
+
         final AbstractResponse response;
 
-        if (!exchange.getRequestHeaders().contains("X-IOTA-API-Version")) {
-            response = ErrorResponse.create("Invalid API Version");
-        } else if (body.length() > maxBodyLength) {
+        if (body.length() > maxBodyLength) {
             response = ErrorResponse.create("Request too long");
         } else {
-            response = process(body, exchange.getSourceAddress());
+            response = process(body, exchange);
         }
+
         sendResponse(exchange, response, beginningTime);
     }
 
-    private AbstractResponse process(final String requestString, InetSocketAddress sourceAddress) throws UnsupportedEncodingException {
+    private String getRequestBody(final HttpServerExchange exchange) throws IOException {
+        final ChannelInputStream cis = new ChannelInputStream(exchange.getRequestChannel());
+        String body = IOUtils.toString(cis, StandardCharsets.UTF_8);
+
+        if(body.length() == 0){
+            body = getQueryParamsBody(exchange.getQueryParameters());
+        }
+        return body;
+    }
+
+    private String getQueryParamsBody(Map<String, Deque<String>> queryParameters) {
+        String queryParamsBody = "{";
+
+        Set<String> keySet = queryParameters.keySet();
+
+        for (String key : keySet) {
+            String param = queryParameters.get(key).getFirst();
+            queryParamsBody += "\"" + key + "\" : " + "\"" + param + "\","; // Json property
+        }
+        
+        // Removes last comma, if multiple params
+        if(queryParamsBody.endsWith(",")){
+            queryParamsBody = queryParamsBody.substring(0, queryParamsBody.length() -1);
+        }
+
+        return queryParamsBody + "}";
+	}
+
+    private AbstractResponse process(final String requestString, final HttpServerExchange exchange) throws UnsupportedEncodingException {
 
         try {
 
@@ -201,7 +226,7 @@ public class API {
             }
 
             if (instance.configuration.string(DefaultConfSettings.REMOTE_LIMIT_API).contains(command) &&
-                    !sourceAddress.getAddress().isLoopbackAddress()) {
+                    !exchange.getSourceAddress().getAddress().isLoopbackAddress()) {
                 return AccessLimitedResponse.create("COMMAND " + command + " is not available on this node");
             }
 
@@ -209,6 +234,10 @@ public class API {
 
             switch (command) {
                 case "storeMessage": {
+                    if(!iotaAPIHeaderDefined(exchange)){
+                        return ErrorResponse.create("Invalid API Version");
+                    }
+
                     if (!testNet) {
                         return AccessLimitedResponse.create("COMMAND storeMessage is only available on testnet");
                     }
@@ -230,11 +259,19 @@ public class API {
                 }
 
                 case "addNeighbors": {
+                    if(!iotaAPIHeaderDefined(exchange)){
+                        return ErrorResponse.create("Invalid API Version");
+                    }
+
                     List<String> uris = getParameterAsList(request,"uris",0);
                     log.debug("Invoking 'addNeighbors' with {}", uris);
                     return addNeighborsStatement(uris);
                 }
                 case "attachToTangle": {
+                    if(!iotaAPIHeaderDefined(exchange)){
+                        return ErrorResponse.create("Invalid API Version");
+                    }
+
                     final Hash trunkTransaction  = new Hash(getParameterAsStringAndValidate(request,"trunkTransaction", HASH_SIZE));
                     final Hash branchTransaction = new Hash(getParameterAsStringAndValidate(request,"branchTransaction", HASH_SIZE));
                     final int minWeightMagnitude = getParameterAsInt(request,"minWeightMagnitude");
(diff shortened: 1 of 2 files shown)


brunoamancio:
To test the IXI part, you can take https://github.com/brunoamancio/TangleNet , an IXI module which provides the client with web content from the Tangle.

src/main/java/com/iota/iri/service/API.java
```diff
@@ -169,24 +169,49 @@ private void readPreviousEpochsSpentAddresses(boolean isTestnet) throws IOExcept
...
+        
+        // Removes last comma, if multiple params
+        if(queryParamsBody.endsWith(",")){
+            queryParamsBody = queryParamsBody.substring(0, queryParamsBody.length() -1);
+        }
+
+        return queryParamsBody + "}";
+	}
+
+    private AbstractResponse process(final String requestString, final HttpServerExchange exchange) throws UnsupportedEncodingException {
```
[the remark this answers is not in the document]
brunoamancio: This method is indeed long. The way commands are structured in a switch is not clean, but that is an existing issue in the code.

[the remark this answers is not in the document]
brunoamancio: Same as before - This method is indeed long. The way commands are structured in a switch is not clean, but that is an existing issue in the code.

src/main/java/com/iota/iri/service/API.java
```diff
@@ -169,24 +168,49 @@ private void readPreviousEpochsSpentAddresses(boolean isTestnet) throws IOExcept
...
```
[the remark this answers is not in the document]
brunoamancio: That's a complex method. The commands execution must be refactored not to use a switch with multi-line cases.

src/main/java/com/iota/iri/service/API.java
```diff
@@ -169,24 +168,49 @@ private void readPreviousEpochsSpentAddresses(boolean isTestnet) throws IOExcept
...
+        return body;
+    }
+
+    private String getQueryParamsBody(Map<String, Deque<String>> queryParameters) {
+        String queryParamsBody = "{";
+
+        Set<String> keySet = queryParameters.keySet();
+
+        for (String key : keySet) {
+            String param = queryParameters.get(key).getFirst();
```
GalRogozinski: I don't like concatenated gets...
Can you seperate to two lines?

src/main/java/com/iota/iri/service/API.java
```diff
@@ -169,24 +168,49 @@ private void readPreviousEpochsSpentAddresses(boolean isTestnet) throws IOExcept
...
+            queryParamsBody += "\"" + key + "\" : " + "\"" + param + "\","; // Json property
```
GalRogozinski: We are still using Java 8,
So can you use StringBuilder?

src/main/java/com/iota/iri/service/API.java
```diff
@@ -201,14 +225,18 @@ private AbstractResponse process(final String requestString, InetSocketAddress s
             }
 
             if (instance.configuration.string(DefaultConfSettings.REMOTE_LIMIT_API).contains(command) &&
-                    !sourceAddress.getAddress().isLoopbackAddress()) {
+                    !exchange.getSourceAddress().getAddress().isLoopbackAddress()) {
```
GalRogozinski: concatenated gets...

src/main/java/com/iota/iri/service/API.java
```diff
@@ -201,14 +225,18 @@ private AbstractResponse process(final String requestString, InetSocketAddress s
...
                 return AccessLimitedResponse.create("COMMAND " + command + " is not available on this node");
             }
 
             log.debug("# {} -> Requesting command '{}'", counter.incrementAndGet(), command);
 
             switch (command) {
                 case "storeMessage": {
+                    if(!iotaAPIHeaderDefined(exchange)){
```
GalRogozinski: Does this always appear?
Instead of copy/pasting it can you write it once before the `switch`?

src/main/java/com/iota/iri/service/API.java
```diff
@@ -1069,6 +1170,21 @@ private void sendResponse(final HttpServerExchange exchange, final AbstractRespo
         sinkChannel.resumeWrites();
     }
 
+    private String convertResponseToClientFormat(AbstractResponse res) {
+        String response = null;
+        if(res instanceof IXIResponse){
+            final String content = ((IXIResponse)res).getContent();
+            if(content != null && !content.equals("")){
```
GalRogozinski: Should `"   "` pass?

Use `StringUtils.isNotEmpty()` or `StringUtils.isNotBlank()` according to your answer

src/main/java/com/iota/iri/service/dto/IXIResponse.java
```diff
@@ -15,4 +17,24 @@ public static IXIResponse create(Object myixi) {
     public Object getResponse() {
         return ixi;
     }
+
+    private String getdefaultContentType() {
+        return "application/json";
+    }
+
+    public String getResponseContentType() {
+        Object fieldObj = getResponseMapper().get("contentType");
```
GalRogozinski: concatenated gets

src/main/java/com/iota/iri/service/dto/IXIResponse.java
```diff
@@ -15,4 +17,24 @@ public static IXIResponse create(Object myixi) {
...
+        String fieldValue = fieldObj == null || "".equals(fieldObj) ? getdefaultContentType() : fieldObj.toString();
```
GalRogozinski: Should `"   "` pass?

Use `StringUtils.isNotEmpty()` or `StringUtils.isNotBlank()` according to your answer

src/main/java/com/iota/iri/service/dto/IXIResponse.java
```diff
@@ -15,4 +17,24 @@ public static IXIResponse create(Object myixi) {
...
+        return fieldValue;
+    }
+
+    private Map<String, Object> getResponseMapper(){
+        return (Map<String, Object>)ixi;
+    }
+
+    public String getContent() {
+        Object fieldObj = getResponseMapper().get("content");
+        String fieldValue = fieldObj == null || "".equals(fieldObj) ? null : fieldObj.toString();
```
GalRogozinski: Should `"   "` pass?

Use `StringUtils.isNotEmpty()` or `StringUtils.isNotBlank()` according to your answer

src/main/java/com/iota/iri/service/API.java
brunoamancio: I'm going to refactor this. But the idea is not to require the api version header for requests to IXI modeules.

brunoamancio: The idea is not to require the API version header for calls to IXI modules. Due to the current use of a switch we have 3 options: 
1 - Paste the check for every API call, except call to IXI modules (which is what I've done)
2 - Refactor the current logic to use command pattern
3 - Use a Map structure to map command name to API call (kind of a command pattern)

I can refactor the command structure, but that should be done in another PR.

src/main/java/com/iota/iri/service/API.java
brunoamancio: I will do that, although the extra variable will be set even when the first condition isn't met. That means we are trading performance for maintainability.

src/main/java/com/iota/iri/service/API.java
```diff
@@ -169,24 +169,43 @@ private void readPreviousEpochsSpentAddresses(boolean isTestnet) throws IOExcept
...
+        for (String key : queryParameters.keySet()) {
+            Deque<String> dequeuedParameter = queryParameters.get(key);
+            String parameterValue = dequeuedParameter.getFirst();
+            parametersMapper.put(key, parameterValue);
+        }
+
+        return gson.toJson(parametersMapper);
+	}
+
+    private AbstractResponse process(final String requestString, final HttpServerExchange exchange) throws UnsupportedEncodingException {
```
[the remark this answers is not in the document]
brunoamancio: @alon-e As I've written to @GalRogozinski :

The idea is to not require the API version header for calls to IXI modules. Due to the current use 
of a switch we have 3 options:
1 - Paste the check for every API call, except call to IXI modules (which is what I've done)
2 - Refactor the current logic to use command pattern
3 - Use a Map structure to map command names to API calls (kind of a command pattern)

I can refactor the command structure, but that should be done in another PR.

Do you have any suggestions related to my changes here? The complexity of this method is already high with the current implementation. This issue can only be better implemented with the suggested refactoring of the command pattern.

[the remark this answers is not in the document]
brunoamancio: Same as the comment below.

footloosejava:
@brunoamancio Pretty cool mate!
What do you see as the end-user application? When would someone request web pages from the network?

brunoamancio:
@footloosejava With these changes, acessing tangle-only and hybrid hosted web content will be a reality. I have already implemented a simple IXI module (https://github.com/brunoamancio/TangleNet) to allow nodes provide that to clients sending requests directly on their browsers. 

I plan on developing it further to make this process as user-friendly as possible.

src/main/java/com/iota/iri/service/API.java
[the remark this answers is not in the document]
yovendielmundo: Using my PR you get rid of the complexity of having everything in one huge function. each command is a single function. adding and removing functionality is super easy, you just need to remove or add new methods.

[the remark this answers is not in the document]
brunoamancio: @yovendielmundo could you check if your branch is up-to-date, and that all api commands are set-up correctly? :)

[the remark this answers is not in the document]
yovendielmundo: I have to update my branch. I did the PR long long time ago and everything has change a lot. I was not expecting someone using this.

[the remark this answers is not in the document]
brunoamancio: @yovendielmundo Could you do that, please? I think this is a very much needed refactoring.

GalRogozinski:
@alon-e 
Can you also review please?

brunoamancio:
I will refactor the command logic in another PR. Until then, this PR will be left as is.

alon-e:
@brunoamancio I'll review this shortly.
a few remarks:

1. have you tested backward compatibility with existing IXI modules?
examples: 
https://github.com/bluedigits/nstats
https://github.com/iotaledger/Snapshot.ixi
https://github.com/iotaledger/MAM.ixi

2. adding `disableHtmlEscaping()` is alarming, could you elaborate on the need and how do you mitigate attacks against input validation? for example, reflective XSS via ```"COMMAND " + command + " is not available on this node"```? (I'll be looking into this as well)

3. the `iotaAPIHeaderDefined()` additions are too much to swallow IMHO. the idea is to make IRI more readable and maintainable, there must be a better way to do that. 
I understand the reason is to not make this PR more complex, but in that case, we should first change the command routing (a la #392) and then redactor this PR. - i.e. use this as a PoC, but not merge it just yet.

brunoamancio:
Answering @alon-e questions:

1 - In case the response does not have the specific fields reserved for specifying the content-type of the response, the behavior is exactly as before.

That means, only if code is like the following will this feature be used:

... javascript  code of IXI module

`return Response.create({ content-type: "specific-content/type", content : "response in the response content-type format"})`
... 

Anyway, I can run those and tell you the results in a few days as well.

2 - The reason is that html content (and any other content format) is converted from the javascript-generated response from the IXI Module to IXIResponse using gson, which makes a desired response with html to be unreadable by the IRI client, since it's escaped.

More details on gson the API.java (where escaping is disabled):
- When gson is used to convert query parameter strings to json, the reason is to match the post request body structure, which is a string with json content. I'll discuss this in the respective point below, since any issues should be handled the same way.

- When the request body has unescaped values, the only case where reflective XSS is possible is the example you gave: ErrorResponse.create("Command [" + command + "] is unknown").
In this case, it is already possible to send in whatever one wants to in the post request body. But, since it is only used to send an error message back to the client, there is no issue.

- When gson is used to convert a response to client format (method convertResponseToClientFormat(...)) there is no issue, since the client will get the response content in the expected format and knows how to handle it. It's just a conveyor belt from the IXI Module in this case.

There are no other usages of gson in the API.java class

3 - I completely agree with that. I don't currently have time for the refactoring of the command logic, but if nothing is checked until I have time, I will do it myself in a separate PR and after the merge this PR can be updated with the better structure.

brunoamancio:
**Please don't delete this comment**
Note to myself: 
In class IXIResponse methods getContent() and getResponseContentType(), ensure variables are empty when properties "content" and "contentType" are not strings in IXI Module (they might be object or number).

GalRogozinski:
@brunoamancio 
Did you perform the tests @alon-e asked you to?

xn3cr0nx:
Hi everyone, I'm interested in this topic. I opened a new PR (#1385) with some little improvements to keep discussing about it and I hope we could be able to eventually implement this feature.

PR #753, opened 2018-05-10
Add variable naming rules to checkstyle

GalRogozinski:
# Description

Adding checkstyle rules for variable naming.

## Type of change

- Style Fix

# How Has This Been Tested?

Travis checks are enough since it is only a style fix

# Checklist:


src/main/java/com/iota/iri/service/TipsManager.java
```diff
@@ -25,20 +25,20 @@
     private final boolean testnet;
     private final int milestoneStartIndex;
 
-    private int RATING_THRESHOLD = 75; // Must be in [0..100] range
+    private int ratingThreshold = 75; // Must be in [0..100] range
```
karimodm: The conflict is just on this line, it's an easy fix. Let's merge this PR after the fix.

Issue #749, opened 2018-05-09
Unnecessary Checked Exceptions  from PersistenceProvider


footloosejava:
Given that:

- None of these exceptions are recoverable.

And also given that:

- They prevent Lambdas and Lazy code from being used in any downstream analysis without verbose and futile exception handling.

- No Streams or Annonymous classes can be used downstream.

I submit that:

- These should all be thrown as RuntimeExceptions or as a custom UnChecked exceptions
- They should be handled where required, which is where all the try{}catch{} blocks are **already**.

That would free up a lot of intermediate code to do things better using more convenient constructs.

As an added benefit:

- This would eliminate a ton of `throws Exception` from almost every method used in the process of analyzing and working with the Tangle and almost every `____ViewModel` class in the code-base.

**For your consideration.**

```
public interface PersistenceProvider {

    void init() throws Exception;
    boolean isAvailable();
    void shutdown();
    boolean save(Persistable model, Indexable index) throws Exception;
    void delete(Class<?> model, Indexable  index) throws Exception;

    boolean update(Persistable model, Indexable index, String item) throws Exception;

    boolean exists(Class<?> model, Indexable key) throws Exception;

    Pair<Indexable, Persistable> latest(Class<?> model, Class<?> indexModel) throws Exception;

    Set<Indexable> keysWithMissingReferences(Class<?> modelClass, Class<?> otherClass) throws Exception;

    Persistable get(Class<?> model, Indexable index) throws Exception;

    boolean mayExist(Class<?> model, Indexable index) throws Exception;

    long count(Class<?> model) throws Exception;

    Set<Indexable> keysStartingWith(Class<?> modelClass, byte[] value);

    Persistable seek(Class<?> model, byte[] key) throws Exception;

    Pair<Indexable, Persistable> next(Class<?> model, Indexable index) throws Exception;
    Pair<Indexable, Persistable> previous(Class<?> model, Indexable index) throws Exception;

    Pair<Indexable, Persistable> first(Class<?> model, Class<?> indexModel) throws Exception;

    boolean saveBatch(List<Pair<Indexable, Persistable>> models) throws Exception;

    void clear(Class<?> column) throws Exception;
    void clearMetadata(Class<?> column) throws Exception;
}
```

footloosejava:
I won't have time to work on this for some while. You guys sound  like you have it all under control anyways. I'm closing this issue.

GalRogozinski:
I think this is still a good issue.
[...]

PR #829, opened 2018-06-22
Error in log statement

gjeee:
small trivial fix

### DIFF
--- a/src/main/java/com/iota/iri/service/API.java
+++ b/src/main/java/com/iota/iri/service/API.java
@@ -361,7 +361,7 @@ public class API {
             log.info("API Validation failed: " + e.getLocalizedMessage());
             return ErrorResponse.create(e.getLocalizedMessage());
         } catch (final Exception e) {
-            log.error("API Exception: ", e);
+            log.error("API Exception: " + e.getLocalizedMessage())
             return ExceptionResponse.create(e.getLocalizedMessage());
         }
     }


src/main/java/com/iota/iri/service/API.java
```diff
@@ -361,7 +361,7 @@ private AbstractResponse process(final String requestString, InetSocketAddress s
             log.info("API Validation failed: " + e.getLocalizedMessage());
             return ErrorResponse.create(e.getLocalizedMessage());
         } catch (final Exception e) {
-            log.error("API Exception: ", e);
+            log.error("API Exception: " + e.getLocalizedMessage())
```
rajivshah3: Don't forget the semicolon ;)

https://travis-ci.org/iotaledger/iri/jobs/395274421#L534

gjeee: oops :-)

GalRogozinski:
Hmm since we are catching a non-typed exception it may be helpful to see the stack trace...

@gjeee, do you have lots of ugly stack traces in your node? What motivated you to fix that?

gjeee:
well the old code didn't show anything......what i did:

log.error("API Exception: ", e);   ----> log.error("API Exception: " + e.getLocalizedMessage());

but if you want a stacktrace i can do so:

log.error("API Exception: ", e);   ----> log.error("API Exception: {}", e);

GalRogozinski:
ok, I understand better now.

When we catch a general `Exception` we assume that
1) It is probably unexpected
2) We want the stack trace for easy debugging.

Since most of the time you are catching the `InvalidAlgorithmParameterException`: You can write the following code:
Where we run `getTransactionsToApprove` add before we catch the `RuntimeException` the following
```
catch(InvalidAlgorithmParameterException e) {
             log.error("API Exception: " + e.getLocalizedMessage());
             return ExceptionResponse.create(e.getLocalizedMessage());
```

And where you did the change write
```
catch(Exception e) {
            log.error("API Exception: " + e.getLocalizedMessage(), e);
            return ExceptionResponse.create(e.getLocalizedMessage());
```
Note a few things:
1. `log.error("API Exception: {}", e.getLocalizedMessage(), e);` doesn't work. It annoys me that they don't have this signature available.
2. `log.error("API Exception: " + e.getLocalizedMessage(), e);` will print the message twice which is a little ugly but you convinced me that the merits of `grep` justify it.
3. `Exception` in IRI should be general but unfortunately it also catches `RocksDb Exceptions`. See #749.
4. I actually prefer `e.getMessage()` just because I don't ever think we will ever use localized messages. `e.getLocalizedMessage()` is just misleading and creates longer code. However other parts of the code contain that string so it is ok for now and definitely shouldn't be changed in this PR.

gjeee:
ok @GalRogozinski i incorporated all your feedback :-) 
regarding point 1) the new signature of log.info is possible since SLF4J 1.6.0. see also https://www.slf4j.org/faq.html#paramException

Issue #724, opened 2018-04-23
Configuration rework


alon-e:
### Description
Rewrite configuration to be accessible by any class and be more user-friendly to non-devs to alter values.

### Requirements

- [ ] Globally available (singleton)
- [ ] Backwards compatible with flags - mandatory
- [ ] Backwards compatible with ini - if possible
- [ ] Config options should be declared in a non-java file, so non-devs can edit it.
- [ ] If possible, all flags should be configured in a text file
- [ ] The documentation of the configs should be part of the file that declares them
- [ ] Type inference of the config option 
- [ ] IDE completion of available options? 
- [ ] Plan defaults of testnets. A separate configuration file for it?

### Open Questions (optional)
Which framework best fills these needs - #723 

### Am I planning to do it myself with a PR?
Yes

footloosejava:
>>  Globally available (singleton)

Doing this eliminates custom testing because of static state and leads to problems later.

Suggested: make sure config is either (1) injected on preferably a constructor or (2) provided via a constructor.  Also make sure configuration is 'locked in' at certain point in code to ensure nothing changes.

Although a 'singleton' is tempting ... it only leads to problems when configurations need changing for various needs. Also, Singletons often have interface and  are globally available - even to code that is not supposed to see certain values.

A properly designed code-base never needs static state and ensuring that makes your job easier down the road when things change. 🙈 Static state configurations are the enemy of flexible testing and encapsulation 🙈 .

All the other requirements look 👍 .

GalRogozinski:
Thanks for your comments!

Yes, we should pass the configuration object as a constructor argument.
The other option is to use a DI framework (ie. Guice). I am currently leaning against this since IRI's object dependency graph isn't very complex. So I don't want to add the complexity of DI. If we see that IRI's codebase will start growing then we can rethink that. If you disagree with me please feel free to retort in a comment.

Since I am leaning against DI, I didn't choose `annoconf` since I understood it is dependent on Guice (I looked at the example file and saw it used Guice's `Injector`). I am going with `Jcommander`, it is similar to `commons CLI` but I like its use of annotations.

A word about IXI (since you mentioned it), I am currently thinking to resolve its scoping issues when we migrate\upgrade to java 9 modules. If you have interesting thoughts about this I will be glad to hear them.

footloosejava:
@GalRogozinski No problem.
Just to clarify, annoconf does not depend on Guice, although it can be used very conveniently with Guice.

It can also be used like:

```
[...]

PR #883, opened 2018-07-23
Fail early on below max depth - WIP

### LINKED ISSUE #846: Tip-Selection freezes when stitching sidetangle to the maintangle
### Bug description
Tip-Selection/`getTransactionsToApprove` freezes on IRI nodes in the network after a sidetangle gets stitched to the main tangle. Both on IRI 1.4.x and 1.5.x
*-> It's not clear whether all 1.4.x versions tip-sel freeze.*

### Steps To Reproduce
Write a program which connects to a ZMQ feed of any IRI node and let the program up on a new sidetangle tx create a new tx referencing the sidetangle tx and a maintangle tx. (I'm happy to share the exact program if needed; idea was to get the sidetangle confirmed but rather this phenomenon unfortunately happend)

Shown in this picture (Stitching txs = orange; Note that the bigger blow ball structure is the maintangle and the leaner structure the sidetangle):
image

### Expected behaviour
Confirmation rate should be steady and tip selection shouldn't freeze.

### Actual behavior
**IRI nodes in the network** become unable to execute `getTransactionsToApprove`.
**A clear almost halt in the Tangle network was observed (obviously, because getting tips was impossible).**
Maybe having the sheer amount of txs of the sidetangle suddenly in the tip-sel might cause it to not complete.

A clear suppression of TPS can be observed in following stats (1:30pm-3:10pm):
image

Starting the stitching at 4pm resulted in a very abrupt decline of incoming txs:
image

GalRogozinski:
# Description
If we analyze transactions in `belowMaxDepth` and we pass a certain threshold of transactions we fail.
We also cache the analyzed transactions so that in the future we can quickly fail `belowMaxDepth`.

Fixes #846 

## Type of change

# How Has This Been Tested?

Unit tests were added:
- Long chain tangle
  - Test from genesis
  - Old snapshot index
  - New snapshot index
- Short chain tangle
  - Test from genesis
  - Old snapshot index
  - New snapshot index

# Checklist:

# TODO:
1. What should be the threshold values?
MAX_CACHE_SIZE? MAX_ANALYZED_TXS?
2. Write tests for bounded set
3. Run canaries on network
4. Document `belowMaxDepth`

### DIFF
--- a/src/main/java/com/iota/iri/IRI.java
+++ b/src/main/java/com/iota/iri/IRI.java
@@ -154,8 +154,8 @@ public class IRI {
             final Option<Integer> milestoneStartIndex = parser.addIntegerOption("milestone-start");
             final Option<Integer> milestoneKeys = parser.addIntegerOption("milestone-keys");
             final Option<Long> snapshotTime = parser.addLongOption("snapshot-timestamp");
-            final Option<Integer> belowMaxDepthTxLimit = parser.addIntegerOption("below-max-depth");
-
+            final Option<Integer> belowMaxDepthTxLimit = parser.addIntegerOption("max-depth-tx-limit");
+            final Option<Integer> walkValidatorCacheSize = parser.addIntegerOption("walk-validator-cache");
 
             try {
                 parser.parse(args);
@@ -336,6 +336,11 @@ public class IRI {
             if (belowMaxDepthLimit != null) {
                 configuration.put(DefaultConfSettings.BELOW_MAX_DEPTH_TRANSACTION_LIMIT,
                         String.valueOf(belowMaxDepthLimit));
+            }
+
+            final Integer walkValidatorCache = parser.getOptionValue(walkValidatorCacheSize);
+            if (walkValidatorCache != null) {
+                configuration.put(DefaultConfSettings.WALK_VALIDATOR_CACHE_SIZE, String.valueOf(walkValidatorCache));
             }
 
             return true;
--- a/src/main/java/com/iota/iri/IRI.java
+++ b/src/main/java/com/iota/iri/IRI.java
@@ -154,6 +154,7 @@ public class IRI {
             final Option<Integer> milestoneStartIndex = parser.addIntegerOption("milestone-start");
             final Option<Integer> milestoneKeys = parser.addIntegerOption("milestone-keys");
             final Option<Long> snapshotTime = parser.addLongOption("snapshot-timestamp");
+            final Option<Integer> belowMaxDepthTxLimit = parser.addIntegerOption("below-max-depth");
 
 
             try {
@@ -330,7 +331,15 @@ public class IRI {
             if (vmaxPeers != null) {
                 configuration.put(DefaultConfSettings.MAX_PEERS, vmaxPeers);
             }
+
+            final Integer belowMaxDepthLimit = parser.getOptionValue(belowMaxDepthTxLimit);
+            if (belowMaxDepthLimit != null) {
+                configuration.put(DefaultConfSettings.BELOW_MAX_DEPTH_TRANSACTION_LIMIT,
+                        String.valueOf(belowMaxDepthLimit));
+            }
+
             return true;
+
         }
 
         private static void printUsage() {
--- a/src/main/java/com/iota/iri/service/tipselection/WalkValidator.java
+++ b/src/main/java/com/iota/iri/service/tipselection/WalkValidator.java
@@ -1,12 +1,22 @@
 package com.iota.iri.service.tipselection;
 
+import com.iota.iri.controllers.TransactionViewModel;
 import com.iota.iri.model.Hash;
+import com.iota.iri.utils.collections.impl.BoundedHashSet;
+import com.iota.iri.utils.collections.interfaces.BoundedSet;
+
+import java.util.LinkedHashSet;
+import java.util.Set;
 
 /**
  * Validates consistency of tails.
  */
 @FunctionalInterface
 public interface WalkValidator {
+
+    int MAX_CACHE_SIZE = 2_000_000;
+    //As long as tip selection is synchronized we are fine with the collection not being thread safe
+    BoundedSet<Hash> FAILED_BELOW_MAX_DEPTH_CACHE = new BoundedHashSet<>(20_000, MAX_CACHE_SIZE);
 
     /**
      * Validation
--- a/src/main/java/com/iota/iri/service/tipselection/WalkValidator.java
+++ b/src/main/java/com/iota/iri/service/tipselection/WalkValidator.java
@@ -1,22 +1,12 @@
 package com.iota.iri.service.tipselection;
 
-import com.iota.iri.controllers.TransactionViewModel;
 import com.iota.iri.model.Hash;
-import com.iota.iri.utils.collections.impl.BoundedHashSet;
-import com.iota.iri.utils.collections.interfaces.BoundedSet;
-
-import java.util.LinkedHashSet;
-import java.util.Set;
 
 /**
  * Validates consistency of tails.
  */
 @FunctionalInterface
 public interface WalkValidator {
-
-    int MAX_CACHE_SIZE = 2_000_000;
-    //As long as tip selection is synchronized we are fine with the collection not being thread safe
-    BoundedSet<Hash> FAILED_BELOW_MAX_DEPTH_CACHE = new BoundedHashSet<>(20_000, MAX_CACHE_SIZE);
 
     /**
      * Validation
--- a/src/main/java/com/iota/iri/service/tipselection/impl/WalkValidatorImpl.java
+++ b/src/main/java/com/iota/iri/service/tipselection/impl/WalkValidatorImpl.java
@@ -28,11 +28,9 @@ import java.util.*;
  */
 public class WalkValidatorImpl implements WalkValidator {
 
-    //calculate this magic number
-    public static final int MAX_CACHE_SIZE = 2_000_000;
+    public static final int INITIAL_CACHE_CAPACITY = 10_000;
     //As long as tip selection is synchronized we are fine with the collection not being thread safe
-    private static final BoundedSet<Hash> FAILED_BELOW_MAX_DEPTH_CACHE = new BoundedSetWrapper<>(
-            new LinkedHashSet<>(10_000), MAX_CACHE_SIZE);
+    private static BoundedSet<Hash> failedBelowMaxDepthCache;
     private int maxAnalyzedTxs;
 
     private final Tangle tangle;
@@ -48,7 +46,7 @@ public class WalkValidatorImpl implements WalkValidator {
     private Set<Hash> myApprovedHashes;
 
     public WalkValidatorImpl(Tangle tangle, LedgerValidator ledgerValidator, TransactionValidator transactionValidator,
-                             Milestone milestone, int maxDepth, int maxAnalyzedTxs) {
+                             Milestone milestone, int maxDepth, int maxAnalyzedTxs, int cacheSize) {
         this.tangle = tangle;
         this.ledgerValidator = ledgerValidator;
         this.transactionValidator = transactionValidator;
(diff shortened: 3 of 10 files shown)


src/main/java/com/iota/iri/service/tipselection/impl/WalkValidatorImpl.java
```diff
@@ -75,19 +82,30 @@ public boolean isValid(Hash transactionHash) throws Exception {
...
         Hash hash;
         while ((hash = nonAnalyzedTransactions.poll()) != null) {
+            if (FAILED_BELOW_MAX_DEPTH_CACHE.contains(hash)) {
+                updateCache(analyzedTransactions);
+                return true;
+            }
+            if (analyzedTransactions.size() == MAX_ANALYZED_TXS) {
+                updateCache(analyzedTransactions);
+                return true;
+            }
```
rajivshah3: If you wanted to make it a bit more concise, I think you could combine these into one `if` statement:
```java
if (FAILED_BELOW_MAX_DEPTH_CACHE.contains(hash) || analyzedTransactions.size() == MAX_ANALYZED_TXS) {
  updateCache(analyzedTransactions);
  return true;
}
```
But I guess that's a matter of preference

src/main/java/com/iota/iri/service/tipselection/WalkValidator.java
```diff
@@ -1,13 +1,23 @@
...
+import java.util.LinkedHashSet;
+import java.util.Set;
 
 /**
  * Validates consistency of tails.
  */
 @FunctionalInterface
 public interface WalkValidator {
 
+    int MAX_CACHE_SIZE = 2_000_000;
```
paulhandy: I'd suggest to beware of the tyranny of the default magic number; wouldn't this do best as a configuration parameter with a default value?

src/main/java/com/iota/iri/service/tipselection/impl/WalkValidatorImpl.java
```diff
@@ -26,6 +28,12 @@
  */
 public class WalkValidatorImpl implements WalkValidator {
 
+    public static final int MAX_CACHE_SIZE = 2_000_000;
+    //As long as tip selection is synchronized we are fine with the collection not being thread safe
+    private static final BoundedSet<Hash> FAILED_BELOW_MAX_DEPTH_CACHE = new BoundedSetWrapper<>(
```
th0br0: You're allocating a `FAILED_BELOW_MAX_DEPTH_CACHE` instance both in the impl and the interface with different evil magic numbers...

src/main/java/com/iota/iri/service/tipselection/impl/WalkValidatorImpl.java
```diff
@@ -75,19 +82,30 @@ public boolean isValid(Hash transactionHash) throws Exception {
         return true;
     }
 
-    private boolean belowMaxDepth(Hash tip, int depth) throws Exception {
+    private boolean belowMaxDepth(Hash tip, int allowedSnapshotIndex) throws Exception {
```
th0br0: could we call this `minimumAllowedSnapshotIndex` or `lower...`?

src/main/java/com/iota/iri/service/tipselection/impl/WalkValidatorImpl.java
```diff
@@ -75,19 +82,30 @@ public boolean isValid(Hash transactionHash) throws Exception {
...
             return false;
         }
         //if tip unconfirmed, check if any referenced tx is confirmed below maxDepth
         Queue<Hash> nonAnalyzedTransactions = new LinkedList<>(Collections.singleton(tip));
         Set<Hash> analyzedTransactions = new HashSet<>();
         Hash hash;
         while ((hash = nonAnalyzedTransactions.poll()) != null) {
+            if (FAILED_BELOW_MAX_DEPTH_CACHE.contains(hash)) {
+                updateCache(analyzedTransactions);
+                return true;
```
th0br0: if it `failed` before (I'm assuming that that is what the cahce is for) shouldn't this return `false`?

th0br0: Could you add a javadoc comment to this method as you're touching it?

src/main/java/com/iota/iri/utils/collections/impl/BoundedSetWrapper.java
```diff
@@ -0,0 +1,173 @@
...
+
+    @Override
+    public <T> T[] toArray(T[] a) {
+        return delegate.toArray(a);
+    }
+
+    @Override
+    public boolean add(E e) {
+        if (isFull()) {
+            Iterator<E> iterator = delegate.iterator();
```
th0br0: this violates the description `A set that doesn't allow to add elements to it once it is full`

src/main/java/com/iota/iri/utils/collections/impl/BoundedSetWrapper.java
```diff
@@ -0,0 +1,173 @@
...
+            return delegate.addAll(c);
+        }
+        else {
+            Set<E> transform = new HashSet<>(c);
+            Set<E> difference = SetUtils.difference(delegate, transform).toSet();
+            if (difference.size() > getMaxSize()) {
+                throw new IllegalArgumentException(String.format("The given set size %d is larger then maxSize %d",
+                        difference.size(), maxSize));
+            }
+            int itemsToDelete = delegate.size() + difference.size() - getMaxSize();
```
th0br0: this violates the description `A set that doesn't allow to add elements to it once it is full` as you're 
deleting items to make room

src/main/java/com/iota/iri/service/tipselection/impl/WalkValidatorImpl.java
alon-e: BelowMaxDepth is a negative predicate (returning true means its "bad", failing is also bad hence true) - this is annoying to reason about and can lead to bugs. I think we should invert it. Maybe in another PR.

rajivshah3:
When I was testing the [JAR](https://s3.eu-central-1.amazonaws.com/iotaledger-dbfiles/pre-releases/iri-1.5.2-below-maxdepth-cached-1.jar) posted by @alon-e in Discord, I think I encountereda a bug that causes an error if you try to promote a transaction. It will try to do the tip selection but then it will say “<tx hash you’re trying to promote> is below max depth” (if you have debug logging enabled) and then “entry point failed consistency check: <tx hash you’re trying to promote>”. I was only able to reproduce this with that JAR, the issue does not show up when using 1.5.1.

src/main/java/com/iota/iri/service/tipselection/WalkValidator.java
GalRogozinski: Yup,
I wanted to get the first iteration of the PR so I only did the necessary changes. I agree with you ofc.

alon-e:
@rajivshah3 could you give an example tx hash?

Was the original transaction sent with the new jar as well?

src/main/java/com/iota/iri/service/tipselection/impl/WalkValidatorImpl.java
GalRogozinski: Something got messed up with the commits I will fix it.

src/main/java/com/iota/iri/service/tipselection/impl/WalkValidatorImpl.java
GalRogozinski: lower

src/main/java/com/iota/iri/utils/collections/impl/BoundedSetWrapper.java
GalRogozinski: Yup, copy paste mistake.
Will change description.

src/main/java/com/iota/iri/utils/collections/impl/BoundedSetWrapper.java
GalRogozinski: ditto

src/main/java/com/iota/iri/service/tipselection/impl/WalkValidatorImpl.java
GalRogozinski: I agree with you actually

rajivshah3:
@alon-e it’s FYQSGNDHJNGHCTAHEYYCAQDLGAZWVTGQYVNMUNFMGYOQCKFZGSZJHUITQAZCSEFQVPFQXCQCOMII99999. I originally sent it with 1.5.1 but it has been reattached with the test jar as well

src/main/java/com/iota/iri/utils/collections/impl/BoundedSetWrapper.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    @Override
+    public boolean addAll(Collection<? extends E> c) {
+        requireCollectionIsNotAboveMaxSize(c, getMaxSize());
+        if (canCollectionBeFullyAdded(c)) {
+            return delegate.addAll(c);
+        }
+        else {
+            int itemsToDelete = delegate.size() + c.size() - getMaxSize();
```
alon-e: so I don't forget - please open an issue (not blocking the PR):
there may be duplicate entries in the sets `delegate` and `c` - in that case you may want to remove less items.
for example:
```
delegate = {a, b, c, d}
c = {c, d, e}
maxSize = 5
itemsToDelete = 4 +3 - 5 => 2
result: {c, d, e}
```

you could argue that it's too costly to check, but then there should be a comment explaining that heuristic.

src/main/java/com/iota/iri/utils/collections/impl/BoundedSetWrapper.java
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return delegate.toString();
+    }
+
+    private void requireCollectionIsNotAboveMaxSize(Collection<? extends E> c, int maxSize) {
+        if (c.size() > maxSize) {
+            throw new IllegalArgumentException(String.format("The given collection size %d is larger then maxSize %d", c.size(),
```
alon-e: typo: "larger then" -> "larger than"

src/test/java/com/iota/iri/service/tipselection/impl/WalkValidatorImplTest.java
```diff
@@ -38,6 +40,7 @@
     public static void tearDown() throws Exception {
         tangle.shutdown();
         dbFolder.delete();
+        WalkValidatorImpl.failedBelowMaxDepthCache.clear();
```
alon-e: This will run only once per class invocation, as `tearDown` is `@AfterClass`.
is that intentional? I don't see a reason to clear the cache after disposing of the DB.

If you want to run after each test:
```
@After
    public void tearDownEach() {
        WalkValidatorImpl.failedBelowMaxDepthCache.clear();
    }
```
* tests still pass if you add this.

src/main/java/com/iota/iri/IRI.java
```diff
@@ -330,7 +331,20 @@ private static boolean isValidated(final Configuration configuration, final Stri
...
                 configuration.put(DefaultConfSettings.MAX_PEERS, vmaxPeers);
             }
+
+            final Integer belowMaxDepthLimit = parser.getOptionValue(belowMaxDepthTxLimit);
+            if (belowMaxDepthLimit != null) {
+                configuration.put(DefaultConfSettings.BELOW_MAX_DEPTH_TRANSACTION_LIMIT,
+                        String.valueOf(belowMaxDepthLimit));
+            }
+
+            final Integer walkValidatorCache = parser.getOptionValue(walkValidatorCacheSize);
```
hmoog: maybe rename to "walkValidatorCacheSize" to be consistent with the naming in the config and Iota.java - could otherwise be mistaken with an instance of the cache (not really :) but you get the point)

GalRogozinski: The problem is that we have 2 variables with the same name...
It will be resolved when we do #724 .

src/test/java/com/iota/iri/service/tipselection/impl/WalkValidatorImplTest.java
GalRogozinski: yes

src/main/java/com/iota/iri/utils/collections/impl/BoundedSetWrapper.java
GalRogozinski: Opened https://github.com/iotaledger/iri/issues/887

Issue #899, opened 2018-07-30
Figure out why the checkConsistency calls fail


jakubcech:
### Description
The JS library obfuscates the exceptions coming from the `checkConsistency` call, let's de-obfuscate and reproduce the issues. We need to figure out in which section we fail and whether this is a regression or not - we've made two changes to `checkConsistency` recently:
- stop if the transaction is not solid
- addition of `belowMaxDepth`

Note: Chris is working on the client call de-obfuscation.

chrisdukakis:
Hey @jakubcech  & @GalRogozinski ,
I've created two pull requests for `develop (iota.lib.js)` & `next (@iota/core)` branches:

https://github.com/iotaledger/iota.lib.js/pull/255
https://github.com/iotaledger/iota.lib.js/pull/251

To print exceptions, `checkConsistency` should be called with `rejectWithReason: true` option:
```
npm i @iota/core @iota/http-client
```
```js
const { createCheckConsistency } = require('@iota/core')
const { createHttpClient } = require('@iota/http-client')

const provider = createHttpClient({ provider: 'http://..' }) // <- change to your node uri
const checkConsistency = createCheckConsistency(provider)

const tail = 'TAIL9HASH...'

checkConsistency(tail, { rejectWithReason: true })
   .then(state => console.log(`State is: ${state}`))
   .catch(err => console.log(`Check failed, reason: ${err}`))
```

cyclux:
In my latest test the API fails with:
Error: Inconsistent subtangle

The problem is occuring 100% of the time now.

alon-e:
>In my latest test the API fails with:
>"Error: Inconsistent subtangle"
@cyclux This eherror obfuscates the real reason why IRI fails on the transaction - any error would trigger the same response.
in Chris' version, the error message is revealed.
Can you run w/ that version? or call IRI directly?
(you can use:)
```
curl -H 'X-IOTA-API-Version: 1' -H 'Content-Type: application/json' -d '{"command":"checkConsistency", "tails":["NDMGKNFYSRFNCGLXWEDBRLDJLRUJEIYGKIHLIDKCHDSMENOORWVEFGRCRGWAEEPCAHHASLVORTHM99999"]}' $URL1 | jq
```

gjeee:
@alon-e my node shows the following. so this confirms indeed the discrepancy you demonstrated in issue #898

[...]

Issue #898, opened 2018-07-30
Are nodes stopping the random walk based on partial solidity checks?


jakubcech:
### Description
Nodes may be stopping random walks based on partial solidity checks. We need to patch the Tip Selection algorithm to report when a transaction stops due to a solidity check. Then we query other nodes for solidity state using `checkConsistency`. 

The assumption is that there is a discrepancy between nodes on the consistency/solidity.

### Open questions
Do we need to check whether this is actually happening on the MainNet or can we use Spamnet?

alon-e:
Example of such discrepancy:
```
curl -H 'X-IOTA-API-Version: 1' -H 'Content-Type: application/json' -d '{"command":"checkConsistency", "tails":["NDMGKNFYSRFNCGLXWEDBRLDJLRUJEIYGKIHLIDKCHDSMENOORWVEFGRCRGWAEEPCAHHASLVORTHM99999"]}' $URL1 | jq
curl -H 'X-IOTA-API-Version: 1' -H 'Content-Type: application/json' -d '{"command":"checkConsistency", "tails":["NDMGKNFYSRFNCGLXWEDBRLDJLRUJEIYGKIHLIDKCHDSMENOORWVEFGRCRGWAEEPCAHHASLVORTHM99999"]}' $URL2 | jq
curl -H 'X-IOTA-API-Version: 1' -H 'Content-Type: application/json' -d '{"command":"checkConsistency", "tails":["NDMGKNFYSRFNCGLXWEDBRLDJLRUJEIYGKIHLIDKCHDSMENOORWVEFGRCRGWAEEPCAHHASLVORTHM99999"]}' $URL3 | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   291  100   166  100   125    151    113  0:00:01  0:00:01 --:--:--   151
{
  "state": false,
  "info": "tails are not solid (missing a referenced tx): NDMGKNFYSRFNCGLXWEDBRLDJLRUJEIYGKIHLIDKCHDSMENOORWVEFGRCRGWAEEPCAHHASLVORTHM99999",
  "duration": 1
}
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   249  100   124  100   125     19     19  0:00:06  0:00:06 --:--:--    32
{
  "state": false,
[...]

PR #907, opened 2018-08-05
Fix solidity propagation from bottom to top

GalRogozinski:
# Description

There is a `propagateSolidTransaction` thread that is supposed to cascade solidity from bottom to top.
If a transaction is solid and its approver's other parent is solid then make the approver solid. Then keep on propagating upwards in this manner.
Unfortunately, this didn't happen. The solidified approver didn't reenter the queue so the change was not cascaded.
 
May help with #899 / #898 

## Type of change

# How Has This Been Tested?

The following unit tests have been added 
- `testTransactionPropagation`
- `testTransactionPropagationFailure`

In addition, canaries were run by @DyrellC and @geminoz.

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/TransactionValidator.java
+++ b/src/main/java/com/iota/iri/TransactionValidator.java
@@ -172,6 +172,11 @@ public class TransactionValidator {
             while(!shuttingDown.get()) {
                 propagateSolidTransactions();
             }
+            try {
+                Thread.sleep(500);
+            } catch (InterruptedException e) {
+                e.printStackTrace();
+            }
         };
     }
 
@@ -204,11 +209,6 @@ public class TransactionValidator {
             } catch (Exception e) {
                 log.error("Error while propagating solidity upwards", e);
             }
-        }
-        try {
-            Thread.sleep(500);
-        } catch (InterruptedException e) {
-            e.printStackTrace();
         }
     }
 
--- a/src/main/java/com/iota/iri/TransactionValidator.java
+++ b/src/main/java/com/iota/iri/TransactionValidator.java
@@ -1,15 +1,14 @@
 package com.iota.iri;
 
 import com.iota.iri.controllers.TipsViewModel;
+import com.iota.iri.controllers.TransactionViewModel;
 import com.iota.iri.hash.Curl;
 import com.iota.iri.hash.Sponge;
 import com.iota.iri.hash.SpongeFactory;
-import com.iota.iri.network.TransactionRequester;
-import com.iota.iri.controllers.TransactionViewModel;
 import com.iota.iri.model.Hash;
-import com.iota.iri.zmq.MessageQ;
+import com.iota.iri.network.TransactionRequester;
 import com.iota.iri.storage.Tangle;
-
+import com.iota.iri.zmq.MessageQ;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -53,15 +52,19 @@ public class TransactionValidator {
     }
 
     public void init(boolean testnet, int mwm) {
+        setMwm(testnet, mwm);
+
+        newSolidThread = new Thread(spawnSolidTransactionsPropagation(), "Solid TX cascader");
+        newSolidThread.start();
+    }
+
+    void setMwm(boolean testnet, int mwm) {
         MIN_WEIGHT_MAGNITUDE = mwm;
 
         //lowest allowed MWM encoded in 46 bytes.
         if (!testnet && MIN_WEIGHT_MAGNITUDE<13){
             MIN_WEIGHT_MAGNITUDE = 13;
         }
-
-        newSolidThread = new Thread(spawnSolidTransactionsPropagation(), "Solid TX cascader");
-        newSolidThread.start();
     }
 
     public void shutdown() throws InterruptedException {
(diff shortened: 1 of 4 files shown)


src/main/java/com/iota/iri/controllers/TransactionViewModel.java
```diff
@@ -1,19 +1,20 @@
...
 import com.iota.iri.utils.Pair;
 
+import java.util.*;
+
 public class TransactionViewModel {
 
     private final com.iota.iri.model.Transaction transaction;
 
     public static final int SIZE = 1604;
+    public static final int TRYTES_SIZE = 2673;
```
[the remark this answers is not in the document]
GalRogozinski: done

src/main/java/com/iota/iri/TransactionValidator.java
```diff
@@ -53,15 +52,19 @@ public TransactionValidator(Tangle tangle, TipsViewModel tipsViewModel, Transact
     }
 
     public void init(boolean testnet, int mwm) {
+        setMwm(testnet, mwm);
+
+        newSolidThread = new Thread(spawnSolidTransactionsPropagation(), "Solid TX cascader");
+        newSolidThread.start();
+    }
+
+    void setMwm(boolean testnet, int mwm) {
```
th0br0: just a nit, but `if(!testnet) MIN_WEIGHT_MAGNITUDE = Math.max(MIN_WEIGHT_MAGNITUDE, 13)` would be less magic-number heavy

src/main/java/com/iota/iri/TransactionValidator.java
```diff
@@ -167,51 +170,50 @@ public void addSolidTransaction(Hash hash) {
...
-                        newSolidTransactionsOne.clear();
+                propagateSolidTransactions();
+            }
+        };
+    }
+
+    void propagateSolidTransactions() {
+        Set<Hash> newSolidHashes = new HashSet<>();
+        useFirst.set(!useFirst.get());
+        synchronized (cascadeSync) {
```
th0br0: Although you're just changing usage; this is a good time to add a comment to the lock object to explain how it's used / what it synchronises. (newSolidTransactions One/Two)

src/main/java/com/iota/iri/TransactionValidator.java
```diff
@@ -167,51 +170,50 @@ public void addSolidTransaction(Hash hash) {
...
-                    } else {
-                        newSolidTransactionsOne.clear();
+                propagateSolidTransactions();
+            }
+        };
+    }
+
+    void propagateSolidTransactions() {
+        Set<Hash> newSolidHashes = new HashSet<>();
+        useFirst.set(!useFirst.get());
```
th0br0: `useFirst` what?

src/main/java/com/iota/iri/TransactionValidator.java
```diff
@@ -167,51 +170,50 @@ public void addSolidTransaction(Hash hash) {
...
+        }
+        synchronized (cascadeSync) {
+            if (useFirst.get()) {
+                newSolidTransactionsTwo.clear();
+            } else {
+                newSolidTransactionsOne.clear();
+            }
+        }
+        try {
+            Thread.sleep(500);
```
th0br0: move the sleep to the runnable? this method shouldn't have a sleepy side effect

src/main/java/com/iota/iri/TransactionValidator.java
```diff
@@ -167,51 +170,50 @@ public void addSolidTransaction(Hash hash) {
...
-                    e.printStackTrace();
-                }
+            } catch (Exception e) {
+                log.error("Error while propagating solidity upwards", e);
             }
-        };
+        }
+        synchronized (cascadeSync) {
+            if (useFirst.get()) {
+                newSolidTransactionsTwo.clear();
```
th0br0: Why do you not clear these directly after `addAll`?
This means that you'll potentially discard transactions that were added after your `addAll` above

src/main/java/com/iota/iri/TransactionValidator.java
```diff
@@ -167,51 +170,50 @@ public void addSolidTransaction(Hash hash) {
...
-                    if (useFirst.get()) {
-                        newSolidTransactionsTwo.clear();
-                    } else {
-                        newSolidTransactionsOne.clear();
+                propagateSolidTransactions();
+            }
+        };
+    }
+
+    void propagateSolidTransactions() {
```
th0br0: method should have explicit visibility

src/main/java/com/iota/iri/TransactionValidator.java
```diff
@@ -266,6 +268,11 @@ private boolean checkApproovee(TransactionViewModel approovee) throws Exception
         return approovee.isSolid();
     }
 
+    //for testing
+    boolean isNewSolidTxSetsEmpty () {
```
th0br0: method should have explicit visibility? (or did we consciously make the choice to allow this?)

src/test/java/com/iota/iri/TransactionTestUtils.java
```diff
@@ -34,4 +36,24 @@ public static TransactionViewModel createTransactionWithTrunkBundleHash(Transact
                 TransactionViewModel.BUNDLE_TRINARY_OFFSET, TransactionViewModel.BUNDLE_TRINARY_SIZE);
         return tx;
     }
+
+    public static TransactionViewModel createTransactionWithTrytes(String trytes) {
+        String expandedTrytes  = expandTrytes(trytes);
+        byte[] trits = Converter.allocatingTritsFromTrytes(expandedTrytes);
+        return new TransactionViewModel(trits, Hash.calculate(SpongeFactory.Mode.CURLP81, trits));
```
th0br0: shouldn't we just add a default constructor to TxVM or a helper method somewhere?

src/main/java/com/iota/iri/TransactionValidator.java
GalRogozinski: I suppose while this code exists this is possible...

src/main/java/com/iota/iri/TransactionValidator.java
GalRogozinski: This is part of the old code. I just wanted to solve the current problem without changing too much of it at the moment. 

I think much of it should be changed bot not atm.

GalRogozinski: Btw,
This has default visibility for the test code.

src/main/java/com/iota/iri/TransactionValidator.java
GalRogozinski: This was part of the old code. I only wanted to fix the bug.
I wanted to change as little as possible from the old code and see the effect.

I will open a general issue on solidification though. Because indeed the way things are written should be changed.

src/main/java/com/iota/iri/TransactionValidator.java
GalRogozinski: I didn't do it because I wanted to change as little as possible from the code.
Since you claim this is also a bug (and I think you are correct) I will fix it now.

src/main/java/com/iota/iri/TransactionValidator.java
GalRogozinski: The reason it doesn't have explicit visibility is because it shouldn't be `public` yet unfortunately it is used in tests so unfortunately `private` won't work...
So default `package` visibility is the best option.

This a common thing we did in my last workplace. You have a better idea?

src/main/java/com/iota/iri/TransactionValidator.java
GalRogozinski: ditto

src/test/java/com/iota/iri/TransactionTestUtils.java
GalRogozinski: It has this `expandTrytes` call.
I don't want this as part of default behavior...

src/main/java/com/iota/iri/TransactionValidator.java
GalRogozinski: I pulled the code to `setMwm` to make the test work,
The implementation logic is old and not really part of this PR.

But since you commented I will change :-)

gjeee:
well done!!!! i think this will have a major impact on how the tangle behaves (especially the random walks)

src/main/java/com/iota/iri/TransactionValidator.java
th0br0: Fair enough, ignore this then.

src/main/java/com/iota/iri/TransactionValidator.java
```diff
@@ -167,51 +170,46 @@ public void addSolidTransaction(Hash hash) {
...
+            if (useFirst.get()) {
+                newSolidHashes.addAll(newSolidTransactionsTwo);
+                newSolidTransactionsTwo.clear();
+            } else {
+                newSolidHashes.addAll(newSolidTransactionsOne);
+                newSolidTransactionsOne.clear();
+            }
+        }
+        Iterator<Hash> cascadeIterator = newSolidHashes.iterator();
+        while(cascadeIterator.hasNext() && !shuttingDown.get()) {
```
th0br0: Does it make sense to check this here by the way? We're checking in the Runnable anyway

src/main/java/com/iota/iri/TransactionValidator.java
```diff
@@ -167,51 +170,46 @@ public void addSolidTransaction(Hash hash) {
...
-                } catch (InterruptedException e) {
-                    e.printStackTrace();
-                }
+            } catch (Exception e) {
+                log.error("Error while propagating solidity upwards", e);
             }
-        };
+        }
+        try {
+            Thread.sleep(500);
```
th0br0: Could we move this to the Runnable instead of the method so you can test this without having to sleep?

src/main/java/com/iota/iri/TransactionValidator.java
GalRogozinski: I think that this will allow for early exit from the loop.
Should it exist? not sure

I just don't want to add more changes for this PR. I want to do sanity tests and then improve it.

gjeee:
the sleep should be within the while loop i think!

alon-e:
@gjeee you are right ofc :100:

src/main/java/com/iota/iri/TransactionValidator.java
```diff
@@ -167,51 +170,46 @@ public void addSolidTransaction(Hash hash) {
...
-                        newSolidTransactionsTwo.clear();
-                    } else {
-                        newSolidTransactionsOne.clear();
+                propagateSolidTransactions();
+            }
+            try {
+                Thread.sleep(500);
+            } catch (InterruptedException e) {
+                e.printStackTrace();
+            }
```
gjeee: this try/catch block should be within the while!!!!

GalRogozinski: Thanks!
Corrected with #911

Issue #761, opened 2018-05-22
TipsViewModel lags reality


gjeee:
### Bug description
When a new tip comes in, it is added by default to the non-solid tip queue (see TransactionValidator.updateStatus()). The TipsManager.scanTipsForSolidity() thread runs every 750 ms. This will pick 1 non-solid tip randomly and checks if it has become solid in the meantime. If this is the case, the tip is put into the solid queue (and removed from the non-solid queue). Now let's say the network runs at 10 tx/s and suppose in this last second these 10 tips all became solid. However the solidity thread in this same timeframe only updates 1 of these. So the TipsViewModel has difficulty keeping up with the current state of things. This has repercussions for the rest of the network as well. Since now and then a neighbor will ask your node for a random tip. Your node will reply then based on its solid tip-view model. But that model misses a lot of recent transactions because they have not been scanned for solidity yet.

### Potential solution

1. Check and update whether a tip is solid immediately after arrival
2. Make the solidity thread do more scans (possibly depending on the size of the non-solid tip queue) and/or more frequent scans

### Am I planning to do it myself with a PR?
Yes / No.

footloosejava:
I am wondering why it runs every 750 ms as well. 

However, let me share this .... the thread does seem very busy when it runs. So I am not sure how many more tips you can scan in that period.

I suspect your CPU usage creep was from Milestone scanning and Tips scanning. As the caches fill, the scanning for both these operations almost becomes full-time.

gjeee:
does it take a lot of resources? that for sure is interesting if you combine it with #756. option 1 seems wise since (i think) 99% of the arriving tx can be "solidity checked" in a quick way....see TransactionValidator.quickSetSolid() routine

footloosejava:
So , do this, change it to check twice or three times, and then sleep 750 ms. See then as an experiment,

gjeee:
That "deepscan" might also use that time-consuming transactionviewmodel call (see #757) which can explain the cost of this call.

footloosejava:
If you can, change it to this, run it for a while and them share how much time is between each number printed on average:

`
    private void scanTipsForSolidity() throws Exception {
        for (int n = 0; n < 10 && tipsViewModel.nonSolidSize() != 0; n++) {
            log.info("checking {}", n);
            Hash hash = tipsViewModel.getRandomNonSolidTipHash();
            if (hash != null) {
                ApproveeViewModel approvers = ApproveeViewModel.load(tangle, hash);
[...]

Issue #573, opened 2018-03-07
Unnecessary gaps in branches


gjeee:
The [checkSolidity routine](https://github.com/iotaledger/iri/blob/2f1fb78b06cbdc7e280b4191326593ef7ea03514/src/main/java/com/iota/iri/TransactionValidator.java#L125) is important since it checks a tx and if any downstream tx are missing it will add it to its transactionRequester. It will also update the solid-property of a tx.

The TipsManager runs a [thread](https://github.com/iotaledger/iri/blob/2e0d1c38b18f196fec727acacff3b3fdf0e06868/src/main/java/com/iota/iri/service/TipsManager.java#L55) every 750 ms to check the solidity regularly for non-solid tips. To this end it picks a random non-solid tip. When it is actually no longer a tip, it will remove the hash from tipsViewModel. That is fine of course. However in this case the checkSolidity() routine is no longer called due to short-circuit evaluator (see code below).

Imo this is unwanted since it can lead to gaps deeper down a branch (and these gaps are not actively worked on by adding it to the transactionRequester). This can also lead to less effective random walks since they are terminated sooner (so less indirect confirms).

```
 private void scanTipsForSolidity() throws Exception {
        int size = tipsViewModel.nonSolidSize();
        if(size != 0) {
            Hash hash = tipsViewModel.getRandomNonSolidTipHash();
            boolean isTip = true;
            if(hash != null && TransactionViewModel.fromHash(tangle, hash).getApprovers(tangle).size() != 0) {
                tipsViewModel.removeTipHash(hash);
                isTip = false;
            }
            if(hash != null  && isTip && transactionValidator.checkSolidity(hash, false)) {
                //if(hash != null && TransactionViewModel.fromHash(hash).isSolid() && isTip) {
                tipsViewModel.setSolid(hash);
            }
        }
}
```

gjeee:
To further speed up the tips-solidity scan I would advise to do a check immediate after receiving a tx.

So I would change the routine [TransactionValidator#updateStatus](https://github.com/iotaledger/iri/blob/2f1fb78b06cbdc7e280b4191326593ef7ea03514/src/main/java/com/iota/iri/TransactionValidator.java#L217)

```
public void updateStatus(TransactionViewModel transactionViewModel) throws Exception {
        transactionRequester.clearTransactionRequest(transactionViewModel.getHash());
        boolean bTip=transactionViewModel.getApprovers(tangle).size() == 0;
        if(bTip) {
[...]

Issue #741, opened 2018-05-03
ZMQ topic prefix matching issue


donsander:
Subscribing to the "tx" topic with ZMQ will also subscribe to the new "tx_trytes" topic, and each new transaction will be received twice.  This is unwanted behavior, and has broken some live tangle visualizers. 

I believe this is due to both topics having a "tx" prefix.  By design ZMQ will perform prefix matching on topics so you can subscribe to multiple topics at once.  This makes it difficult to only subscribe to the "tx" topic alone because it will also match "tx_trytes".

One hacky workaround is to subscribe to "tx " instead of "tx".  Another is to use message filtering option available for ZMQ clients but this is not ideal.  

Perhaps the topic names should be changed to "tx_bytes" and "tx_trytes".  Or to retain some backwards compatibility (which was already lost with the last update) just change the name of the "tx_trytes" topic to something like "tryts".

alon-e:
@donsander thanks for bringing this up.
>has broken some live tangle visualizers.

would be great if you also give more context to what implementation it broke and who are the developers, so we can more extensively test in future - for example, in #530 we asked known ZMQ user to test it: https://github.com/iotaledger/iri/pull/540#issuecomment-378636008

>One hacky workaround is to subscribe to "tx " instead of "tx". Another is to use message filtering option available for ZMQ clients but this is not ideal.

[I agree :)](https://github.com/alon-e/iota-ctps/commit/39bcb8ee784a81deee759ca1a1c6a0339d51d58e
), however, until the issue is resolved this workaround can be used.

I want to remind users that ZMQ [is still WIP,](https://github.com/iotaledger/iri/tree/dev/src/main/java/com/iota/iri/zmq) and will break functionality sometimes, topic names added and changed etc - that's why it is disabled by default. however, we'll try to minimize this and communicate changes more clearly.

donsander:
@alon-e The change caused issues with the 3D tangle viewer I created at live.iotaknot.com - the backend is node.js connecting to an IRI instance. The additional messages from "tx_trytes" were not expected, and for obvious reasons parsing the message would cause issues. If you have any more questions I can answer them.  I have since used the "tx " workaround so everything is working as it should now.

As for ZMQ being WIP, I am aware, and this issue report is just a heads up.  I am thankful that #530 has resolved an issue that I have been dealing with for a long time which required me to restart my IRI instance constantly during high traffic times.  My visualizer is now stable again!

GalRogozinski:
[...]

PR #913, opened 2018-08-06
some TransactionValidator fixes regarding solidity

gjeee:
TipsViewModel is responsible for keeping an accurate picture of the solid and non-solid tips. Currently this view is managed within #TransactionValidator.updateStatus() and a separate TipsSolidifier thread (which is not really efficient since it picks one random non-solid tip at a time each 750ms). This leads to a lag in the tips reality. See also issue #761 and #573. With PR #907 in place, we can manage the TipsViewModel fully within TransactionValidator and keep an accurate and up-to-date picture of the current (solid/non-solid) tips.

### DIFF
--- a/src/main/java/com/iota/iri/service/TipsSolidifier.java
+++ b/src/main/java/com/iota/iri/service/TipsSolidifier.java
@@ -34,7 +34,7 @@ public class TipsSolidifier {
             while (!shuttingDown) {
                 try {
                     scanTipsForSolidity();
-                    if(log.isDebugEnabled()) {
+                    if (log.isDebugEnabled()) {
                         long now = System.currentTimeMillis();
                         if ((now - lastTime) > 10000L) {
                             lastTime = now;
--- a/src/main/java/com/iota/iri/service/TipsSolidifier.java
+++ b/src/main/java/com/iota/iri/service/TipsSolidifier.java
@@ -52,8 +52,7 @@ public class TipsSolidifier {
     }
 
     private void scanTipsForSolidity() throws Exception {
-        int size = tipsViewModel.nonSolidSize();
-        
+        int size = tipsViewModel.nonSolidSize();        
         if (size != 0) {
             Hash hash = tipsViewModel.getRandomNonSolidTipHash();
             boolean isTip = true;
@@ -65,8 +64,7 @@ public class TipsSolidifier {
                 //if(hash != null && TransactionViewModel.fromHash(hash).isSolid() && isTip) {
                 tipsViewModel.setSolid(hash);
             }
-        }
-        
+        }        
     }
 
     public void shutdown() {
--- a/src/main/java/com/iota/iri/service/TipsSolidifier.java
+++ b/src/main/java/com/iota/iri/service/TipsSolidifier.java
@@ -29,6 +29,7 @@ public class TipsSolidifier {
 
     public void init() {
         solidityRescanHandle = new Thread(() -> {
+            
             long lastTime = 0;
             while (!shuttingDown) {
                 try {
--- a/src/main/java/com/iota/iri/service/TipsSolidifier.java
+++ b/src/main/java/com/iota/iri/service/TipsSolidifier.java
@@ -29,7 +29,7 @@ public class TipsSolidifier {
 
     public void init() {
         solidityRescanHandle = new Thread(() -> {
-            
+
             long lastTime = 0;
             while (!shuttingDown) {
                 try {
--- a/src/main/java/com/iota/iri/service/TipsSolidifier.java
+++ b/src/main/java/com/iota/iri/service/TipsSolidifier.java
@@ -48,6 +48,8 @@ public class TipsSolidifier {
 
     private void scanTipsForSolidity() throws Exception {
         int size = tipsViewModel.nonSolidSize();
+        log.info("#Solid/NonSolid: {}/{}",tipsViewModel.size()-size,size);
+        /*
         if (size != 0) {
             Hash hash = tipsViewModel.getRandomNonSolidTipHash();
             boolean isTip = true;
@@ -60,6 +62,7 @@ public class TipsSolidifier {
                 tipsViewModel.setSolid(hash);
             }
         }
+        */
     }
 
     public void shutdown() {
--- a/src/main/java/com/iota/iri/service/TipsSolidifier.java
+++ b/src/main/java/com/iota/iri/service/TipsSolidifier.java
@@ -52,7 +52,7 @@ public class TipsSolidifier {
     }
 
     private void scanTipsForSolidity() throws Exception {
-        int size = tipsViewModel.nonSolidSize();        
+        int size = tipsViewModel.nonSolidSize();
         if (size != 0) {
             Hash hash = tipsViewModel.getRandomNonSolidTipHash();
             boolean isTip = true;
@@ -64,7 +64,7 @@ public class TipsSolidifier {
                 //if(hash != null && TransactionViewModel.fromHash(hash).isSolid() && isTip) {
                 tipsViewModel.setSolid(hash);
             }
-        }        
+        }
     }
 
     public void shutdown() {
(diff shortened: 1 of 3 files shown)


gjeee:
the last commit fails the test because i moved the persistence before the approver loop (which is more elegant imo)

gjeee:
I also have a version which moves all the heavy db work (getting approvers & persisting the solid property) to the "Solid TX cascader"-thread. This way updateStatus() which is called directly after rcv a new tx is as lean as possible. I will show the relevant code fragments here. I will not commit yet since it does not pass the test (however that is imo not a show-stopper). This version makes the TipsSolidifier thread redundant. Further tipsViewModel now only collects references to tips once they become solid. Keeping tracking of non-solid ones is not of any practical use afaik.

```
private void updateTipsView(TransactionViewModel transactionViewModel) throws Exception {
	
	if(transactionViewModel.getApprovers(tangle).size() == 0){
		tipsViewModel.addTipHash(transactionViewModel.getHash());
                if(transactionViewModel.isSolid()){
			tipsViewModel.setSolid(transactionViewModel.getHash());
		}
         }
         else{
	   tipsViewModel.removeTipHash(transactionViewModel.getHash());
	 }
    
	tipsViewModel.removeTipHash(transactionViewModel.getTrunkTransactionHash());
        tipsViewModel.removeTipHash(transactionViewModel.getBranchTransactionHash());
}
	
	
public void updateStatus(TransactionViewModel transactionViewModel) throws Exception {
	transactionRequester.clearTransactionRequest(transactionViewModel.getHash());
      
	if(quickSetSolid(transactionViewModel)) {
		addSolidTransaction(transactionViewModel.getHash());
        }
       
}
	
	
private boolean quickSetSolid(final TransactionViewModel transactionViewModel) throws Exception {
        
	if(!transactionViewModel.isSolid()) {
		boolean solid=true;
                if (!checkApproovee(transactionViewModel.getTrunkTransaction(tangle))) {
			solid = false;
                }
                if (!checkApproovee(transactionViewModel.getBranchTransaction(tangle))) {
                        solid = false;
                }
                if(solid) {
                      return true;
	        }
        }
	return false;
}

void propagateSolidTransactions() {
	Set<Hash> newSolidHashes = new HashSet<>();
       
	useFirst.set(!useFirst.get());
	//synchronized to make sure no one is changing the newSolidTransactions collections during addAll
	synchronized (cascadeSync) {
		if (useFirst.get()) {
			newSolidHashes.addAll(newSolidTransactionsTwo);
			newSolidTransactionsTwo.clear();
		} else {
			newSolidHashes.addAll(newSolidTransactionsOne);
			newSolidTransactionsOne.clear();
		}
	}
	
	Iterator<Hash> cascadeIterator = newSolidHashes.iterator();
	while(cascadeIterator.hasNext()) {
		try {
			Hash hash = cascadeIterator.next();
			TransactionViewModel transaction = TransactionViewModel.fromHash(tangle, hash);
			Set<Hash> approvers = transaction.getApprovers(tangle).getHashes();
			
			transaction.updateHeights(tangle);
			transaction.updateSolid(true);
			transaction.update(tangle,"solid|height");
			updateTipsView(transaction);
			for(Hash h: approvers) {
				TransactionViewModel tx = TransactionViewModel.fromHash(tangle, h);
				if(quietQuickSetSolid(tx)) {			
					addSolidTransaction(h);			
				}            
			}
		} catch (Exception e) {
			log.error("Error while propagating solidity upwards", e);
		}
	}
      
}
```

gjeee:
my latest 2 commits have minimal (but important) changes: 
1. it ensures that the solid property is stored in #TransactionValidator.updateStatus()
2. if quickSetSolid is true it calls tipsViewModel setSolid(hash). The tips-state is now timely and correctly updated.

GalRogozinski:
@karimodm 
Could you please try to run your promotion tests with this PR?

src/main/java/com/iota/iri/service/TipsSolidifier.java
```diff
@@ -48,6 +48,8 @@ public void init() {
 
     private void scanTipsForSolidity() throws Exception {
         int size = tipsViewModel.nonSolidSize();
+        log.info("#Solid/NonSolid: {}/{}",tipsViewModel.size()-size,size);
```
GalRogozinski: I think that this should be a debug message.
Also  add `getNumOfSolidTips()` in `TipsViewModel`

alon-e: if left on purpose and made debug (agree w/ Gal), still this should be printed very infrequently imo.
iiuc, in the current PR this will be printed more than every second.

gjeee: shall i add  a function called .solidSize() instead? would be more consistent naming-wise with nonSolidSize() :-)

gjeee: @alon-e rgd the debug info shall i make a construct like https://github.com/iotaledger/iri/blob/a45fa1a63063aec2bd5944875eb9e0f38f0ef8a5/src/main/java/com/iota/iri/network/Node.java#L537

src/main/java/com/iota/iri/service/TipsSolidifier.java
```diff
@@ -30,9 +30,15 @@ public TipsSolidifier(final Tangle tangle,
...
         solidityRescanHandle = new Thread(() -> {
 
+            long lastTime = 0;
             while (!shuttingDown) {
                 try {
                     scanTipsForSolidity();
+                    long now = System.currentTimeMillis();
+                    if ((now - lastTime) > 10000L) {
+                        lastTime = now;
+                        log.debug("#Solid/NonSolid: {}/{}",tipsViewModel.solidSize(),tipsViewModel.nonSolidSize());
```
GalRogozinski: Just add a space after `solidSize(),`

gjeee:
@GalRogozinski  done!

src/main/java/com/iota/iri/service/TipsSolidifier.java
```diff
@@ -30,9 +30,15 @@ public TipsSolidifier(final Tangle tangle,
     public void init() {
         solidityRescanHandle = new Thread(() -> {
 
+            long lastTime = 0;
             while (!shuttingDown) {
                 try {
                     scanTipsForSolidity();
+                    long now = System.currentTimeMillis();
```
GalRogozinski: Since you will only run this logic in debug mode and you don't want to do a system call every time, add `log.isDebugEnabled()`.

src/main/java/com/iota/iri/service/TipsSolidifier.java
```diff
@@ -30,9 +30,15 @@ public TipsSolidifier(final Tangle tangle,
...
+                    if ((now - lastTime) > 10000L) {
+                        lastTime = now;
+                        log.debug("#Solid/NonSolid: {}/{}", tipsViewModel.solidSize(), tipsViewModel.nonSolidSize());
```
GalRogozinski: Can you also write this to ZMQ? Use the following topic "debug-TipsSolidifier" we want to change our ZMQ topics in the future anyhows.

karimodm:
After some tests comparing this changes with the latest dev I can say that this PR dramatically improves the solid / nonSolid tips ratio. Good for me @gjeee !

src/main/java/com/iota/iri/service/TipsSolidifier.java
GalRogozinski: @gjeee after thinking about this farther this change shouldn't be part of this PR.
I wanted to use the ZMQ for testing but it should be done as part of this issue:
https://github.com/iotaledger/iri/issues/741.

Maybe in order to create a test we will add a ZMQ message separate from #741 but let's not do it as part of this PR.

So you have only 1 change.

gjeee: roger.

src/main/java/com/iota/iri/service/TipsSolidifier.java
gjeee: and we're done :-)

gjeee:
@karimodm  the number of non-solid tips can still grow in a malicious environment (see issue #938) but at least this PR tracks and updates the solidity like it is.

src/main/java/com/iota/iri/service/TipsSolidifier.java
gjeee: @GalRogozinski i am not sure why the build [failed](https://travis-ci.org/iotaledger/iri/jobs/415925027)

PR #914, opened 2018-08-08
Fix: TransactionValidator instance based / TransactionRequester fixed

hmoog:


src/main/java/com/iota/iri/TransactionValidator.java
```diff
@@ -76,7 +76,7 @@ public int getMinWeightMagnitude() {
         return MIN_WEIGHT_MAGNITUDE;
     }
 
-    private static boolean hasInvalidTimestamp(TransactionViewModel transactionViewModel) {
+    private boolean hasInvalidTimestamp(TransactionViewModel transactionViewModel) {
```
GalRogozinski: I agree with removing `static` on all `public` methods because you don't want to have a global scope `TransactionValidator` logic. However on *private* methods I do prefer to keep the `static` modifier.
  
This is because of 2 reasons:
1. I can quickly know that the method doesn't change the instance state by just looking at the signature.
2. There may be better optimizations that the JIT will do now that the method is marked as static. Today modern JVM are heavily optimized and I am not sure how important this consideration is nowadays. Still the results will be JVM dependent.

src/main/java/com/iota/iri/network/TransactionRequester.java
```diff
@@ -84,48 +82,49 @@ private boolean transactionsToRequestIsFull() {
...
-                    }
-                    break;
+
+                    // ... and continue to the next element in the set
+                    continue;
                 }
+
+                // ... otherwise -> re-add it at the end of the set ...
+                //
+                // Note: we always have enough space since we removed the element before
```
GalRogozinski: Are you positive about that?
In `requestTransaction` method we add without checking if we are full (which may be another bug). But if I call on this method enough times I can exceed capacity and then I won't have enough space even though I just removed a hash.

Anyhows, now you introduced a logic change that should be tested...
The reason it takes time to refactor IRI is becuase we want to add proper tests to the changes and not just replace bugs with other bugs.

I personally want to refactor every single class here but I want to do it correctly.

hmoog: I am positive about that!

In line 71 (requestTransaction) we DO check if !transactionsToRequestIsFull() if it is a normal transaction. the request set for milestoneTransactions is not limited in size, which makes sense since all of the milestones are relevant for our node.

src/main/java/com/iota/iri/TransactionValidator.java
hmoog: Atm we use this method to filter old transactions that are still being gossiped about in the network. Since the cut off time is the timestamp of the global snapshot that is not really a big issue in the current environment since it is quite unlikely that somebody has a wrong time set in his node that dates back before the global snapshot.

When we introduce local snapshots and consequently local cutoff times it can however happen (and it does), that we face a transaction with an invalid timestamp that dates back before our cutoff time, even tho it was issued after.

To still be able to consider this transaction valid if we need it for solidification we have to check if the transaction was part of the TransactionRequester. And to be able to access the transactionRequester instance - this method has to be non static.

Regarding 2: I don't think that these kind of micro-optimizations play any role at all - especially if they help us solve a problem that otherwise would break the synced state of a node.

see: https://github.com/iotadevelopment/iri/commit/40df07b8fbf8a7fc4020e4ffc4d700b07b081f97

GalRogozinski: :+1:
If you need access to the object than there is no choice. 
I still like making private methods static if they don't need the instance.

src/main/java/com/iota/iri/network/TransactionRequester.java
GalRogozinski: :+1:

src/main/java/com/iota/iri/TransactionValidator.java
GalRogozinski: Even though I am fine with this change, note that the fact that the `TransactionValidator` has a `TransactionRequester` instance is a violation of the Single Responsibility Principle, and in my opinion it shouldn't be there. `TransactionValidator` should ultimately become a stateless Service.

Of course that it is not your fault that IRI is this way and definitely you should be more concerned about getting local snapshots done rather than fixing IRI architecture.
I just wanted to merge now and I looked at the code again and had to share this thought.

src/main/java/com/iota/iri/network/TransactionRequester.java
```diff
@@ -84,48 +82,49 @@ private boolean transactionsToRequestIsFull() {
...
-                 requestSet = transactionsToRequest;
-             }
-        } else {
-            requestSet = transactionsToRequest;
-            if(requestSet.size() == 0) {
-                requestSet = milestoneTransactionsToRequest;
-            }
-        }
         synchronized (syncObj) {
+            // repeat while we have transactions that shall be requested
```
GalRogozinski: For next time instead of adding comments you can just refactor the code to be more readable.
For example you could have wrote `!requestSize.isEmpty()` or if you want a null set to be considered as an empty set (in this specific example it won't ever be null) `CollectionUtils.isNotEmpty(requestSet)`

src/main/java/com/iota/iri/network/TransactionRequester.java
```diff
@@ -84,48 +82,49 @@ private boolean transactionsToRequestIsFull() {
...
             while (requestSet.size() != 0) {
+                // remove the first item in our set for further examination
                 Iterator<Hash> iterator = requestSet.iterator();
                 hash = iterator.next();
                 iterator.remove();
+
+                // if we have received the transaction in the mean time ....
                 if (TransactionViewModel.exists(tangle, hash)) {
+                    // ... dump a log message ...
```
GalRogozinski: This comment is completely extraneous.

PR #933, opened 2018-08-13
Skipping milestones

hmoog:
This PR fixes the milestone ordering problem which leads to an inconsistent ledger state and "Skipping negative value for ... " messages in iri

It does so by:

1. reading the milestoneStartIndex and the latestMilestoneIndex directly from the snapshot rather than some copied values (against race coditions and gets rid of some obsolete variables - also necessary for local snapshots later since those values will then be dependant of the local snapshots taken)
2. changing the MilestoneViewModel.findClosestNextMilestone method to retrieve the milestones by following index rather than db insertion order
3. reacting to appearance of mis-ordered milestones by resetting the ledger to a state before the erroneous event. The mis-ordering is determined by:
    3.1 checking the snapshotIndex of milestones when they get processed
    3.2 checking if new milestones get found while the latestSolidMilestoneIndex is higher already


hmoog:
ready for review

src/test/java/com/iota/iri/service/tipselection/impl/EntryPointSelectorImplTest.java
```diff
@@ -1,63 +1,82 @@
...
+@FixMethodOrder(MethodSorters.NAME_ASCENDING)
 public class EntryPointSelectorImplTest {
+    private static final TemporaryFolder dbFolder = new TemporaryFolder();
+    private static final TemporaryFolder logFolder = new TemporaryFolder();
+    private static Tangle tangle;
 
     @Mock
     private Milestone milestone;
-    @Mock
-    private Tangle tangle;
```
GalRogozinski: The reason tangle was mocked was to increase isolation.
We want `EntryPointSelectorImplTest` to fail because of a bug in `EntryPointSelector`.
We have enough unit tests that use `tangle` as it is that slow the tests down.
Can you return the mock?

src/main/java/com/iota/iri/Milestone.java
```diff
@@ -35,6 +36,11 @@
         INCOMPLETE
     }
 
+    /**
+     * This variable is used to keep track of the asynchronous tasks, that the "Solid Milestone Tracker" should wait for.
+     */
+    private AtomicInteger solidMilestoneTrackerTasks;
```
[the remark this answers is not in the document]
hmoog: updateLatestSolidSubtangleMilestone() checks if there are background jobs running and skips the logic if we are doing a hard or soft reset or didnt finish scanning the milestones, yet - so it doesnt stop the thread but it skips the updates

You could achieve the same with a lock but "[...] at high contention levels locking tends to outperform atomic variables, but at more realistic contention levels atomic variables outperform locks."

src/main/java/com/iota/iri/Milestone.java
```diff
@@ -92,16 +102,28 @@ public void init(final SpongeFactory.Mode mode, final LedgerValidator ledgerVali
...
+                long lastLogTime = System.currentTimeMillis();
 
                 try {
                     final int previousLatestMilestoneIndex = latestMilestoneIndex;
                     Set<Hash> hashes = AddressViewModel.load(tangle, coordinator).getHashes();
                     { // Update Milestone
                         { // find new milestones
                             for(Hash hash: hashes) {
+                                // show the scanning progress, since the first scan can potentially take a lot of time
+                                if(firstRun && (System.currentTimeMillis() - lastLogTime) >= 5000) {
```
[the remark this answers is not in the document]
hmoog: agree - will change

src/main/java/com/iota/iri/Milestone.java
```diff
@@ -115,6 +137,12 @@ public void init(final SpongeFactory.Mode mode, final LedgerValidator ledgerVali
                                                 }
                                                 break;
                                             case INCOMPLETE:
+                                                // issue a solidity check to solidify incomplete milestones
+                                                // Note: otherwise a milestone that was followed by a coo-snapshot might
+                                                //       never get solidifed again since it doesnt have connections to
+                                                //       the tips
+                                                transactionValidator.checkSolidity(t.getHash(), true);
+
```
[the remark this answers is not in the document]
hmoog: IRI doesn't persist the tips between iri restarts. So if you restart IRI or happen to download a database with an incomplete milestone it will never be "fixed".

src/main/java/com/iota/iri/Milestone.java
```diff
@@ -179,6 +207,121 @@ public void init(final SpongeFactory.Mode mode, final LedgerValidator ledgerVali
...
 
+    /**
+     * This method allows us to soft reset the ledger state, in case we face an inconsistent SnapshotState.
+     *
+     * It simply resets the latest snapshot to the initial one and rebuilds the ledger state. This will also make the
+     * updateLatestSolidSubtangleMilestone trigger again and give it a chance to detect corruptions.
+     */
+    public void softReset() {
+        // increase a counter for the background tasks to pause the "Solid Milestone Tracker"
+        solidMilestoneTrackerTasks.incrementAndGet();
```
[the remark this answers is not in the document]
hmoog: we pause the updates since the while loop checks for running jobs

src/main/java/com/iota/iri/Snapshot.java
```diff
@@ -108,7 +114,7 @@ public int index() {
         return i;
     }
 
-    private Snapshot(Map<Hash, Long> initialState, int index) {
+    public Snapshot(Map<Hash, Long> initialState, int index) {
```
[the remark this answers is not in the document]
hmoog: since we use the index of the snapshots as the real "latestSolidMilestoneIndex" rather than a copied value to avoid race conditions and to prepare for local snapshots, we need to create Snapshots in some tests manually (see EntryPointSelectorImplTest.java) for mocking

src/main/java/com/iota/iri/Milestone.java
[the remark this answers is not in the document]
GalRogozinski: I see you added it to the `unsolidMilestones` set which is way better.
I guess this change is necessary until we sort out solidifications...
@karimodm note my previous comment

src/main/java/com/iota/iri/LedgerValidator.java
```diff
@@ -220,6 +220,15 @@ private MilestoneViewModel buildSnapshot() throws Exception {
 
                     log.info(logMessage.toString());
                 }
+
+                // if we face a milestone that wasn't processed by updateSnapshot, correctly -> abort and let the
+                // "Solid Milestone Tracker" do it's magic
+                //
+                // NOTE: this can happen if a new subtangle becomes solid before a previous one while syncing
+                if(TransactionViewModel.fromHash(tangle, candidateMilestone.getHash()).snapshotIndex() != candidateMilestone.index()) {
```
GalRogozinski: nitpicking:
I would say that the code will be clearer and easier to debug if it will be written in two lines:
`TransactionViewModel milestoneTx = TransactionViewModel.fromHash(tangle, candidateMilestone.getHash())`
`if (milestoneTx.snapshotIndex() != candidateMilestone.index())`

src/main/java/com/iota/iri/MilestoneTracker.java
```diff
@@ -0,0 +1,549 @@
...
+     * This method analyzes a transaction to determine if it is a milestone.
+     *
+     * In contrast to {@link #validateMilestone} it also updates the internal variables keeping track of the latest
+     * milestone and dumps a log message whenever a change is detected. So this method can be seen as the core logic for
+     * scanning and keeping track of the latest milestones. Since it internally calls {@link #validateMilestone} a
+     * milestone will have its corresponding {@link MilestoneViewModel} created after being processed by this method.
+     *
+     * The method first checks if the transaction originates from the coordinator address, so the check is relatively
+     * cheap and can easily performed on any incoming transaction.
+     *
```
GalRogozinski: Great javadocs!
The only problem is that they are a duplication of the javadocs just below.

Use the `@see` annotation so you won't have to change in 2 places everytime.
e.g.
```
 /**
     * This method analyzes a transaction to determine if it is a milestone.
     *
     * @see #analyzeMilestoneCandidate(TransactionViewModel)
     * @param potentialMilestoneTransactionHash hash of the transaction that shall get checked for being a milestone
     * @return VALID if the transaction is a milestone, INCOMPLETE if the bundle is not complete and INVALID otherwise
     * @throws Exception if something goes wrong while retrieving information from the database
     */
```

You still repeating some stuff but I don't know of a better solution.

src/main/java/com/iota/iri/MilestoneTracker.java
```diff
@@ -0,0 +1,549 @@
...
+                        if(!shuttingDown && analyzedMilestoneCandidates.add(hash) && analyzeMilestoneCandidate(hash) == INCOMPLETE) {
+                            analyzedMilestoneCandidates.remove(hash);
+                        }
+
+                        scanningMilestonesProgress.progress();
+                    }
+                    scanningMilestonesProgress.finish();
+
+                    // allow the "Solid Milestone Tracker" to continue if we finished the first run in rescanning mode
+                    if(firstRun && isRescanning) {
```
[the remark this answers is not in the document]
GalRogozinski: Shouldn't there be a `finally` block as well?

src/main/java/com/iota/iri/MilestoneTracker.java
```diff
@@ -0,0 +1,549 @@
...
+        // block write access to the ledger from other threads
+        lock.writeLock().lock();
+
+        // create a progress logger and start the logging
+        ProgressLogger hardResetLogger = new ProgressLogger(
+            "Resetting ledger to milestone " + targetMilestone.index() + " due to \"" + reason + "\"", log
+        ).start(latestMilestoneIndex - targetMilestone.index() + 2); // +1 for softReset and +1 for starting milestone
+
+        // prune all potentially invalid database fields
+        try {
```
GalRogozinski: should be immediately after `lock`

src/main/java/com/iota/iri/MilestoneTracker.java
```diff
@@ -0,0 +1,549 @@
...
+            MilestoneViewModel currentMilestone = targetMilestone;
+            while(currentMilestone != null) {
+                // reset the snapshotIndex of the milestone and its referenced approvees + it's StateDiff
+                resetSnapshotIndexOfMilestone(currentMilestone);
+                tangle.delete(StateDiff.class, currentMilestone.getHash());
+
+                int currentStep = latestMilestoneIndex - currentMilestone.index() + 2 - hardResetLogger.getStepCount();
```
GalRogozinski: Actually this is a good place for a comment. How do you decide on steps?

src/main/java/com/iota/iri/MilestoneTracker.java
```diff
@@ -0,0 +1,549 @@
...
+        }
+
+        // after we have cleaned up the database we do a soft reset to rescan the existing data
+        softReset();
+
+        // dump message when we are done
+        hardResetLogger.finish();
+
+        // unblock write access to the ledger from other threads
+        lock.writeLock().unlock();
```
GalRogozinski: should be in `finally`

src/main/java/com/iota/iri/MilestoneTracker.java
```diff
@@ -0,0 +1,549 @@
...
+     * since it's milestoneIndex would still be 0.
+     *
+     * @param currentMilestone the milestone that shall have its confirmed transactions reset
+     * @throws Exception if something goes wrong while accessing the database
+     */
+    public void resetSnapshotIndexOfMilestone(MilestoneViewModel currentMilestone) throws Exception {
+        // initialize variables used for traversing the graph
+        TransactionViewModel milestoneTransaction = TransactionViewModel.fromHash(tangle, currentMilestone.getHash());
+        Set<Hash> seenMilestoneTransactions = new HashSet<>();
+        Queue<TransactionViewModel> transactionsToExamine = new LinkedList<>(Collections.singleton(milestoneTransaction));
```
GalRogozinski: ArrayLists are usually faster, so probably use `ArrayDeque`. 
If you can estimate its size (doesn't have to be exact) you can improve performance.

src/main/java/com/iota/iri/MilestoneTracker.java
```diff
@@ -0,0 +1,549 @@
...
+        TransactionViewModel currentTransaction;
+        while((currentTransaction = transactionsToExamine.poll()) != null) {
+            if(seenMilestoneTransactions.add(currentTransaction.getHash())) {
+                // reset the snapshotIndex to allow a repair
+                currentTransaction.setSnapshot(tangle, 0);
+
+                // only examine transactions that still belong to our milestone
+                if(!Hash.NULL_HASH.equals(currentTransaction.getBranchTransactionHash())) {
+                    TransactionViewModel branchTransaction = currentTransaction.getBranchTransaction(tangle);
+                    if(branchTransaction.getType() != TransactionViewModel.PREFILLED_SLOT && branchTransaction.snapshotIndex() >= currentMilestone.index()) {
```
GalRogozinski: if you use `>=` I find it confusing with the comment "belong to our milestone". Perhaps change the comment?

src/main/java/com/iota/iri/MilestoneTracker.java
```diff
@@ -0,0 +1,549 @@
...
+                        transactionsToExamine.add(branchTransaction);
+                    }
+                }
+
+                // only examine transactions that still belong to our milestone
+                if(!Hash.NULL_HASH.equals(currentTransaction.getTrunkTransactionHash())) {
+                    TransactionViewModel trunkTransaction = currentTransaction.getTrunkTransaction(tangle);
```
GalRogozinski: You have repeated logic here.
Consider:
int milestoneIndex = currentMilestone.index()
```
TransactionViewModel tx = currentTransaction.getBranchTransaction(tangle);
queueIfBelongsToMilestone(tx, transactionsToExamine, milestoneIndex)
tx = currentTransaction.getTrunkTransaction(tangle)
queueIfBelongsToMilestone(tx, transactionsToExamine, milestoneIndex)
```

This way you only have to change logic in one place.

src/main/java/com/iota/iri/MilestoneTracker.java
```diff
@@ -0,0 +1,549 @@
...
+                    if(trunkTransaction.getType() != TransactionViewModel.PREFILLED_SLOT && trunkTransaction.snapshotIndex() >= currentMilestone.index()) {
+                        transactionsToExamine.add(trunkTransaction);
+                    }
+                }
+            }
+        }
+    }
+
+    public Validity validateMilestone(SpongeFactory.Mode mode, TransactionViewModel transactionViewModel, int index) throws Exception {
+        if (index < 0 || index >= 0x200000) {
```
GalRogozinski: I guess I am missing some IOTA knowledge. Where is the upper bound coming from?

Old code actually, I give this question to @alon-e :-)

GalRogozinski:
closed in favor of #1001

PR #944, opened 2018-08-21
Added defaults for remote-limt-api so sensitive calls are blocked by …

### LINKED ISSUE #306: Restrict --remote flag to opt-in sensitive API calls instead of opt-out
A few days ago some of us on Slack discussed a potential issue with the handling of `--remote` flag of IRI. By default it opens the door to `getNeighbors`, `addNeighbors` and `removeNeighbors` if you don't set to restrict them by `--remote-limit-api`. We came to the conclusion that it might be better to *opt-in* those sensitive API calls for `--remote` instead of *opt-out* them because the average user does not know or care about it and opens unintentionally the door for potential harmful operations (gathering IPs, tampering with neighbors). Additionally a lot of users post their IPs on slack despite the warning not to do so. We think that most users of `--remote` don't use those API calls and the low effort to change this is worth the improvement of security.

kwek20:
# Description
Provides defaults for keeping an iri node secure when you do not provide flags.

Fixes #306

## Type of change

# How Has This Been Tested?
Started default IRI node with port parameter -> correct calls are blocked
Started IRI with --remote-api-limit for just blocking getNeighbors, and only that call is blocked

# Checklist:
- [ ] My code follows the style guidelines for this project

### DIFF
--- a/src/main/java/com/iota/iri/conf/BaseIotaConfig.java
+++ b/src/main/java/com/iota/iri/conf/BaseIotaConfig.java
@@ -604,7 +604,7 @@ public abstract class BaseIotaConfig implements IotaConfig {
 		//API
         int API_PORT = 14265;
         String API_HOST = "localhost";
-        List<String> REMOTE_LIMIT_API = IotaUtils.createImmutableList("addNeighbors", "getNeighbors", "removeNeighbors");
+        List<String> REMOTE_LIMIT_API = IotaUtils.createImmutableList("addNeighbors", "getNeighbors", "removeNeighbors", "attachToTangle", "interruptAttachingToTangle");
         int MAX_FIND_TRANSACTIONS = 100_000;
         int MAX_REQUESTS_LIST = 1_000;
         int MAX_GET_TRYTES = 10_000;
--- a/src/main/java/com/iota/iri/conf/BaseIotaConfig.java
+++ b/src/main/java/com/iota/iri/conf/BaseIotaConfig.java
@@ -601,7 +601,7 @@ public abstract class BaseIotaConfig implements IotaConfig {
     }
 
     public interface Defaults {
-		//API
+        //API
         int API_PORT = 14265;
         String API_HOST = "localhost";
         List<String> REMOTE_LIMIT_API = IotaUtils.createImmutableList("addNeighbors", "getNeighbors", "removeNeighbors", "attachToTangle", "interruptAttachingToTangle");
--- a/src/main/java/com/iota/iri/conf/BaseIotaConfig.java
+++ b/src/main/java/com/iota/iri/conf/BaseIotaConfig.java
@@ -9,6 +9,7 @@ import com.iota.iri.utils.IotaUtils;
 import org.apache.commons.lang3.ArrayUtils;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 /*
@@ -25,7 +26,7 @@ public abstract class BaseIotaConfig implements IotaConfig {
     //API
     protected int port = Defaults.API_PORT;
     protected String apiHost = Defaults.API_HOST;
-    protected List<String> remoteLimitApi = new ArrayList<>();
+    protected List<String> remoteLimitApi = new ArrayList<>(Arrays.asList(Defaults.REMOTE_LIMIT_API));
     protected int maxFindTransactions = Defaults.MAX_FIND_TRANSACTIONS;
     protected int maxRequestsList = Defaults.MAX_REQUESTS_LIST;
     protected int maxGetTrytes = Defaults.MAX_GET_TRYTES;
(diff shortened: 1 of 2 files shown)


src/main/java/com/iota/iri/conf/BaseIotaConfig.java
```diff
@@ -600,9 +601,10 @@ protected void setBelowMaxDepthTransactionLimit(int maxAnalyzedTransactions) {
     }
 
     public interface Defaults {
-        //API
+		//API
         int API_PORT = 14265;
         String API_HOST = "localhost";
+        String[] REMOTE_LIMIT_API = new String[] {"addNeighbors", "getNeighbors", "removeNeighbors"};
```
alon-e: I think we should also add `attachToTangle` & `interruptAttachingToTangle` to the default list.

src/main/java/com/iota/iri/conf/BaseIotaConfig.java
```diff
@@ -600,9 +601,10 @@ protected void setBelowMaxDepthTransactionLimit(int maxAnalyzedTransactions) {
     }
 
     public interface Defaults {
-        //API
+		//API
```
alon-e: fix indentation

src/main/java/com/iota/iri/conf/BaseIotaConfig.java
```diff
@@ -25,7 +26,7 @@
     //API
     protected int port = Defaults.API_PORT;
     protected String apiHost = Defaults.API_HOST;
-    protected List<String> remoteLimitApi = new ArrayList<>();
+    protected List<String> remoteLimitApi = new ArrayList<>(Arrays.asList(Defaults.REMOTE_LIMIT_API));
```
GalRogozinski: Hmm I think wrapping with `Collections.unmodifiableList()` would have been more suitable than ArrayList...

src/main/java/com/iota/iri/conf/BaseIotaConfig.java
GalRogozinski: Change to `List<String> REMOTE_LIMIT_API`

In `IotaUtils` you can add the following method:
```
 public static <T> List<T> createImmutableList(T... a) {
        return Collections.unmodifiableList(Arrays.asList(a));
    }
```

Note that once we upgrade Java we can start using `List.of()` instead of doing this ugly thing.
See:
https://docs.oracle.com/javase/9/docs/api/java/util/List.html

src/main/java/com/iota/iri/conf/BaseIotaConfig.java
kwek20: Not sure why it did that, i checked in code and it doesnt look like that :o

src/main/java/com/iota/iri/conf/BaseIotaConfig.java
kwek20: @alon-e isnt attachToTangle used in a lot of tutorials? That will probably confuse people if its disabled by default

GalRogozinski: Since this is also a product decision I am tagging @jakubcech.
I want to remind everyone that since changing the defaults breaks backwards compatibility, perhaps it shouldn't be released together with the changes we made for configurations in #724.

jakubcech: `attachToTangle` & `interruptAttachingToTangle` should be in the list. We should update our docs if it's broken by this change somewhere.

src/main/java/com/iota/iri/conf/BaseIotaConfig.java
rajivshah3: It might happen if you use tabs instead of spaces (https://www.youtube.com/watch?v=SsoOG6ZeyUI :trollface: )

GalRogozinski: Hmm I don't see it fixed yet... You can configure your IDE to convert tabs to spaces.

src/main/java/com/iota/iri/conf/BaseIotaConfig.java
```diff
@@ -600,9 +601,10 @@ protected void setBelowMaxDepthTransactionLimit(int maxAnalyzedTransactions) {
     }
 
     public interface Defaults {
-        //API
+		//API
         int API_PORT = 14265;
         String API_HOST = "localhost";
+        List<String> REMOTE_LIMIT_API = IotaUtils.createImmutableList("addNeighbors", "getNeighbors", "removeNeighbors", "attachToTangle", "interruptAttachingToTangle");
```
GalRogozinski: Can you set your IDE to have a column width of 120.
I will try to add more checkstyle rules real soon. I know style comments are annoying. A computer should give them and not a human.

Issue #931, opened 2018-08-12
Flag names are not always the same as INI configuration names


GalRogozinski:
### Bug description
Follow up on #910 
Some config flag names don't have the same INI config name.
For example, `--db` flag is `MAIN_DB=` in INI file.

Changing this however may hurt backwards compatibility.

### Expected Behaviour
Flag names should always have the same name as INI config.

GalRogozinski:
Another example that we should take care of is the `--remote` flag. It currently changes the `API_HOST` config. Does it make sense to have a "pseudo" configuration that affects another configuration?

PR #953, opened 2018-08-23
`--remote` flag will enable API interface to all hosts

### LINKED ISSUE #952: --remote flag replaced for --api-host in PR #910
### Bug description
`--remote` switch removal broke backwards compatibility.

### Expected behaviour
API listener binds on `0.0.0.0`.

### Actual behaviour
API listener binds only to `127.0.0.1`.

GalRogozinski:
# Description
The `--remote` flag that opens the api interface to all hosts was accidentally removed. It is now returned

Fixes #952 

## Type of change

# How Has This Been Tested?
Unit tests were added

- `testRemoteFlag()` ascertains that `api_host == 0.0.0.0`
- `testIniParsingTestnet()` now makes sure that `REMOTE` in INI has no effect (This conforms to backwards compatibility)

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/conf/BaseIotaConfig.java
+++ b/src/main/java/com/iota/iri/conf/BaseIotaConfig.java
@@ -3,6 +3,7 @@ package com.iota.iri.conf;
 import com.beust.jcommander.JCommander;
 import com.beust.jcommander.Parameter;
 import com.beust.jcommander.ParameterException;
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.iota.iri.IRI;
 import com.iota.iri.utils.IotaUtils;
@@ -31,6 +32,8 @@ public abstract class BaseIotaConfig implements IotaConfig {
     protected int maxGetTrytes = Defaults.MAX_GET_TRYTES;
     protected int maxBodyLength = Defaults.MAX_BODY_LENGTH;
     protected String remoteAuth = Defaults.REMOTE_AUTH;
+    //We don't have a REMOTE config but we have a remote flag. We must add a field for JCommander
+    private boolean remote;
 
 
     //Network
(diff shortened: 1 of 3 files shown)


src/main/java/com/iota/iri/conf/BaseIotaConfig.java
```diff
@@ -31,6 +32,8 @@
     protected int maxGetTrytes = Defaults.MAX_GET_TRYTES;
     protected int maxBodyLength = Defaults.MAX_BODY_LENGTH;
     protected String remoteAuth = Defaults.REMOTE_AUTH;
+    //We don't have a REMOTE config but we have a remote flag. We must add a field for JCommander
+    private boolean remote;
```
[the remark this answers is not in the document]
GalRogozinski: Must be used becuase of JCommander

src/test/java/com/iota/iri/conf/ConfigTest.java
```diff
@@ -207,6 +216,8 @@ public void testIniParsingTestnet() throws Exception {
                 .append("NUMBER_OF_KEYS_IN_A_MILESTONE = 3").append(System.lineSeparator())
                 .append("DONT_VALIDATE_TESTNET_MILESTONE_SIG = true").append(System.lineSeparator())
                 .append("TIPSELECTION_ALPHA = 1.1").append(System.lineSeparator())
+                //doesn't do anything
+                .append("REMOTE")
```
alon-e: I think this will be confusing - there is no real reason why `REMOTE` should not work in the config file - I would not call that "not backward compatible".

GalRogozinski: The thing is that I don't know if it is correct that you have two ways to manipulate API_HOST...
I can totally imagine someone setting his API host to some IP. Later on he will google how to open his node to the world and add the REMOTE option. He won't notice the API_HOST becuase there are so many configs...

The same problem can also exist with flags, but it is less bad since you have to retype them.

alon-e: ok, I agree for now.
Can you open an issue about that then?

GalRogozinski: I added a comment on https://github.com/iotaledger/iri/issues/931 regarding this problem.

PR #962, opened 2018-08-30
Test/db benchmark

### LINKED ISSUE #924: Profile the rocksDB performance in deleting, fetching and writing
### Description
Profile the rocksDB performance in deleting and writing. What happens when we start deleting or writing a large number of transactions? Just a one-off pick up of numbers, not continuous testing.

Do we just enable the existing tests? testManyTxInDb in the transactionViewModelTest. 

Right after restarting IRI and when IRI has been running for a while. 

Collect the data and analyze it to some extent, how do the data change in different batch sizes. 

### Motivation
Identify the bottlenecks in deleting and writing in the current RocksDB implementation. 

#### Writing
- [ ] How long does it take to write a transaction after restarting node
- [ ] How long does it take to write a batch in different sizes (10, 100, 500, 1000, 3000) of transactions after restarting node 
- [ ] How long does it take to write a transaction after a synced node has been running for a while (no process is working hard on getting up to speed)
- [ ] How long does it take to write a batch in different sizes (10, 100, 500, 1000, 3000) of transactions after a synced IRI has been running for a while (no process is working hard on getting up to speed)
#### Deleting
- [ ] How long does it take to delete a transaction after restarting node
- [ ] How long does it take to delete a batch of transactions in different batch sizes (10, 100, 500, 1000, 3000) of transactions after restarting IRI
- [ ] How long does it take to delete a transaction after a synced node has been running for a while (no process is working hard on getting up to speed)
- [ ] How long does it take to delete a batch of transactions in different sizes (10, 100, 500, 1000, 3000) of transactions after a synced IRI has been running for a while (no process is working hard on getting up to speed)
#### Fetching
- [ ] How long does it take to fetch a transaction after restarting node
- [ ] How long does it take to fetch a batch of transactions in different batch sizes (10, 100, 500, 1000, 3000) of transactions after restarting IRI
- [ ] How long does it take to fetch a transaction after a synced node has been running for a while (no process is working hard on getting up to speed)
- [ ] How long does it take to fetch a batch of transactions in different sizes (10, 100, 500, 1000, 3000) of transactions after a synced IRI has been running for a while (no process is working hard on getting up to speed)

GalRogozinski:
# Description
This utilizes JMH (Java Microbenchmark Harness) to profile IRI's db provider performance.

We have the following test:
- Persisting transactions
- Fetching transcations
- Deleting transactions in loop
- Deleting transactions in batch
- Dropping columns

After running `mvn clean install` the easiest way to run is to do: `mvn surefire:test -Dtest=BenchmarkRunner`. I decided not to add them to the `post-integration-test` phase because I imagine we will have buildkite running it in separate.
 
Fixes #924 

## Type of change

# How Has This Been Tested?
Maven had been run to make sure the benchmark is excluded in the normal built.
The benchmarks themselves were run on my laptop and on a Hetzner machine.

# Checklist:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -177,6 +177,19 @@
             <version>1.72</version>
         </dependency>
 
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-core</artifactId>
+            <version>1.19</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.openjdk.jmh</groupId>
+            <artifactId>jmh-generator-annprocess</artifactId>
+            <version>1.19</version>
+            <scope>test</scope>
+        </dependency>
+
     </dependencies>
 
     <build>
@@ -352,6 +365,7 @@
                 <configuration>
                     <excludes>
                         <exclude>integration/*.java</exclude>
+                        <exclude>benchmarks/**/*</exclude>
                     </excludes>
                 </configuration>
                 <executions>
--- a/pom.xml
+++ b/pom.xml
@@ -203,6 +203,7 @@
                     <target>${java-version}</target>
                     <testSource>${java-version}</testSource>
                     <testTarget>${java-version}</testTarget>
+                    <useIncrementalCompilation>false</useIncrementalCompilation>
                 </configuration>
             </plugin>
             <plugin>
--- a/src/test/java/com/iota/iri/benchmarks/BenchmarkRunner.java
+++ b/src/test/java/com/iota/iri/benchmarks/BenchmarkRunner.java
@@ -1,0 +1,31 @@
+package com.iota.iri.benchmarks;
+
+import com.iota.iri.benchmarks.dbbenchmark.RocksDbBenchmark;
+import org.junit.Test;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.runner.Runner;
+import org.openjdk.jmh.runner.RunnerException;
+import org.openjdk.jmh.runner.options.Options;
+import org.openjdk.jmh.runner.options.OptionsBuilder;
+
+import java.util.concurrent.TimeUnit;
+
+public class BenchmarkRunner {
+
+    @Test
+    public void launchDbBenchmarks() throws RunnerException {
+        Options opts = new OptionsBuilder()
+                .include(RocksDbBenchmark.class.getName() + ".*")
+                .mode(Mode.AverageTime)
+                .timeUnit(TimeUnit.MILLISECONDS)
+                .warmupIterations(5)
+                .forks(1)
+                .measurementIterations(10)
+                .shouldFailOnError(true)
+                .shouldDoGC(false)
+                .build();
+
+        //possible to do assertions over run results
+        new Runner(opts).run();
+    }
+}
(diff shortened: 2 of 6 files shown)


pom.xml
```diff
@@ -190,6 +203,7 @@
                     <target>${java-version}</target>
                     <testSource>${java-version}</testSource>
                     <testTarget>${java-version}</testTarget>
+                    <useIncrementalCompilation>false</useIncrementalCompilation>
```
GalRogozinski: Without this some maven goals fail and it makes it harder to run the test (which is why this is a part of this PR).
I observed no performance loss, and  [apache](https://cwiki.apache.org/confluence/display/MAVEN/Incremental+Builds) and [maven users](https://stackoverflow.com/questions/8918165/does-maven-support-incremental-builds) seem to be on my side here.

pom.xml
```diff
@@ -352,6 +366,7 @@
                 <configuration>
                     <excludes>
                         <exclude>integration/*.java</exclude>
+                        <exclude>benchmarks/**/*</exclude>
```
GalRogozinski: Since my test doesn't follow maven name conventions(i.e doesn't end with Test) it will be excluded without this line.
But I added this anyhows :-P.

src/test/java/com/iota/iri/benchmarks/BenchmarkRunner.java
```diff
@@ -0,0 +1,31 @@
...
+public class BenchmarkRunner {
+
+    @Test
+    public void launchDbBenchmarks() throws RunnerException {
+        Options opts = new OptionsBuilder()
+                .include(RocksDbBenchmark.class.getName() + ".*")
+                .mode(Mode.AverageTime)
+                .timeUnit(TimeUnit.MILLISECONDS)
+                .warmupIterations(5)
+                .forks(1)
```
GalRogozinski: official docs recommend more forks. But since the test is already extremely long and I didn't see those forks help very much, I decreased it to 1.

src/test/java/com/iota/iri/benchmarks/dbbenchmark/states/DbState.java
```diff
@@ -0,0 +1,89 @@
...
+        for (int i = 0; i < numTxsToTest; i++) {
+            trytes = nextWord(trytes);
+            TransactionViewModel tvm = TransactionTestUtils.createTransactionWithTrytes(trytes);
+            transactions.add(tvm);
+        }
+        transactions = Collections.unmodifiableList(transactions);
+    }
+
+    public void teardown() throws Exception {
+        System.out.println("-----------------------trial teardown--------------------------------");
```
GalRogozinski: JMH doesn't print the result in a log anyhows, so I used primitive printing as well.

src/test/java/com/iota/iri/benchmarks/dbbenchmark/states/DbState.java
```diff
@@ -0,0 +1,74 @@
...
+    private List<TransactionViewModel> transactions;
+
+    @Param({"10", "100", "500", "1000", "3000"})
+    private int numTxsToTest;
+
+
+    public void setup() throws Exception {
+        System.out.println("-----------------------trial setup--------------------------------");
+        boolean mkdirs = dbFolder.mkdirs();
+        System.out.println("mkdirs success: " + mkdirs );
```
kwek20: We should probably directly fail/throw an error if this fails.
If it succeeds there's no need to log it right? (less=more)

GalRogozinski: :+1:

PR #978, opened 2018-09-09
SSL setup via docker-compose

### LINKED ISSUE #950: Document setting up SSL with IRI
### Description
Create and document docker compose that will front IRI with a SSL-enabled nginx. 

Can also be inspired by Playbook and isolate a simple solution.

Example guide
http://www.automationlogic.com/using-lets-encrypt-and-docker-for-automatic-ssl/

### Motivation
Provide an easy way to set up SSL with IRI. 

### Requirements
- [ ] Find the right docker image or create our own. Nginx with Let's Encrypt and IRI latest.
- [ ] A docker compose that pieces all of this together.
- [ ] Document how to use the docker-compose with IRI on docs.iota.org.

rajivshah3:
# Description
Uses docker-compose to accomplish the following:

- Obtain a SSL certificate from Let's Encrypt
- Configure `nginx` proxy-pass
- Set up IRI container

Closes #950 

## Type of change

# How Has This Been Tested?

- All unit tests pass
- Local and remote calls to `getNodeInfo` on port 443 return successfully 
- Remote calls to `getNeighbors` return `COMMAND getNeighbors is not available on this node` (as specified by the IRI Dockerfile)

# Checklist:


docker/ssl/docker-compose.yml
```diff
@@ -0,0 +1,30 @@
...
+  volumes:
+    - /etc/letsencrypt
+    - /var/lib/letsencrypt
+  ports:
+    - "80"
+    - "443"
+  environment:
+    - TERM=xterm
+iri:
+  image: iotaledger/iri:v1.5.3
```
karimodm: This should change on every release. Should we change it to `:latest` ?

docker/ssl/docker-compose.yml
```diff
@@ -0,0 +1,30 @@
...
+  environment:
+    - MY_DOMAIN_NAME=__DOMAIN__
+    - PORT=__PORT__
+  ports:
+    - "80:80"
+    - "443:443"
+  volumes_from:
+    - letsencrypt
+letsencrypt:
+  image: quay.io/letsencrypt/letsencrypt:latest
```
karimodm: I don't really like the idea of pointing to a non-official image for this. Although there are none available. I suggest therefore to pull from a more reputable and long-term image source (https://hub.docker.com/r/linuxserver/letsencrypt/ seems to serve the same functionality) and to point to it by digest, rather than by tag: so we make sure we are always pulling the image we have reviewed with our own eyes, instead of relying on a floating docker image tag.
For example you can pull the image `linuxserver/letsencrypt:170` by its hash: `docker pull linuxserver/letsencrypt@sha256:ab7cafe902eed5b606ebf4b468f826d9cd5846ea0905f844554fd56ad6a78bdc`

docker/ssl/start.sh
```diff
@@ -0,0 +1,58 @@
...
+  if [[ "$TOS" != "y" ]]; then
+    echo "Error: You must agree to the ToS in order to continue"
+    exit 1
+  fi
+fi
+
+# Modify docker-compose.yml
+sed -i "" "s/__DOMAIN__/$DOMAIN/g" ./docker-compose.yml
+sed -i "" "s/__EMAIL__/$EMAIL/g" ./docker-compose.yml
+sed -i "" "s/__PORT__/$PORT/g" ./docker-compose.yml
```
karimodm: I believe you have an extra `""` after the `-i`

docker/ssl/nginx-image/dhparams.pem
```diff
@@ -0,0 +1,8 @@
+-----BEGIN DH PARAMETERS-----
```
karimodm: Even if this file is not a secret, it should be generated rather than static.

docker/ssl/nginx-image/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM nginx:stable
+
+# Install dependencies and then remove apt-get lists to save space
+RUN apt-get update && apt-get install -y procps && rm /var/lib/apt/lists/*
```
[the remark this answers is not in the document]
karimodm: Agree with this

karimodm: `rm /var/lib/apt/lists/*` you need to make this command recursive `rm -rf rm /var/lib/apt/lists/*`

docker/ssl/docker-compose.yml
```diff
@@ -0,0 +1,30 @@
...
+  command:  bash -c "sleep 6 && certbot certonly --standalone --email "__EMAIL__" -d "__DOMAIN__" -n --agree-tos"
```
karimodm: Why the `sleep 6`? Are you trying to coordinate the `nginx` container startup with the `letsencrypt` one?

docker/ssl/docker-compose.yml
```diff
@@ -0,0 +1,30 @@
...
+  ports:
+    - "80"
+    - "443"
+  environment:
+    - TERM=xterm
+iri:
+  image: iotaledger/iri:v1.5.3
+  command: -p __PORT__
+  ports:
+    - "__PORT__:__PORT__"
```
karimodm: This will publish the plaintext IRI API port, do we want to do this? I believe that we should only expose Nginx ports: HTTP and HTTPS.
You can put the `nginx` and the `iri` containers on the same docker network / or linking them like you already did, and use the internal Docker DNS to resolve IRI's container's IP from `nginx` instead of `sed`ing `__APPLICATION_IP__`

docker/ssl/nginx-image/nginx-secure.conf
```diff
@@ -0,0 +1,51 @@
...
+			proxy_set_header X-Forwarded-Proto https;
+		}
+
+		location / {
+      add_header "Access-Control-Allow-Origin"  *;
+      proxy_set_header Host \$host;
+      proxy_set_header X-Real-IP \$remote_addr;
+      proxy_set_header X-Forwarded-For \$proxy_add_x_forwarded_for;
+      proxy_set_header X-Forwarded-Proto \$scheme;
+      proxy_pass http://___APPLICATION_IP___:___APPLICATION_PORT___;
```
karimodm: Better using internal Docker IP of the IRI's container.

docker/ssl/docker-compose.yml
rajivshah3: Sounds good, I agree with your idea of pointing to the digest. However, I believe the quay.io is official. I just realized that they also push to https://hub.docker.com/r/certbot/certbot/ and will be completely transitioning there soon (the only difference is that the one on Docker Hub is based on Alpine Linux). If you want me to switch to that one instead, I can

docker/ssl/start.sh
rajivshah3: I think this was necessary for macOS (I use a Mac). I'll see if I can find the SO thread that told me to add `""`. If this causes it to break on Linux I can remove it

docker/ssl/nginx-image/Dockerfile
rajivshah3: 🤦‍♂️ not sure how I missed that

docker/ssl/docker-compose.yml
karimodm: Digest is good enough for me, I will leave to you the choice of following the official EFF one or not.

docker/ssl/start.sh
karimodm: Yep, it breaks on the GNU version.

docker/ssl/nginx-image/start.sh
```diff
@@ -0,0 +1,72 @@
+#!/usr/bin/env bash
+
+sleep 3
```
karimodm: This makes the letsencrypt container to attempt verification before nginx is up to serve the `/.well-known` URI, thus creating a race condition.

docker/ssl/nginx-image/start.sh
```diff
@@ -0,0 +1,72 @@
...
+
+echo build starting nginx config
+
+echo replacing ___my.example.com___/$MY_DOMAIN_NAME
+echo replacing ___LETSENCRYPT_IP___/$LETSENCRYPT_PORT_80_TCP_ADDR
```
karimodm: By using $LETSENCRYPT_* variables you are using a deprecated docker compose feature https://docs.docker.com/compose/link-env-deprecated/ . This got me baffled when I first read the script, wondering where these were coming from (the [PR](https://github.com/xibosignage/xibo-docker/pull/47/files) from which you ported the behaviour is more than an year old indeed, which is a *lot* of time in the Docker world :stuck_out_tongue: )
I would suggest to use once again the Docker embedded DNS to resolve to obtain the address of the letsencrypt container.

docker/ssl/nginx-image/start.sh
```diff
@@ -0,0 +1,72 @@
...
+echo replacing ___LETSENCRYPT_PORT___/$LETSENCRYPT_PORT_80_TCP_PORT
```
karimodm: For the ports you can just hardcode 80 and 443.

docker/ssl/start.sh
```diff
@@ -0,0 +1,58 @@
...
+  read DOMAIN
+  if [[ -z "$DOMAIN" ]]; then
+    echo "Error: You must enter a domain"
+    exit 1
+  fi
+fi
+
+# Port
+if [[ -z "$PORT" ]]; then
+  echo "Please enter your desired IRI API port:"
```
karimodm: Do we still need this since we do not expose the API port directly anymore?

docker/ssl/nginx-image/nginx-secure.conf
```diff
@@ -0,0 +1,51 @@
...
+
+		location /.well-known/acme-challenge {
+			proxy_pass http://___LETSENCRYPT_IP___:___LETSENCRYPT_PORT___;
+			proxy_set_header Host            $host;
+			proxy_set_header X-Forwarded-For $remote_addr;
+			proxy_set_header X-Forwarded-Proto https;
+		}
+
+		location / {
+			return         301 https://$server_name$request_uri;
```
karimodm: Will wallets / clients follow the redirect to the SSL version? In addition all our API calls are POST, and by default clients won't follow redirects on a POST, thus switching to the GET method, resulting in a 500 from the API.
I would say to just send the plain request to IRI. 80 for plain HTTP and 443 for HTTPS. 


```
* Rebuilt URL to: http://irissl.EDITED/
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0*   Trying 52.49.30.40...
* TCP_NODELAY set
* Connected to irissl.EDITED (52.49.30.40) port 80 (#0)
> POST / HTTP/1.1
> Host: irissl.EDITED
> User-Agent: curl/7.61.0
> Accept: */*
> X-IOTA-API-VERSION: 1.4.1.2
> Content-Length: 28
> Content-Type: application/x-www-form-urlencoded
> 
} [28 bytes data]
* upload completely sent off: 28 out of 28 bytes
< HTTP/1.1 301 Moved Permanently
< Server: nginx/1.14.0
< Date: Mon, 17 Sep 2018 16:31:31 GMT
< Content-Type: text/html
< Content-Length: 185
< Connection: keep-alive
< Location: https://irissl.EDITED/
< 
* Ignoring the response-body
{ [185 bytes data]

100   213  100   185  100    28   1697    256 --:--:-- --:--:-- --:--:--  1972
* Connection #0 to host irissl.EDITED left intact
* Issue another request to this URL: 'https://irissl.EDITED/'
* Switch from POST to GET
*   Trying 52.49.30.40...
* TCP_NODELAY set
* Connected to irissl.EDITED (52.49.30.40) port 443 (#1)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: none
  CApath: /etc/ssl/certs
} [5 bytes data]
* (304) (OUT), TLS handshake, Client hello (1):
} [512 bytes data]
* (304) (IN), TLS handshake, Server hello (2):
{ [108 bytes data]
* TLSv1.2 (IN), TLS handshake, Certificate (11):
{ [2737 bytes data]
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
{ [300 bytes data]
* TLSv1.2 (IN), TLS handshake, Server finished (14):
{ [4 bytes data]
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
} [37 bytes data]
* TLSv1.2 (OUT), TLS change cipher, Client hello (1):
} [1 bytes data]
* TLSv1.2 (OUT), TLS handshake, Finished (20):
} [16 bytes data]
* TLSv1.2 (IN), TLS handshake, Finished (20):
{ [16 bytes data]
* SSL connection using TLSv1.2 / ECDHE-RSA-AES256-GCM-SHA384
* ALPN, server accepted to use http/1.1
* Server certificate:
*  subject: CN=irissl.EDITED
*  start date: Sep 17 15:05:15 2018 GMT
*  expire date: Dec 16 15:05:15 2018 GMT
*  subjectAltName: host "irissl.EDITED" matched cert's "irissl.EDITED"
*  issuer: C=US; O=Let's Encrypt; CN=Let's Encrypt Authority X3
*  SSL certificate verify ok.
} [5 bytes data]
> GET / HTTP/1.1
> Host: irissl.EDITED
> User-Agent: curl/7.61.0
> Accept: */*
> X-IOTA-API-VERSION: 1.4.1.2
> 
{ [5 bytes data]
< HTTP/1.1 500 Internal Server Error
< Server: nginx/1.14.0
< Date: Mon, 17 Sep 2018 16:31:31 GMT
< Content-Type: application/json
< Content-Length: 66
< Connection: keep-alive
< Access-Control-Allow-Origin: *
< 
{ [66 bytes data]

100    66  100    66    0     0    164      0 --:--:-- --:--:-- --:--:--   164
* Connection #1 to host irissl.EDITED left intact
{"exception":"Invalid request payload: \u0027\u0027","duration":0}
```

jakubcech:
Hi @rajivshah3, will you have room to address the above comments please? :)

rajivshah3:
Hi @jakubcech , sorry for the delay. I've been busy with a lot of other stuff lately. I'll see if I have some time today

docker/ssl/start.sh
rajivshah3: This is still needed in order to tell `nginx` which port to forward requests to. It's also necessary for the IRI docker image:
> Run the official iotaledger/iri container, passing the mandatory -p option:
> `docker run iotaledger/iri:v1.5.4 -p 14265`

https://github.com/iotaledger/iri/blob/dev/DOCKER.md

karimodm: I am just saying we can always just go for the default 14265 one. But it's fine to me anyway.

jakubcech:
Good stuff!

rajivshah3:
Thanks @karimodm @jakubcech !

PR #981, opened 2018-09-11
Feat: added config variables for local snapshots

hmoog:
# Description

This PR introduces the config variables required for local snapshots.

        String LOCAL_SNAPSHOTS_ENABLED = "Flag that determines if local snapshots are enabled.";
        String LOCAL_SNAPSHOTS_PRUNING_ENABLED = "Flag that determines if pruning of old data is enabled.";
        String LOCAL_SNAPSHOTS_PRUNING_DELAY = "Only prune data that precedes the local snapshot by n milestones.";
        String LOCAL_SNAPSHOTS_INTERVAL_SYNCED = "Take local snapshots every n milestones if the node is fully synced.";
        String LOCAL_SNAPSHOTS_INTERVAL_UNSYNCED = "Take local snapshots every n milestones if the node is syncing.";
        String LOCAL_SNAPSHOTS_DEPTH = "Number of milestones to keep.";
        String LOCAL_SNAPSHOTS_BASE_PATH = "Path to the snapshot files (without file extensions).";

### DIFF
--- a/src/main/java/com/iota/iri/conf/BaseIotaConfig.java
+++ b/src/main/java/com/iota/iri/conf/BaseIotaConfig.java
@@ -80,6 +80,15 @@ public abstract class BaseIotaConfig implements IotaConfig {
     protected double alpha = Defaults.ALPHA;
     private int maxAnalyzedTransactions = Defaults.MAX_ANALYZED_TXS;
 
+    //Snapshot
+    protected boolean localSnapshotsEnabled = Defaults.LOCAL_SNAPSHOTS_ENABLED;
+    protected boolean localSnapshotsPruningEnabled = Defaults.LOCAL_SNAPSHOTS_PRUNING_ENABLED;
+    protected int localSnapshotsPruningDelay = Defaults.LOCAL_SNAPSHOTS_PRUNING_DELAY;
+    protected int localSnapshotsIntervalSynced = Defaults.LOCAL_SNAPSHOTS_INTERVAL_SYNCED;
+    protected int localSnapshotsIntervalUnsynced = Defaults.LOCAL_SNAPSHOTS_INTERVAL_UNSYNCED;
+    protected int localSnapshotsDepth = Defaults.LOCAL_SNAPSHOTS_DEPTH;
+    protected String localSnapshotsBasePath = Defaults.LOCAL_SNAPSHOTS_BASE_PATH;
+
     public BaseIotaConfig() {
         //empty constructor
     }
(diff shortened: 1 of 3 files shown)


src/main/java/com/iota/iri/conf/BaseIotaConfig.java
```diff
@@ -669,6 +755,13 @@ protected void setBelowMaxDepthTransactionLimit(int maxAnalyzedTransactions) {
                 "KPWCHICGJZXKE9GSUDXZYUAPLHAKAHYHDXNPHENTERYMMBQOPSQIDENXKLKCEYCPVTZQLEEJVYJZV9BWU";
 
         //Snapshot
+        boolean LOCAL_SNAPSHOTS_ENABLED = true;
+        boolean LOCAL_SNAPSHOTS_PRUNING_ENABLED = false;
+        int LOCAL_SNAPSHOTS_PRUNING_DELAY = 1000;
+        int LOCAL_SNAPSHOTS_INTERVAL_SYNCED = 10;
+        int LOCAL_SNAPSHOTS_INTERVAL_UNSYNCED = 5000;
+        String LOCAL_SNAPSHOTS_BASE_PATH = "mainnet";
+        int LOCAL_SNAPSHOTS_DEPTH = 500;
```
karimodm: Do we have any sort of "story" backing up these default values?

src/main/java/com/iota/iri/conf/BaseIotaConfig.java
```diff
@@ -669,6 +755,13 @@ protected void setBelowMaxDepthTransactionLimit(int maxAnalyzedTransactions) {
                 "KPWCHICGJZXKE9GSUDXZYUAPLHAKAHYHDXNPHENTERYMMBQOPSQIDENXKLKCEYCPVTZQLEEJVYJZV9BWU";
 
         //Snapshot
+        boolean LOCAL_SNAPSHOTS_ENABLED = true;
+        boolean LOCAL_SNAPSHOTS_PRUNING_ENABLED = false;
+        int LOCAL_SNAPSHOTS_PRUNING_DELAY = 1000;
+        int LOCAL_SNAPSHOTS_INTERVAL_SYNCED = 10;
+        int LOCAL_SNAPSHOTS_INTERVAL_UNSYNCED = 5000;
+        String LOCAL_SNAPSHOTS_BASE_PATH = "mainnet";
```
karimodm: Do we actually need this configuration variable? This should be automatically be adjusted depending to the network you are using. This way we can also always rely on a standard location for local snapshot files.

src/main/java/com/iota/iri/conf/SnapshotConfig.java
```diff
@@ -42,6 +77,13 @@
 
     interface Descriptions {
 
+        String LOCAL_SNAPSHOTS_ENABLED = "Flag that determines if local snapshots are enabled.";
+        String LOCAL_SNAPSHOTS_PRUNING_ENABLED = "Flag that determines if pruning of old data is enabled.";
+        String LOCAL_SNAPSHOTS_PRUNING_DELAY = "Only prune data that precedes the local snapshot by n milestones.";
+        String LOCAL_SNAPSHOTS_INTERVAL_SYNCED = "Take local snapshots every n milestones if the node is fully synced.";
+        String LOCAL_SNAPSHOTS_INTERVAL_UNSYNCED = "Take local snapshots every n milestones if the node is syncing.";
+        String LOCAL_SNAPSHOTS_DEPTH = "Number of milestones to keep.";
```
karimodm: Is this the number of filestones to keep prior to latest local snapshot? Can you elaborate the description a bit?

src/main/java/com/iota/iri/conf/BaseIotaConfig.java
hmoog: **LOCAL_SNAPSHOTS_INTERVAL_SYNCED = 10**
Taking local snapshots regularly makes the processing faster since it doesnt have to examine so many new transactions so we want to do it relatively often but also not too often since it still consumes as few resources - 5 minutes have proven to be a good value while testing to keep the processing times short (below a few seconds) and still don't bother the node too many times with creating the snapshots.

**LOCAL_SNAPSHOTS_INTERVAL_UNSYNCED = 5000**
When we are syncing we usually process a lot of milestones very rapidly. To not bother the syncing process too much but still have a fast resume point when IRI crashes for example, we create a snapshot every 5000 milestones. In addition to reducing the amount of local snapshots taken and the corresponding load on the node, it also creates a situation where we have a bigger chunk of the old tangle data available most of the time because the transactions only get pruned when a snapshot was taken. This allows us to revert the state of the ledger further back in case we detect a milestone transaction that was skipped previously and continue the syncing without an unrecoverable error - repairing milestones will be included in a future PR.

**int LOCAL_SNAPSHOTS_DEPTH = 500;**
I reduced the snapshot depth back to 500 because it works 99.9% of the time (i only once faced a situation where data was referenced by the coo that was older than 500 milestones. Since we also have a pruning delay now (new feature) which keeps old transactions prior to the taken snapshot., this situation should not be a problem anymore. At the same time a lower snapshot depth allows it to sync new nodes extremely fast.

src/main/java/com/iota/iri/conf/BaseIotaConfig.java
karimodm: We already allow users to specify a data directory for their IRI nodes, I believe the local snapshots should be just dropped into that same folder; without giving the user this option, which I don't find the use for.

alon-e: the DB/persistable folder structure should be revisited, but as a different issue IMO.

src/main/java/com/iota/iri/conf/SnapshotConfig.java
karimodm: Yep, that's clear; just to be more verbose in the doc.

alon-e: I agree that this description should be more informative.
what are the ranges, what do they mean for the users?

src/main/java/com/iota/iri/conf/BaseIotaConfig.java
karimodm: > Since we also have a pruning delay now (new feature) which keeps old transactions prior to the taken snapshot.
So effectively you will be able to reference data up to `LOCAL_SNAPSHOTS_DEPTH + LOCAL_SNAPSHOTS_PRUNING_DELAY`, assuming pruning is enabled.

hmoog: yep

src/main/java/com/iota/iri/conf/BaseIotaConfig.java
hmoog: Can we create issues for minor things like this so i can change it afterwards? because once we have it merged i can more easily work on changes since i don't have to do them in 10 different branches.

src/main/java/com/iota/iri/conf/SnapshotConfig.java
hmoog: Maybe also create a separate issue after merging so we can get these changes in first - its kind of tricky to break the whole shit down anyway

src/main/java/com/iota/iri/conf/BaseIotaConfig.java
jakubcech: Yes, please, create an issue for this.

PR #982, opened 2018-09-11
Feat: added a generic helper class for dag traversal

hmoog:
Utility class for dag traversals used by the local snapshots logic


src/main/java/com/iota/iri/utils/dag/DAGUtils.java
```diff
@@ -0,0 +1,366 @@
...
+                                  TraversalCondition condition,
+                                  TraversalConsumer currentTransactionConsumer,
+                                  Set<Hash> processedTransactions) throws TraversalException {
+        final Queue<TransactionViewModel> transactionsToExamine = new LinkedList<>();
+        try {
+            startingTransaction.getApprovers(tangle).getHashes().stream().forEach(approverHash -> {
+                try {
+                    transactionsToExamine.add(TransactionViewModel.fromHash(tangle, approverHash));
+                } catch(Exception e) {
+                    throw new RuntimeException(e);
```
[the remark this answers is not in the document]
hmoog: This is intentionally since the lambda of forEach does not allow to throw unchecked exceptions - the exception will get caught by the outer try and wrapped accordingly

src/main/java/com/iota/iri/utils/dag/DAGUtils.java
```diff
@@ -0,0 +1,366 @@
...
+     * desired set of transactions and only then will be passed on to the currentTransactionConsumer lambda.
+     *
+     * @param startingTransaction the starting point of the traversal (does not get passed to the consumer)
+     * @param condition a lambda function that allows to control how long the traversal should continue
+     * @param currentTransactionConsumer a lambda function that allows us to "process" the found transactions
+     * @param processedTransactions a set of hashes that shall be considered as "processed" already and that will
+     *                              consequently be ignored in the traversal
+     * @throws TraversalException if anything goes wrong while traversing the graph and processing the transactions
+     */
+    public void traverseApprovers(TransactionViewModel startingTransaction,
```
karimodm: I hate to be the guy to tell this: but what about a unit tests for this method? :grimacing:

src/main/java/com/iota/iri/utils/dag/DAGUtils.java
```diff
@@ -0,0 +1,366 @@
...
+     *
+     * It uses an non-recursive iterative algorithm that is able to handle huge chunks of the tangle without running out
+     * of memory. It creates a queue of transactions that are being examined and processes them one by one. As new
+     * approvers are found, they will be added to the queue and processed accordingly.
+     *
+     * Every found transaction is passed into the provided condition lambda, to determine if it still belongs to the
+     * desired set of transactions and only then will be passed on to the currentTransactionConsumer lambda.
+     *
+     * @param startingTransaction the starting point of the traversal (does not get passed to the consumer)
+     * @param condition a lambda function that allows to control how long the traversal should continue
```
karimodm: Can you please add to the documentation that you pass the current traversed transaction to the condition lambda?

src/main/java/com/iota/iri/utils/dag/DAGUtils.java
```diff
@@ -0,0 +1,366 @@
...
+            while((currentTransaction = transactionsToExamine.poll()) != null) {
+                if(
+                processedTransactions.add(currentTransaction.getHash()) &&
+                currentTransaction.getType() != TransactionViewModel.PREFILLED_SLOT &&
+                condition.check(currentTransaction)
+                ) {
+                    currentTransactionConsumer.consume(currentTransaction);
+
+                    transactionsToExamine.add(TransactionViewModel.fromHash(tangle, currentTransaction.getBranchTransactionHash()));
+                    transactionsToExamine.add(TransactionViewModel.fromHash(tangle, currentTransaction.getTrunkTransactionHash()));
```
karimodm: This implementation is BFS, any chance we might need a DFS one since we are at it?

src/main/java/com/iota/iri/utils/dag/DAGUtils.java
```diff
@@ -0,0 +1,366 @@
...
+     * desired set of transactions and only then will be passed on to the currentTransactionConsumer lambda.
+     *
+     * @param startingTransaction the starting point of the traversal (does not get passed to the consumer)
+     * @param condition a lambda function that allows to control how long the traversal should continue
+     * @param currentTransactionConsumer a lambda function that allows us to "process" the found transactions
+     * @param processedTransactions a set of hashes that shall be considered as "processed" already and that will
+     *                              consequently be ignored in the traversal
+     * @throws TraversalException if anything goes wrong while traversing the graph and processing the transactions
+     */
+    public void traverseApprovees(TransactionViewModel startingTransaction,
```
karimodm: Unit tests?

src/main/java/com/iota/iri/utils/dag/DAGUtils.java
hmoog: Unit tests will be added once we have everything merged - a lot of the code parts contain tests that test complex behaviours which are hard to break up because they need each other.

To ceck if a snapshot can be rolled back i need the changed milestone tracker logic and so on.
So the unit tests will be added once we have merged the biggest stuff (before flipping the switch and activating the code).

src/main/java/com/iota/iri/utils/dag/DAGUtils.java
hmoog: We can add more functions when we need them - for now BFS is the "fairest" when it comes to traversing subtangles because it doesnt prioritize a single path (and the most used one in the code).

src/main/java/com/iota/iri/utils/dag/DAGUtils.java
hmoog: see other comment - unit tests come last (need to refactor the existing ones and write the missing ones)

src/main/java/com/iota/iri/utils/dag/DAGUtils.java
```diff
@@ -0,0 +1,366 @@
...
+
+/**
+ * This class offers generic functions for recurring tasks that are related to the tangle and that otherwise would have
+ * to be implemented over and over again in different parts of the code.
+ */
+public class DAGUtils {
+    /**
+     * Holds references to the singleton DAGUtils instances.
+     */
+    protected static HashMap<Object, DAGUtils> instances = new HashMap<>();
```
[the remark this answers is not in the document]
hmoog: It doesn't make a difference and i was thinking about adding additional parameters at some point but i can change that so its conistent for now

src/main/java/com/iota/iri/utils/dag/DAGUtils.java
hmoog: can do

src/main/java/com/iota/iri/utils/dag/DAGUtils.java
```diff
@@ -0,0 +1,366 @@
...
+     *                              consequently be ignored in the traversal
+     * @throws TraversalException if anything goes wrong while traversing the graph and processing the transactions
+     */
+    public void traverseApprovees(TransactionViewModel startingTransaction,
+                                  TraversalCondition condition,
+                                  TraversalConsumer currentTransactionConsumer,
+                                  Set<Hash> processedTransactions) throws TraversalException {
+        final Queue<TransactionViewModel> transactionsToExamine = new LinkedList<>();
+        try {
+            transactionsToExamine.add(TransactionViewModel.fromHash(tangle, startingTransaction.getBranchTransactionHash()));
```
alon-e: duplicate lines later on, maybe it would siplify the code (in both implementations here) if the Queue is initialized w/ the `startingTransaction` and the traversal only happens in the `while`?

src/main/java/com/iota/iri/utils/dag/DAGUtils.java
```diff
@@ -0,0 +1,366 @@
...
+            transactionsToExamine.add(TransactionViewModel.fromHash(tangle, startingTransaction.getTrunkTransactionHash()));
+
+            TransactionViewModel currentTransaction;
+            while((currentTransaction = transactionsToExamine.poll()) != null) {
+                if(
+                processedTransactions.add(currentTransaction.getHash()) &&
+                currentTransaction.getType() != TransactionViewModel.PREFILLED_SLOT &&
```
alon-e: side note: we should really add a new method to TxVM that eradicates this legacy `PREFILLED_SLOT`. will open an issue :).

src/main/java/com/iota/iri/utils/dag/DAGUtils.java
```diff
@@ -0,0 +1,366 @@
...
+     *                              consequently be ignored in the traversal
+     * @throws TraversalException if anything goes wrong while traversing the graph and processing the transactions
+     */
+    public void traverseApprovers(TransactionViewModel startingTransaction,
+                                  TraversalCondition condition,
+                                  TraversalConsumer currentTransactionConsumer,
+                                  Set<Hash> processedTransactions) throws TraversalException {
+        final Queue<TransactionViewModel> transactionsToExamine = new LinkedList<>();
+        try {
+            startingTransaction.getApprovers(tangle).getHashes().stream().forEach(approverHash -> {
```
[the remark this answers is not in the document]
hmoog: But getApprovers does exactly that?

```
    public ApproveeViewModel getApprovers(Tangle tangle) throws Exception {
        if(approovers == null) {
            approovers = ApproveeViewModel.load(tangle, hash);
        }
        return approovers;
    }
```

This comment would only make sense if we would not have the tvm's already but the hashes, because then we could skip fetching the TVM.

In fact since getApprovers caches the result i think the existing way tends to be faster than what you wrote.

src/main/java/com/iota/iri/utils/dag/DAGUtils.java
```diff
@@ -0,0 +1,366 @@
...
+     * @param currentTransactionConsumer a lambda function that allows us to "process" the found transactions
+     * @param processedTransactions a set of hashes that shall be considered as "processed" already and that will
+     *                              consequently be ignored in the traversal
+     * @throws TraversalException if anything goes wrong while traversing the graph and processing the transactions
+     */
+    public void traverseApprovers(TransactionViewModel startingTransaction,
+                                  TraversalCondition condition,
+                                  TraversalConsumer currentTransactionConsumer,
+                                  Set<Hash> processedTransactions) throws TraversalException {
+        final Queue<TransactionViewModel> transactionsToExamine = new LinkedList<>();
```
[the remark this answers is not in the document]
hmoog: Since we need to load the tvm for the traversal of the approvees (to get the branch and trunk tx hash) i wanted to have the functionality kind of the same for the other direction.

So you know how these utility methods work without having to think about it too much or even look it up. In addition its supposed to be a "utility method" that makes stuff easy and solveable in a few lines of code. Having to manually retrieve the transactions of interest kind of contradicts this.

Usually when you iterate through transactions you want to do sth with those tx's so it makes sense to get them right away)

src/main/java/com/iota/iri/utils/dag/DAGUtils.java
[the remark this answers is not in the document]
karimodm: It is for consistency and readability.

src/main/java/com/iota/iri/utils/dag/DAGUtils.java
```diff
@@ -0,0 +1,366 @@
...
+     * name).
+     *
+     * It uses an non-recursive iterative algorithm that is able to handle huge chunks of the tangle without running out
+     * of memory. It creates a queue of transactions that are being examined and processes them one by one. As new
+     * approvers are found, they will be added to the queue and processed accordingly.
+     *
+     * Every found transaction is passed into the provided condition lambda, to determine if it still belongs to the
+     * desired set of transactions and only then will be passed on to the currentTransactionConsumer lambda.
+     *
+     * @param startingTransaction the starting point of the traversal (does not get passed to the consumer)
```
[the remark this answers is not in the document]
hmoog: we can add that later - shouldnt block this pr for now

src/main/java/com/iota/iri/utils/dag/DAGUtils.java
```diff
@@ -0,0 +1,366 @@
...
+     * @param condition a lambda function that allows to control how long the traversal should continue
+     * @param currentTransactionConsumer a lambda function that allows us to "process" the found transactions
+     * @param processedTransactions a set of hashes that shall be considered as "processed" already and that will
```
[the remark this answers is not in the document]
hmoog: When repairing the ledger (when milestones were processed in the wrong order or a milestone was skipped) you reset the snapshotIndex of all transactions that were referenced by this missing milestone. While traversing its transaction you can face transactions that were referenced by a future milestone that consequently also has to be reset (since the missing milestone didnt mark the child transactions correctly and the future milestones claimed too many tx).

When you reset the future milestones you do not want to reset the exact same transactions that where reset by the milestone that you repaired first (again), so you can share a processedTransactions set between all those potentially huge amount of calls an never process tx twice or in the worst case even more than that. 

You will see it in a future PR

src/main/java/com/iota/iri/utils/dag/DAGUtils.java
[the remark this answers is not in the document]
GalRogozinski: 1. I agree with @karimodm, please change to `Tangle`. `Object` should be used as last resort imo. It gives lots of flexibility on one hand, but as the code grows too much flexibility can cause confusing design which leads to waste of time and bugs.
2. Since this is a static object which we only initialize in the constructor I would also add the `final` word to it. I don't remember if I ranted to you how much I usually hate the overuse of `final` in the code, but here it should be written down because: 

a. It is a general convention when creating static fields in java
b. It ensures you don't change the reference later on (you are dealing with global scope here)
c. It provides gurantees about references not escaping in multithreaded environments.(https://wiki.sei.cmu.edu/confluence/display/java/OBJ10-J.+Do+not+use+public+static+nonfinal+fields).

Now you may claim that the reasons I enlisted here may less affect the current code. However I think that sticking to conventions usually help preventing future bugs created by other programmers that come after you. I also need subconcientiously see it.


I will also rant about `protected`. Unless you need this inherited I will keep it as private. Here I don't have much good arguements about why not just make everything protected so inherited class will always have access. I just think it a good principle to expose only what you need.

src/main/java/com/iota/iri/utils/dag/DAGUtils.java
```diff
@@ -0,0 +1,194 @@
...
+     *
+     * To save memory and to be able to use things like caches effectively, we do not allow to create multiple instances
+     * of this class for the same set of parameters, but create a singleton instance of it.
+     *
+     * @param tangle database interface used to retrieve data
+     * @return DAGUtils instance that allows to use the implemented methods.
+     */
+    public static DAGUtils get(Tangle tangle) {
+        DAGUtils instance = instances.get(tangle);
+        if(instance == null) {
```
GalRogozinski: now notice that this isn't thread safe. Two different threads will call `get` the same time and each will return a different `DAGUtils` object.

I think that under your current code this will not cause any problems (there will just be a waste). However in the future some smart dev may come along and will add another field to DAGUtils…
We want to create a design that won't do anything surprising.

There are 2 ways to circumvent this:

1. Write safe thread code using `synchronized`. You can also look at [double-check locking](https://en.wikipedia.org/wiki/Double-checked_locking#Usage_in_Java), but this is ugly,
2. (my preferred way) - I usually think of a utils class as a collection of static methods. If you just have static methods that accept `tangle` as a parameter you will gain the following benefits:
  - None of this thread safety bs
  - Static will run faster than instance methods
  - You wont encourage other devs to add fields to this class (instances are good for OOP, but for utils I think this way is better)

src/main/java/com/iota/iri/utils/dag/DAGUtils.java
```diff
@@ -0,0 +1,194 @@
...
+
+    /**
+     * Constructor of the class which allows to provide the required dependencies for the implemented methods.
+     *
+     * Since the constructor is protected we are not able to manually create instances of this class and have to
+     * retrieve them through their singleton accessor {@link #get}.
+     *
+     * @param tangle database interface used to retrieve data
+     */
+    protected DAGUtils(Tangle tangle) {
```
GalRogozinski: private

src/main/java/com/iota/iri/utils/dag/DAGUtils.java
```diff
@@ -0,0 +1,194 @@
...
+     * @param currentTransactionConsumer a lambda function that allows us to "process" the found transactions
+     * @param processedTransactions a set of hashes that shall be considered as "processed" already and that will
+     *                              consequently be ignored in the traversal
+     * @throws TraversalException if anything goes wrong while traversing the graph and processing the transactions
+     */
+    public void traverseApprovers(Hash startingTransactionHash,
+                                  TraversalCondition condition,
+                                  TraversalConsumer currentTransactionConsumer,
+                                  Set<Hash> processedTransactions) throws TraversalException {
+        final Queue<Hash> transactionsToExamine = new LinkedList<>(Collections.singleton(startingTransactionHash));
```
GalRogozinski: In the local scope I hate `final`. There is nothing really bad about it except that it adds verbosity and java 8 has the "effectively final" concept.


Another thing is that using an Array based implementation (ArrayDeque) will be faster (even though LinkedList seems more logical here). The reason is that you do less memory allocations.

You can run a benchmark if you don't believe me :-)

src/main/java/com/iota/iri/utils/dag/TraversalCondition.java
```diff
@@ -0,0 +1,12 @@
+package com.iota.iri.utils.dag;
+
+import com.iota.iri.controllers.TransactionViewModel;
+
+/**
+ * Functional interface for the lambda function getting passed into the DAGUtils traverse method to check if the found
+ * transaction is relevant for our processing.
+ */
+@FunctionalInterface
+public interface TraversalCondition {
```
GalRogozinski: I personally would have been fine if you would have used java's built in `Predicate` and named the variable traversalCondition (especially since you add such good javadocs).


However, I have no problem whatsoever with this code 👍 

Just telling you for the future if you want to save time :-)

src/main/java/com/iota/iri/utils/dag/TraversalConsumer.java
```diff
@@ -0,0 +1,12 @@
+package com.iota.iri.utils.dag;
+
+import com.iota.iri.controllers.TransactionViewModel;
+
+/**
+ * Functional interface for the lambda function getting passed into the DAGUtils traverse method to process the found
+ * transactions.
+ */
+@FunctionalInterface
+public interface TraversalConsumer {
```
GalRogozinski: There is also a `Consumer` built in.
Once again I am totally fine with the code.

src/main/java/com/iota/iri/utils/dag/TraversalException.java
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * It allows us to distinct between the different kinds of errors that can happen during the execution of the code.
+ */
+public class TraversalException extends Exception {
+    /**
+     * Constructor of the exception which allows us to provide a specific error message and the cause of the error.
+     *
+     * @param message reason why this error occured
+     * @param cause wrapped exception that caused this error
+     */
```
GalRogozinski: Wow such detailed javadocs :-)

I would have been lazy and just used the `@see` but I will not complain :-)

src/main/java/com/iota/iri/utils/dag/DAGUtils.java
hmoog: protected to be able to still create manual utils objects for testing (test are in the same packet).

src/main/java/com/iota/iri/utils/dag/DAGHelper.java
```diff
@@ -0,0 +1,210 @@
...
+
+/**
+ * This class offers generic functions for recurring tasks that are related to the tangle and that otherwise would have
+ * to be implemented over and over again in different parts of the code.
+ */
+public class DAGHelper {
+    /**
+     * Holds references to the singleton DAGHelper instances.
+     */
+    protected static HashMap<Tangle, DAGHelper> instances = new HashMap<>();
```
GalRogozinski: make `static final` please

PR #983, opened 2018-09-12
Feat: renamed Milestone to MilestoneTracker

hmoog:
# Description

Renamed Milestone class to MilestoneTracker to better reflect its functionality and to ease the review of future PR's since Local Snapshots uses MilestoneTracker in its codebase.

Should be reviewed an approved as fast as possible to make future PR's easier to check (otherwise it doesnt show the diffs but all changes as a new file).

### DIFF
--- a/src/test/java/com/iota/iri/service/tipselection/impl/WalkValidatorImplTest.java
+++ b/src/test/java/com/iota/iri/service/tipselection/impl/WalkValidatorImplTest.java
@@ -1,7 +1,7 @@
 package com.iota.iri.service.tipselection.impl;
 
 import com.iota.iri.LedgerValidator;
-import com.iota.iri.Milestone;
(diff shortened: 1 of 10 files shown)


src/test/java/com/iota/iri/service/tipselection/impl/WalkValidatorImplTest.java
```diff
@@ -36,7 +36,7 @@
     @Mock
     private TransactionValidator transactionValidator;
     @Mock
-    private Milestone milestoneTracker;
+    private MilestoneTracker milestoneTrackerTracker;
```
antalm: I believe this is a search/replace error. This private variable was already called milestoneTracker before the modification, and by replacing all milestone strings to milestoneTracker, now it is called milestoneTrackerTracker by mistake.

PR #988, opened 2018-09-16
add pmd ruleset

GalRogozinski:
# Description
Add rules I have chosen for PMD.
PMD is ran by Codacy and can be run by the developer via an IDE plugin or PMD cli.
None of those rules will break the build. It will simply enforce coding standards that I think are important.
When you receive a Codacy comment you disagree on, please comment and explain the problem.
There may be changes in the future where we add/omit rules.
If you think some rules need to be added/removed then please comment.

### DIFF
--- a/checkstyle.xml
+++ b/checkstyle.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0"?>
 <!DOCTYPE module PUBLIC
-          "-//Puppy Crawl//DTD Check Configuration 1.3//EN"
-          "http://checkstyle.sourceforge.net/dtds/configuration_1_3.dtd">
+          "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
+          "https://checkstyle.org/dtds/configuration_1_3.dtd">
 <!--
 
 IOTA checkstyle draft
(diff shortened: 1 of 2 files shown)


legacycode:
Hi,
there is also a checkstyle plugin in the pom.xml that is not configured correctly. I have fixed this in a local branch. There is only one rule specified in the checkstyle.xml. Does it make sense to use checkstyle in addition to pmd or should we remove the checkstyle plugin from mavens pom?
Best regards.
legacycode

GalRogozinski:
There should already be more than one rule on the `dev` branch. See #753.
The idea is that checkstyle rule should be more heavily enforced. Currently they fail the `mvn validate` goal. Each addition of a rule will constitute changes in IRI code. I want to add more rules to it once every few weeks.

PMD won't fail maven on the other hand. This is because I believe its rules are more controversial.
For example, it shouts when you hardcode ip addresses (which is generally a good thing). However if you hard code `0.0.0.0` is it really that bad? 
Anyhows, PMD allows me to insert lots of rules quickly so that future code will be of better quality.
When we will get to refactoring existing code it will be used to help us as well.

legacycode:
That is a good idea. I created a RP #990 and optimized the checkstyle configuration. Now the rules from checkstyle.xml will also be available by running "mvn checkstyle:check" and "checkstyle:checkstyle".

PR #1001, opened 2018-09-18
 Feat: Introducing a dedicated MilestoneSolidifier for Local Snapshots

hmoog:
This PR introduces the MilestoneSolidifier used in the Local Snapshots package.

It helps to separate the solidification of milestones from their application to the ledger and allows us to NOT skip missing milestones in their validation and application to the ledger state while at the same time ALLOWING TO SKIP milestones during their solidification (while trying to discover previous milestones).

In addition we can keep track of the earliest milestone (by index) rather than db insertion order and solidify the milestones much faster (by discovering the earliest unsolid one as fast as possible).

The usage of this class will be introduced in a later PR.


src/main/java/com/iota/iri/service/milestone/MilestoneSolidifier.java
```diff
@@ -0,0 +1,247 @@
...
+     * This method checks if the current unsolid milestone has become solid or irrelevant and advances to the next one
+     * if that is the case.
+     *
+     * It is getting called by the solidification thread in regular intervals.
+     *
+     * @return true if the current solidification task was successful and the next milestone is due or false otherwise
+     */
+    private boolean processSolidificationTask() {
+        if (earliestMilestoneHash != null && (
+            earliestMilestoneIndex <= snapshotManager.getInitialSnapshot().getIndex() ||
```
karimodm: How could this ever happen if you only add to the `unsolidMilestones` when:

```
   if (milestoneIndex > snapshotManager.getInitialSnapshot().getIndex()) {
            if (milestoneIndex < earliestMilestoneIndex) {
                earliestMilestoneHash = milestoneHash;
                earliestMilestoneIndex = milestoneIndex;
                earliestMilestoneTries = 0;
            }

            unsolidMilestones.put(milestoneHash, milestoneIndex);
        }
```

src/main/java/com/iota/iri/service/milestone/MilestoneSolidifier.java
```diff
@@ -0,0 +1,247 @@
...
+
+        if (unsolidMilestones.size() > 1) {
+            statusLogger.status("Solidifying milestone #" + earliestMilestoneIndex + " [" + unsolidMilestones.size() + " left]");
+        }
+
+        try {
+            return transactionValidator.checkSolidity(
+                earliestMilestoneHash,
+                true,
+                2 ^ Math.min(
```
karimodm: XOR operator? I don't get this entire expression.

src/main/java/com/iota/iri/utils/log/StatusLogger.java
```diff
@@ -0,0 +1,133 @@
...
+
+    /**
+     * This method handles the actual output of messages through the {@link Logger} instance.
+     *
+     * It first checks if the message that shall get printed differs from the last message that was printed and then
+     * issues the output of the message. After printing the message, it updates the internal variables to handle the
+     * next message accordingly.
+     */
+    private void printStatusMessage() {
+        if (!statusMessage.equals(lastPrintedStatusMessage)) {
```
karimodm: You are already checking in the `status` method if the message you just received was equal to the previous one.

src/main/java/com/iota/iri/utils/log/StatusLogger.java
```diff
@@ -0,0 +1,133 @@
...
+     *
+     * @param message status message that shall get printed
+     */
+    public void status(String message) {
+        if (!message.equals(statusMessage)) {
+            statusMessage = message;
+
+            if (System.currentTimeMillis() - lastLogTime >= LOG_INTERVAL) {
+                printStatusMessage();
+            } else if (outputThread == null) {
```
karimodm: This could lead to some messages going to /dev/null: if the `outputThread` is already sleeping to deliver a message, all messages in the meanwhile would get lost. If you change for `else` then this method will synchronize calls (assuming status messages are different), making this logger a bottleneck on `LOG_INTERVAL`.

src/main/java/com/iota/iri/TransactionValidator.java
```diff
@@ -126,16 +126,56 @@ public TransactionViewModel validateBytes(final byte[] bytes, int minWeightMagni
...
+     * @return true if the transaction is solid and false otherwise
+     * @throws Exception if anything goes wrong while trying to solidify the transaction
+     */
+    public boolean checkSolidity(Hash hash, boolean milestone, int maxProcessedTransactions) throws Exception {
         if(TransactionViewModel.fromHash(tangle, hash).isSolid()) {
             return true;
         }
         Set<Hash> analyzedHashes = new HashSet<>(Collections.singleton(Hash.NULL_HASH));
+        if(maxProcessedTransactions != Integer.MAX_VALUE) {
+            maxProcessedTransactions += analyzedHashes.size();
```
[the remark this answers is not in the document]
hmoog: In local snapshots there will be more solid entrypoints than the null_hash (dynamic amount)

src/main/java/com/iota/iri/service/milestone/MilestoneSolidifier.java
```diff
@@ -0,0 +1,247 @@
...
+     *
+     * To allow for faster processing it only "waits" for another check if the current solidification task was not
+     * finished (otherwise it immediately "continues" with the next one).
+     */
+    public void start() {
+        running = true;
+
+        new Thread(() -> {
+            while(running) {
+                if (processSolidificationTask() && running) {
```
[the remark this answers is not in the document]
hmoog: You are right - doesnt make much sense before i had some long running method in the if clause that i wanted to prevent from getting executed if the thread was terminated while the solidification took place

src/main/java/com/iota/iri/service/milestone/MilestoneSolidifier.java
```diff
@@ -0,0 +1,247 @@
...
+     * This method returns the earliest seen Milestone from the internal Map of unsolid milestones.
+     *
+     * If no unsolid milestone was found it returns an Entry with the Hash being null and the index being
+     * Integer.MAX_VALUE.
+     *
+     * @return the Map.Entry holding the earliest milestone or a default Map.Entry(null, Integer.MAX_VALUE)
+     */
+    private Map.Entry<Hash, Integer> getEarliestUnsolidMilestoneEntry() {
+        try {
+            return Collections.min(unsolidMilestones.entrySet(), comparingDouble(Map.Entry::getValue));
```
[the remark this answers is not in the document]
hmoog: Ohhh you are right that doesnt make much sense :P

src/main/java/com/iota/iri/service/milestone/MilestoneSolidifier.java
```diff
@@ -0,0 +1,247 @@
...
+     * from the oldest milestone to the newest one and not "block" the solidification with a very recent milestone that
+     * needs to traverse huge chunks of the tangle. If we fail to solidify a milestone for a certain amount of tries, we
+     * increase the amount of transactions that may be processed by using an exponential binary backoff strategy. The
+     * main goal of this is to give the solidification enough "resources" to discover the previous milestone (if it ever
+     * gets stuck because of a very long path to the previous milestone) while at the same time allowing fast solidity
+     * checks in "normal conditions".
+     *
+     * @return true if there are no unsolid milestones that have to be processed or if the earliest milestone is solid
+     */
+    private boolean earliestMilestoneIsSolid() {
```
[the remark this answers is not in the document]
hmoog: i will adjust this

src/main/java/com/iota/iri/service/milestone/MilestoneSolidifier.java
hmoog: Since we have 3 threads that solidify transaction, a milestone could become solid outside of this thread. If this happens and a snapshot gets taken it could start pruning the exact same milestone which would make this thread start to solidify it again (and with it the whole tangle) that just started to get pruned.

While this is a very rare edge case it could theoretically happen (and i think i saw it happening once) so i only try to solidify milestones that are still ahead of the cut off point of the last local snapshot (just to be safe).

Note: Traversing the graph can take a really long time if the missing tx is "behind" a huge chunk of known transactions. This will however be solved once we add the bottom->top solidification anyway. But for now its required.

src/main/java/com/iota/iri/utils/log/StatusLogger.java
hmoog: Imagine the followign example:

.status("A") -> gets printed
.status("B") -> gets scheduled
.status("A") -> would get printed again even tho we already showed this exact status.

src/main/java/com/iota/iri/utils/log/StatusLogger.java
hmoog: Thats the whole point of this class - to not show "all" messages but to just dump a message every 3 seconds with the "latest" status.

Since we store the next status message that shall be printed a few lines before in:

statusMessage = message;

The output thread will print this last message and drop the ones that happened in between.

src/main/java/com/iota/iri/service/milestone/MilestoneSolidifier.java
karimodm: In Java the `^` operator is not power of, it is binary XOR! You have to use `Math.pow`.

src/main/java/com/iota/iri/utils/log/StatusLogger.java
karimodm: Oh I see: it's last received vs last printed.

src/main/java/com/iota/iri/utils/log/StatusLogger.java
karimodm: > Thats the whole point of this class - to not show "all" messages but to just dump a message every 3 seconds with the "latest" status.

I thought the purpose of the class if how you described in the JavaDoc:

> It checks if the given message is new and then either prints the message immediately (if enough time has passed since the last message)

src/main/java/com/iota/iri/service/milestone/MilestoneSolidifier.java
```diff
@@ -0,0 +1,247 @@
...
+
+        if (unsolidMilestones.size() > 1) {
+            statusLogger.status("Solidifying milestone #" + earliestUnsolidMilestoneIndex + " [" + unsolidMilestones.size() + " left]");
+        }
+
+        try {
+            return transactionValidator.checkSolidity(
+            earliestUnsolidMilestoneHash,
+                true,
+                2 ^ Math.min(
```
karimodm: In Java the `^` operator is not power of, it is binary XOR! You have to use `Math.pow`.

src/main/java/com/iota/iri/utils/log/StatusLogger.java
karimodm: > Thats the whole point of this class - to not show "all" messages but to just dump a message every 3 seconds with the "latest" status.

I've understood something different from the JavaDoc:

> It checks if the given message is new and then either prints the message immediately (if enough time has passed since the last message)

but then again I don't see the point of this if:
`if (!message.equals(statusMessage)) {`
if you always disregard the previous statusMessage anyway, either equal or not.

hmoog: If the "new" message is the same as the current status message it was either printed already or is scheduled to be printed, so we can ignore it.

Since with this check only, the message can change to sth else and then back to what it was before we have an additional check in the print method to not print the same status update twice.

src/main/java/com/iota/iri/utils/thread/ThreadUtils.java
```diff
@@ -0,0 +1,139 @@
...
+     * method multiple times.
+     *
+     * @param threadIdentifier the identifier of the thread that we want to stop
+     * @return the {@link Thread} that will be terminated or null if there was no {@link Thread} associated to the given
+     *         identifier
+     */
+    public static Thread stopThread(ThreadIdentifier threadIdentifier) {
+        if (threads.get(threadIdentifier) != null && !threads.get(threadIdentifier).isInterrupted()) {
+            synchronized(threadIdentifier) {
+                if (threads.get(threadIdentifier) != null && !threads.get(threadIdentifier).isInterrupted()) {
```
karimodm: Same here.

src/main/java/com/iota/iri/utils/thread/ThreadUtils.java
```diff
@@ -0,0 +1,139 @@
...
+     * multiple times.
+     *
+     * @param runnable method reference that is responsible for processing the thread (i.e. {@code this::myThreadMethod})
+     * @param threadIdentifier identifier object that is used to associate the {@link Thread} and synchronize the access
+     * @return the thread that got spawned by this method
+     */
+    public static Thread spawnThread(Runnable runnable, ThreadIdentifier threadIdentifier) {
+        if (threads.get(threadIdentifier) == null || threads.get(threadIdentifier).isInterrupted()) {
+            synchronized(threadIdentifier) {
+                if (threads.get(threadIdentifier) == null || threads.get(threadIdentifier).isInterrupted()) {
```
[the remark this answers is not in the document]
hmoog: I don't think that using a ConcurrentHashMap will be sufficient because you are also writing to the map after which also have to be synchronized. So you want to synchronize those 2 calls rather than just have the get synchronized

src/main/java/com/iota/iri/utils/thread/ThreadUtils.java
hmoog: I don't think that using a ConcurrentHashMap will be sufficient because you are also writing to the map after which also have to be synchronized. So you want to synchronize those 2 calls rather than just have the get synchronized.

And in addition the stopping of the thread also have to be synchronized over the same object. Otherwise the following could happen.

You call spawnThread and immediately after stopThread:

1. spawnThread determines there is no running job
2. stopThread determines there is no running job and terminates
3. spawnThread creates a job

With the current implementation synchronizing both calls over the same object:

1. spawnThread determines there is no running job
2. spawnThread creates a job
3. stopThread terminates it

src/main/java/com/iota/iri/network/TransactionRequester.java
```diff
@@ -76,6 +76,32 @@ public void requestTransaction(Hash hash, boolean milestone) throws Exception {
...
+     *
+     * @param transactionHash hash of the transaction to check
+     * @return true if the transaction is in the set of transactions to be requested and false otherwise
+     */
+    public boolean contains(Hash transactionHash) {
+        return milestoneTransactionsToRequest.contains(transactionHash) || transactionsToRequest.contains(transactionHash);
+    }
+
+    /**
+     * This method allows to check if a transaction was requested by the TransactionRequester.
```
karimodm: Change doc to reflect that you are looking into Milestones being requested already.

src/main/java/com/iota/iri/TransactionValidator.java
```diff
@@ -179,14 +179,18 @@ public boolean checkSolidity(Hash hash, boolean milestone, int maxProcessedTrans
...
-                        break;
-                    } else {
-                        if (solid) {
-                            nonAnalyzedTransactions.offer(transaction.getTrunkTransactionHash());
-                            nonAnalyzedTransactions.offer(transaction.getBranchTransactionHash());
+
+                        if (milestone && !transactionRequester.containsMilestoneRequest(hashPointer)) {
+                            transactionRequester.requestTransaction(hashPointer, milestone);
+                            break;
+                        } else if (!milestone && !transactionRequester.contains(hashPointer)) {
```
karimodm: What's the point of these extra ifs if `requestTransaction` already does:
```
                if(milestone) {
                    transactionsToRequest.remove(hash);
                    milestoneTransactionsToRequest.add(hash);
```

src/main/java/com/iota/iri/service/milestone/MilestoneSolidifier.java
```diff
@@ -146,118 +145,191 @@ public void shutdown() {
...
+
+            if (youngestMilestoneInQueue == null || milestoneEntry.getValue() > youngestMilestoneInQueue.getValue()) {
+                youngestMilestoneInQueue = milestoneEntry;
             }
+        } else if (milestoneEntry.getValue() < youngestMilestoneInQueue.getValue()) {
+            milestonesToSolidify.remove(youngestMilestoneInQueue.getKey());
+            milestonesToSolidify.put(milestoneEntry.getKey(), milestoneEntry.getValue());
 
-            ThreadUtils.sleep(SOLIDIFICATION_INTERVAL);
+            youngestMilestoneInQueue = null;
```
karimodm: You already have `youngestMilestoneInQueue = null;` as part of `determineYoungestMilestoneInQueue`

src/main/java/com/iota/iri/service/milestone/MilestoneSolidifier.java
```diff
@@ -146,118 +145,191 @@ public void shutdown() {
...
      *
-     * We limit the amount of transactions that may be processed during the solidity check, since we want to solidify
-     * from the oldest milestone to the newest one and not "block" the solidification with a very recent milestone that
-     * needs to traverse huge chunks of the tangle. If we fail to solidify a milestone for a certain amount of tries, we
-     * increase the amount of transactions that may be processed by using an exponential binary backoff strategy. The
-     * main goal of this is to give the solidification enough "resources" to discover the previous milestone (if it ever
-     * gets stuck because of a very long path to the previous milestone) while at the same time allowing fast solidity
-     * checks in "normal conditions".
+     * It iterates through the queue and checks if the corresponding milestones are still relevant for our node, or if
+     * they could be successfully solidified. If the milestones become solid or irrelevant, we remove them from the
```
karimodm: Just make more clear in the doc that that this is actually the call that will attempt to solidify the queue by eventually issuing `requestTransaction` calls from `isSolid`.

src/main/java/com/iota/iri/TransactionValidator.java
hmoog: The point of these changes is not to "upgrade" a request to a milestone request (which is what the transactionRequester does in your mentioned code part), but to ONLY "BREAK" if we didnt request the missing transaction as part of the correct set yet.

So in case we call checkSolidity multiple times without having received the answer to our previous request yet, we will queue additional transaction requests for additional transactions that are ALSO missing. Before we always stopped at the same transaction requesting it again and again and again and only progressing to further transactions when we received an answer.

This little tweak is what made the sync spead go up 10 fold.

src/main/java/com/iota/iri/service/milestone/MilestoneSolidifier.java
hmoog: true

src/main/java/com/iota/iri/TransactionValidator.java
karimodm: :100:

src/main/java/com/iota/iri/utils/log/StatusLogger.java
```diff
@@ -0,0 +1,133 @@
...
+                    if (outputThread == null) {
+                        outputThread = new Thread(() -> {
+                            try {
+                                Thread.sleep(Math.max(LOG_INTERVAL - (System.currentTimeMillis() - lastLogTime), 1));
+                            } catch(InterruptedException e) { /* do nothing */ }
+
+                            printStatusMessage();
+
+                            outputThread = null;
+                        }, Thread.currentThread().getName());
```
GalRogozinski: So you give the `outputThread` the same name of the thread that happened to call the `status` method??

src/main/java/com/iota/iri/utils/log/StatusLogger.java
```diff
@@ -0,0 +1,133 @@
...
+
+            if (System.currentTimeMillis() - lastLogTime >= LOG_INTERVAL) {
+                printStatusMessage();
+            } else if (outputThread == null) {
+                synchronized(this) {
+                    if (outputThread == null) {
+                        outputThread = new Thread(() -> {
+                            try {
+                                Thread.sleep(Math.max(LOG_INTERVAL - (System.currentTimeMillis() - lastLogTime), 1));
+                            } catch(InterruptedException e) { /* do nothing */ }
```
GalRogozinski: why do nothing?

src/main/java/com/iota/iri/utils/log/StatusLogger.java
```diff
@@ -0,0 +1,133 @@
...
+import org.slf4j.Logger;
+
+/**
+ * This class represents a wrapper for the {@link Logger} used by IRI that implements a logic to regularly issue status
+ * updates about a certain task.
+ *
+ * Instead of printing all messages immediately and unnecessarily spamming the console output, it rate limits the amount
+ * of messages shown and only prints messages if the status has changed since the last output.
+ */
+public class StatusLogger {
```
GalRogozinski: Hmm interesting idea :-)

src/main/java/com/iota/iri/utils/log/StatusLogger.java
GalRogozinski: Anyhows, just for readability i would wrap the logic of creating the thread in a private method.

src/main/java/com/iota/iri/utils/thread/ThreadUtils.java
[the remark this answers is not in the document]
GalRogozinski: As I spoke to you in private I prefer that you use the `SingleThreadExecutor` or `ScheduledSingleThreadExecutor` (which can replace this class), Instead of doing the global scope solution.
This is because:
1. global scope is evil
2. I don't think that an "infrastructure" `spawnThread` method should have a `synchronized` in it. Especially when it is easy enough to do a solution that doesn't necessitate this.

Sorry for not telling you about the `SingleThreadExecutor` before. When I said you can wrap the logic that creates thread I didn't think you would do this global scope solution to manage threads.

src/main/java/com/iota/iri/service/milestone/MilestoneSolidifier.java
```diff
@@ -0,0 +1,339 @@
...
+     * Note: While checking the solidity of the milestones we issue
+     *       {@link TransactionValidator#checkSolidity(Hash, boolean, int)} calls using the {@link #isSolid(Map.Entry)}
+     *       method, which requests the missing transactions and marks the milestones as solid once all referenced
+     *       transactions are known.
+     */
+    private void processSolidificationQueue() {
+        for (
+            Iterator<Map.Entry<Hash, Integer>> iterator = milestonesToSolidify.entrySet().iterator();
+            iterator.hasNext();
+        ) {
```
GalRogozinski: Heads Up,
I will change checkstyle later on to enforce 
```
for ( ...
          ....) {
```

src/main/java/com/iota/iri/service/milestone/MilestoneSolidifier.java
```diff
@@ -0,0 +1,339 @@
...
+    private void refillSolidificationQueue() {
+        if(youngestMilestoneInQueue == null && milestonesToSolidify.size() >= 1) {
+            determineYoungestMilestoneInQueue();
+        }
+
+        Map.Entry<Hash, Integer> nextSolidificationCandidate;
+        while (
+            milestonesToSolidify.size() < SOLIDIFICATION_QUEUE_SIZE &&
+            (nextSolidificationCandidate = getNextSolidificationCandidate()) != null
+        ) {
```
GalRogozinski: Heads Up,
I will change checkstyle later on to enforce 
```
for ( ...
          ....) {
```

src/main/java/com/iota/iri/service/milestone/MilestoneSolidifier.java
```diff
@@ -0,0 +1,339 @@
...
+     *
+     * We process them lazy to decrease the synchronization requirements and speed up the addition of milestones from
+     * outside {@link Thread}s.
+     *
+     * It simply iterates over the milestones and adds them to the pool. If they are older than the
+     * {@link #youngestMilestoneInQueue}, we add the to the solidification queue.
+     */
+    private void processNewlyAddedMilestones() {
+        for (
+            Iterator<Map.Entry<Hash, Integer>> iterator = newlyAddedMilestones.entrySet().iterator();
```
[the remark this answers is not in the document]
hmoog: No it cannot occur because we use a ConcurrentHashMap

src/main/java/com/iota/iri/service/milestone/MilestoneSolidifier.java
```diff
@@ -0,0 +1,339 @@
...
+     * pool and the queue and reset the {@link #youngestMilestoneInQueue} marker (if necessary).
+     *
+     * Note: While checking the solidity of the milestones we issue
+     *       {@link TransactionValidator#checkSolidity(Hash, boolean, int)} calls using the {@link #isSolid(Map.Entry)}
+     *       method, which requests the missing transactions and marks the milestones as solid once all referenced
+     *       transactions are known.
+     */
+    private void processSolidificationQueue() {
+        for (
+            Iterator<Map.Entry<Hash, Integer>> iterator = milestonesToSolidify.entrySet().iterator();
```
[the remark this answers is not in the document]
hmoog: The processSolidificationQueue() method is only called from the "worker thread" - there is no concurrent access.

src/main/java/com/iota/iri/service/milestone/MilestoneSolidifier.java
```diff
@@ -0,0 +1,339 @@
...
+        ) {
+            Map.Entry<Hash, Integer> currentEntry = iterator.next();
+
+            unsolidMilestonesPool.put(currentEntry.getKey(), currentEntry.getValue());
+
+            if (youngestMilestoneInQueue == null || currentEntry.getValue() < youngestMilestoneInQueue.getValue()) {
+                addToSolidificationQueue(currentEntry);
+            }
+
+            iterator.remove();
```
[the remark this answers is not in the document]
hmoog: no clone required since its a ConcurrentHashMap (see other comment)

src/main/java/com/iota/iri/service/milestone/MilestoneSolidifier.java
```diff
@@ -0,0 +1,339 @@
...
+     * This method takes an entry from the {@link #unsolidMilestonesPool} and adds it to the
+     * {@link #milestonesToSolidify} queue.
+     *
+     * It first checks if the given milestone is already part of the queue and then tries to add it. If the queue is not
+     * full yet, the addition to the queue is relatively cheap, because the {@link #youngestMilestoneInQueue} marker can
+     * be updated without iterating over all entries. If the queue reached its capacity already, we replace the entry
+     * marked by the {@link #youngestMilestoneInQueue} marker and update the marker by recalculating it using
+     * {@link #determineYoungestMilestoneInQueue()}.
+     *
+     * @param milestoneEntry entry from the {@link #unsolidMilestonesPool} that shall get added to the queue
```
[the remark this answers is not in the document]
hmoog: its explained in the unsolidMilestonesPool docs

src/main/java/com/iota/iri/utils/log/interval/IntervalLogger.java
```diff
@@ -0,0 +1,443 @@
...
+                lastPrintedMessage = this;
+                lastLogTime = System.currentTimeMillis();
+            }
+        }
+
+        /**
+         * {@inheritDoc}
+         */
+        @Override
+        public boolean equals(Object obj) {
```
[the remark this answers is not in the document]
hmoog: was changed already

src/main/java/com/iota/iri/utils/log/interval/IntervalLogger.java
```diff
@@ -0,0 +1,443 @@
...
+
+        /**
+         * This method allows us to compare different {@link Message}s for equality and check if we have received or
+         * processed the same {@link Message} already.
+         *
+         * @param obj object that shall be compared to this instance
+         * @return true if it represents the same message or false otherwise
+         */
+        @Override
+        public abstract boolean equals(Object obj);
```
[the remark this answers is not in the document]
hmoog: was changed already

src/main/java/com/iota/iri/service/milestone/MilestoneSolidifier.java
```diff
@@ -0,0 +1,339 @@
...
+
+    /**
+     * This method determines the youngest milestone in the solidification queue.
+     *
+     * It iterates over all milestones in the Queue and keeps track of the youngest one found (the one with the highest
+     * milestone index).
+     */
+    private void determineYoungestMilestoneInQueue() {
+        youngestMilestoneInQueue = null;
+        for (Map.Entry<Hash, Integer> currentEntry : milestonesToSolidify.entrySet()) {
```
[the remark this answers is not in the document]
hmoog: Its only getting called by

1. addToSolidificationQueue() (which in return is called by processNewlyAddedMilestones())
2. refillSolidificationQueue()

Both methods (processNewlyAddedMilestones and refillSolidificationQueue) are only called by the Thread method: milestoneSolidificationThread()

src/main/java/com/iota/iri/service/milestone/MilestoneSolidifier.java
```diff
@@ -0,0 +1,339 @@
...
+     * {@link #milestonesToSolidify} queue yet.
+     *
+     * It simply iterates over all milestones in the pool and looks for the one with the lowest index, that is not
+     * getting actively solidified, yet.
+     *
+     * @return the Map.Entry holding the earliest milestone or null if the pool does not contain any new candidates.
+     */
+    private Map.Entry<Hash, Integer> getNextSolidificationCandidate() {
+        Map.Entry<Hash, Integer> nextSolidificationCandidate = null;
+        for (Map.Entry<Hash, Integer> milestoneEntry : unsolidMilestonesPool.entrySet()) {
```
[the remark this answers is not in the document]
hmoog: unsolidMilestonesPool is a ConcurrentHashMap iterating over it is thread-safe.

src/main/java/com/iota/iri/service/milestone/MilestoneSolidifier.java
```diff
@@ -0,0 +1,330 @@
...
+     */
+    private boolean isSolid(Map.Entry<Hash, Integer> currentEntry) {
+        if (unsolidMilestonesPool.size() > 1) {
+            statusLogger.info("Solidifying milestone #" + currentEntry.getValue() +
+                    " [" + milestonesToSolidify.size() + " / " + unsolidMilestonesPool.size() + "]");
+        }
+
+        try {
+            return transactionValidator.checkSolidity(currentEntry.getKey(), true,
+                    SOLIDIFICATION_TRANSACTIONS_LIMIT * 4);
```
GalRogozinski: huh why `* 4`?

src/main/java/com/iota/iri/network/TransactionRequester.java
```diff
@@ -76,6 +76,22 @@ public void requestTransaction(Hash hash, boolean milestone) throws Exception {
...
+     *
+     * It can for example be used to determine if a transaction that was received by the node was actively requested
+     * while i.e. solidifying transactions or if a transaction arrived due to the gossip protocol.
+     *
+     * @param transactionHash hash of the transaction to check
+     * @param milestoneRequest flag that indicates if the hash was requested by a milestone request
+     * @return true if the transaction is in the set of transactions to be requested and false otherwise
+     */
+    public boolean isTransactionRequested(Hash transactionHash, boolean milestoneRequest) {
+        return (milestoneRequest && milestoneTransactionsToRequest.contains(transactionHash)) || (!milestoneRequest &&
```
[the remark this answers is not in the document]
hmoog: We want to be able to check if a transaction was requested by a milestone request and not just if it was requested at all. The local snapshots logic for example will allow transactions with an invalid timestamp to be stored if they have been requested while trying to solidify a milestone (backreferencing transactions).

    private boolean hasInvalidTimestamp(TransactionViewModel transactionViewModel) {
        // ignore invalid timestamps for transactions that where requested by our node while solidifying a milestone
        if(transactionRequester.isTransactionRequested(transactionViewModel.getHash(), true)) {
            return false;
        }

src/main/java/com/iota/iri/service/milestone/MilestoneSolidifier.java
```diff
@@ -0,0 +1,330 @@
...
+     */
+    private static final int SOLIDIFICATION_INTERVAL = 100;
+
+    /**
+     * Defines the maximum amount of transactions that are allowed to get processed while trying to solidify a milestone.
+     *
+     * Note: We want to find the next previous milestone and not get stuck somewhere at the end of the tangle with a
+     *       long running {@link TransactionValidator#checkSolidity(Hash, boolean)} call.
+     */
+    private static final int SOLIDIFICATION_TRANSACTIONS_LIMIT = 20000;
```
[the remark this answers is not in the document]
hmoog: maybe not worth it considering we will rework the solidification logic.

src/main/java/com/iota/iri/service/milestone/MilestoneSolidifier.java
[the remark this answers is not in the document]
hmoog: Thats intended - the next run of processNewlyAddedMilestones() will pick them up so they wont be forgotten. This little "trick" (processing the additions to the queue lazy) allows us to have this class working without any synchronized blocks and make it fast to call.

I actually think that milestones should be analyzed on arrival rather than rescanning all potential milestones over and over again I wanted this class to be as fast as possible. Tbh i think that the milestonetracker and this part of the code is the problem that causes IRI versions to consume more and more memory and CPU over time while the local snapshots version is stable at 7% cpu usage

src/main/java/com/iota/iri/service/milestone/MilestoneSolidifier.java
hmoog: LOL that was a bug - looks like i missed to replace the file in this PR with the updated one from the dev branch.

Its fixed (default level slighly increased an no * 4)

src/main/java/com/iota/iri/network/TransactionRequester.java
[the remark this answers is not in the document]
GalRogozinski: I think I understand what you are saying.
I misread the parenthesis... Can you please reformat like this: 
```
 return (milestoneRequest && milestoneTransactionsToRequest.contains(transactionHash))
                || (!milestoneRequest && milestoneTransactionsToRequest.contains(transactionHash) ||
                transactionsToRequest.contains(transactionHash));
```

The first `||` starts a new line implying that we have a new independent condition. The second one is on the same line implying we are continuing the conditional.

I simply thought before that `milestoneRequest` doesn't change the output. If I got confused others can get confused (or maybe I just can'r read :-P )

Issue #854, opened 2018-07-07
[Discussion] Checking if a node supports attachToTangle


rajivshah3:
### Description
Currently, Trinity will send an empty `attachToTangle` request to check if a node supports remote PoW. However, this seems to show in the IRI logs as `API Validation failed: Invalid parameters`. If the node does not support remote PoW, Trinity recognizes that from the response "attachToTangle is not available on this node" (or something similar). I think there should be a more definitive way (and it should be documented) for checking if a node supports `attachToTangle`. 

### Motivation
As more and more applications and projects use IOTA, there should be definitive and documented ways of checking if nodes support certain features, such as `attachToTangle`. The current implementation requires an empty `attachToTangle` request to be sent and causes a message in the logs to appear (`API Validation failed`). This could mislead users into thinking that something is wrong.

### Proposal
I have two solutions in mind:

#### 1
- Establish empty `attachToTangle` requests as the standard way to check if a node supports remote PoW, and document it
- Prevent these empty requests from triggering log messages which may mislead IRI users into thinking something is wrong

#### 2
- Incorporate a field in the `getNodeInfo` response (`supportsAttachToTangle` or `supportsRemotePow`, for example). It can be either `true` or `false`
- Added benefit: this will kill two birds with one stone, as only the `getNodeInfo` request is needed (which is already necessary to determine whether a node is in sync) to determine support for remote PoW. This will eliminate the need to send another API request

### Am I planning to do it myself with a PR?
Yes, but I first wanted to get other opinions on which proposal would be preferred

georgmittendorfer:
Imo option 2 is better because it is more explicit.

Option 1 is aequivalent to sending an _invalid and (potentially) unauthorized_ command, so it is not clear what takes precedence: validation or authorization check. In my opinion that would be a confusing API. But it is a valid solution if defined.

Also when using validation frameworks it would probably be cumbersome to work around validation to adhere to the API and to not output misleading error messages.

**Option 3** could be a separate command for checking if pow is enabled or a more general for checking certain features (easier to extend with further features). Something like  

- Request => `{"command": "querySupportedFeatures", "features": ["remotePow", "autoPromotion", "autoReattach"]}`
- Response => `{"features": ["remotePow"]}`

rajivshah3:
@georgmittendorfer nice! I think Option 3 is a great idea as well

nuriel77:
[...]

Issue #496, opened 2018-01-22
Enable apps to discover request size limits programmatically


todofixthis:
The IRI enforces certain configurable limits on requests (e.g., [`maxRequestList`](https://github.com/iotaledger/iri/blob/v1.4.1.6/src/main/java/com/iota/iri/service/API.java#L111)).

Applications can (respecfully) work around these limits by "chunking" large requests into several smaller ones.

However, currently the applications developer has to configure these limits manually to match the IRI configuration.

It would be a nice QoL enhancement if there were a way for an application to get these limits programmatically (e.g., from a `getNodeInfo` response), so that it can auto-configure its request chunking.

PR #1062, opened 2018-10-16
Obtain features of a node

### LINKED ISSUE #1038: As a developer, I want to retrieve the configuration of API limits on IRI nodes
### Description
Developers need a way to discover how request size limits on APIs on an IRI node are configured.

Previous PR on this #760, we should now use the new configuration interfaces that will be part of 1.5.5 - `ApiConfiguration`. 

### Motivation
Developers need to be able to programmatically retrieve request limits on IRI nodes. For example request size limits as per #496 . 

### Requirements
1. As a developer, I call an endpoint on an IRI node. 
2. I get back an object with the value of the following parameters.
          - minRandomWalks;
          - maxRandomWalks;
          - maxFindTxs;
          - maxRequestList;
          - maxGetTrytes;
          - maxBodyLength;

### Other
Paging can be a follow-up later for specific endpoints.

### Implementation
I think this should be a separate endpoint, like in #760.

### LINKED ISSUE #1039: As a developer, I want to be able to programmatically discover whether an IRI node support remote PoW
### Description
Developers need a way to discover whether an IRI node supports remote proof of work. This can now be fulfilled in a not very nice way by making an empty `attachToTangle`  request as discussed in #854. This also logs an error on the node. 

The node doesn't have the API call on the remote limit list as well. 

### Motivation
Developers need a supported and documented way to programmmatically discover that a node supports remote PoW. 
Trinity currently needs this functionality and uses an empty `attachToTangle` call and has a dependency on the return results of that empty call. We should have a clear contract for this scenario. 

### Requirements
1. As a developer, I call an endpoint on an IRI node. 
2. I get a response a that explicitly states that a node supports or doesn't support remote PoW. 

### Implementation
As discussed in #854, this could be a value returned via `getNodeInfo`.

kwek20:
# Description
This PR adds a field called `features` to the `getNodeInfo()` API call. 
It includes a list of features this node has enabled. 

All features are calculated after construction of the node, and before loading of the node, through the `Features` enum. This enum also contains all the possible features a node can have. 

Discussion: *What should the actual feature list be?*

Fixes #1039 
Solves #854 

## Type of change

# How Has This Been Tested?
- Ran getNodeInfo() and the expected features are send back
- Unit tests do not break

# Checklist:


chrisdukakis:
What do you think about a list of accessible commands? Simple as this:
```
{
    "accessibleCommands": [
        "getTransactionsToApprove",
        "attachToTangle",
        "storeTransactions",
        "etc..."
     ]
}
```

src/main/java/com/iota/iri/service/Feature.java
```diff
@@ -0,0 +1,84 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import com.iota.iri.Iota;
+import com.iota.iri.conf.IotaConfig;
+
+public enum Feature {
+    ATTACH_TO_TANGLE("attachToTange"),
```
jakubcech: attachToTange -> attachToTangle

src/main/java/com/iota/iri/service/Feature.java
```diff
@@ -0,0 +1,84 @@
...
+            features.add(ZMQ);
+        }
+        
+        List<Feature> apiFeatures = Arrays.asList(new Feature[] {
+                ATTACH_TO_TANGLE
+        });
+        
+        for (String disabled : configuration.getRemoteLimitApi()) {
+            switch (disabled) {
+            case "attachToTange":
```
jakubcech: attachToTange -> attachToTangle

jakubcech:
Not sure if this is expecting some more commits still. Just pointing out that this should have a form of `features that are enabled`, not just commands that are available. To make the contract clear, you should be able to query a node and discover that `PoW is enabled` in a clear way, regardless of the logic that enables/disables the remote PoW (which now is attachToTangle not in the remote limit list). Just listing that `attachToTangle` is available doesn't make that clear unless it's documented (which this should be as a scenario anyway). That's also why we just didn't go with a list of commands after some discussions @chrisdukakis. 

Also, #1038 is about request limits.

src/main/java/com/iota/iri/service/Feature.java
```diff
@@ -0,0 +1,112 @@
...
+    public String toString() {
+        return name;
+    }
+    
+    /**
+     * Calculates all features for this Iota node
+     * @param instance the instance of this node
+     * @return A list of features
+     */
+    public static Feature[] calculateFeatures(Iota instance) {
```
GalRogozinski: Can this receive `IotaConfig`?

src/main/java/com/iota/iri/service/Feature.java
```diff
@@ -0,0 +1,112 @@
...
+        }
+        features.addAll(apiFeatures);
+        
+        return features.toArray(new Feature[]{});
+    }
+    
+    /**
+     * Calculates all features for this Iota node
+     * @param instance the instance of this node
+     * @return A list of the features in readable name format
```
GalRogozinski: We are returning an array not a list

src/main/java/com/iota/iri/service/Feature.java
```diff
@@ -0,0 +1,112 @@
...
+     */
+    public static String[] calculateFeatureNames(Iota instance) {
+        Feature[] features = calculateFeatures(instance);
+        
+        List<String> featureNames = Arrays.stream(features)
+                .map(feature -> feature.toString())
+                .collect(Collectors.toList());
```
GalRogozinski: it is better to do `.toArray(Sting[]::new)`

src/main/java/com/iota/iri/service/Feature.java
```diff
@@ -0,0 +1,112 @@
...
+        
+        return features.toArray(new Feature[]{});
+    }
+    
+    /**
+     * Calculates all features for this Iota node
+     * @param instance the instance of this node
+     * @return A list of the features in readable name format
+     */
+    public static String[] calculateFeatureNames(Iota instance) {
```
GalRogozinski: Can this receive IotaConfig?

It doesn't need the whole instance

src/main/java/com/iota/iri/service/Feature.java
kwek20: It could be, but maybe some day we have to calculate features based on external factors outside of the config. Like only enable X if the database is up to date.

Or want to change to that when we have the need for it, and use config for now?

GalRogozinski: use configs for now
global scope is evil

python-regression/tests/features/machine1/1_api_tests.feature
```diff
@@ -22,6 +22,7 @@ Feature: Test API calls on Machine 1
 		|appName					|
 		|appVersion					|
 		|duration					|
+		|features                 |
```
GalRogozinski: indentation

legacycode:
See also #496 and #1200

PR #1063, opened 2018-10-16
Python Regression

GalRogozinski:
### Description
Currently we have a separate regression for [iri-regression tests](https://github.com/iotaledger/iri-regression). We want to combine it all to one repository.

### Motivation
When one contributes to the code, one may have to add/change a regression test. We want the change to IRI's code to be strongly coupled with the regression code. On one PR we can change both business logic and regression tests to ensure that the build will pass the CI.

### Open Questions:
Should we use [git subtree](https://www.atlassian.com/blog/git/alternatives-to-git-submodule-git-subtree) so we can keep the old iri-regression repo. It may be adopted by other node implementations. It will help to leave some decoupling between the projects but will addd complexity.


.gitignore
```diff
@@ -1,90 +1,202 @@
...
+/site
+
+# mypy
+.mypy_cache/
+
+# pyc files
+*.pyc 
+
+#Metafiles
+.metadata/
```
alon-e: is this for pyReg? if so, it should be under the `python-regression/` folder? see https://github.com/iotaledger/iri/commit/082af42c6c1d953036fefabf48a0652e221fcc2a

in general, this file becomes overly complex - can you open an issue to refactor it?

GalRogozinski: Yeah I'll separate it into 2

Issue #1054, opened 2018-10-12
Confusing way of constructing a random tip request 


gjeee:
### Bug description
In [Node.sendPacket()](https://github.com/iotaledger/iri/blob/29378a5604c9e3a11bf61de8f9ff548b5a5f5470/src/main/java/com/iota/iri/network/Node.java#L467)  when asking for a random tip, the hash of the tvm itself is used in the requestedHash placeholder:

        synchronized (sendingPacket) {
            System.arraycopy(transactionViewModel.getBytes(), 0, sendingPacket.getData(), 0, TransactionViewModel.SIZE);
            Hash hash = transactionRequester.transactionToRequest(rnd.nextDouble() < configuration.getpSelectMilestoneChild());
            System.arraycopy(hash != null ? hash.bytes() : transactionViewModel.getHash().bytes(), 0,
                    sendingPacket.getData(), TransactionViewModel.SIZE, reqHashSize);
            neighbor.send(sendingPacket);
        }

In [Node.preProcessReceivedData()](https://github.com/iotaledger/iri/blob/29378a5604c9e3a11bf61de8f9ff548b5a5f5470/src/main/java/com/iota/iri/network/Node.java#L273) it converts this to a NULL_HASH when they are the same:

                if (requestedHash.equals(receivedTransactionHash)) {
                    //requesting a random tip
                    requestedHash = Hash.NULL_HASH;
                }

My suggestion is to use the NULL_HASH directly in the requestedHash field. This is much more clear imo. So the adjusted code would look like:

    synchronized (sendingPacket) {
            System.arraycopy(transactionViewModel.getBytes(), 0, sendingPacket.getData(), 0, TransactionViewModel.SIZE);
            Hash hash = transactionRequester.transactionToRequest(rnd.nextDouble() < configuration.getpSelectMilestoneChild());
            System.arraycopy(hash != null ? hash.bytes() : Hash.NULL_HASH.bytes(), 0,
                    sendingPacket.getData(), TransactionViewModel.SIZE, reqHashSize);
            neighbor.send(sendingPacket);
        }

and no "translation" is necessary anymore in Node.preProcessReceivedData() 

### Open Questions (optional)
Anything you want to discuss.

### Am I planning to do it myself with a PR?
Yes if you like

GalRogozinski:
You will still need to keep the translation due to messages coming from old IRI versions.
the equality check is not expensive but your change does make the protocol clearer and easier to reason about.

[...]

PR #1095, opened 2018-10-24
Feat: Fixes + Improvements for IRI that are required for local snapshots

hmoog:
This PR contains a few fixes and improvements for IRI that are not directly related to local snapshots but that have been identified during its development and that are required for the activation of this feature later on.


src/main/java/com/iota/iri/network/Node.java
```diff
@@ -103,6 +113,7 @@ public void init() throws Exception {
         parseNeighborsConfig();
 
         executor.submit(spawnBroadcasterThread());
+        executor.submit(spawnRequesterThread());
         executor.submit(spawnTipRequesterThread());
```
gjeee: would it be an idea to rename spawnTipRequesterThread() to spawnLatestMilestoneRequesterThread() for clarity (and adjust the thread name as well)   ?

hmoog: It's not a LatestMilestoneRequester - the tip requester sends the hash of the milestone (request part) along with the data of that milestone (data part).

If we send a transaction that has a hash in its request part that matches the transaction in the data part, the receiving node interprets it as "i don't need anything - please send me a random tip".

So what this thread does is regularly requesting a random tip from its neighbours.

src/main/java/com/iota/iri/network/Node.java
```diff
@@ -103,6 +113,7 @@ public void init() throws Exception {
         parseNeighborsConfig();
 
         executor.submit(spawnBroadcasterThread());
+        executor.submit(spawnRequesterThread());
```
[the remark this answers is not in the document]
hmoog: Its the thread that works through the requestQueue rather than specifically requesting selective transactions - dunno if another name would make it easier to understand - but i don't really mind how its called

src/main/java/com/iota/iri/network/Node.java
gjeee: aha now i understand this. so it is actually a random tip request and as well sending the latest milestone.

thanks for clearing this up.....i was wondering about this construct in issue #1054  as well.

src/main/java/com/iota/iri/model/persistables/Transaction.java
```diff
@@ -35,6 +45,12 @@
     //public boolean confirmed = false;
     public boolean parsed = false;
     public boolean solid = false;
+
+    /**
+     * This flag indicates if the transaction is a coordinator issued milestone.
+     */
+    public boolean isSnapshot = false;
```
GalRogozinski: localSnapshot

since we don't change `snapshot` ...

src/main/java/com/iota/iri/network/Node.java
```diff
@@ -52,7 +62,7 @@
     private final DatagramPacket sendingPacket;
     private final DatagramPacket tipRequestingPacket;
 
-    private final ExecutorService executor = Executors.newFixedThreadPool(5);
+    private final ExecutorService executor = Executors.newFixedThreadPool(6);
```
GalRogozinski: why id you change this?

src/main/java/com/iota/iri/network/Node.java
```diff
@@ -504,6 +515,49 @@ private Runnable spawnBroadcasterThread() {
...
+     *
+     * Without this thread we are limited to the amount of incoming transactions while syncing the node, because we can
+     * only request new transactions when we are re-broadcasting these received transactions.
+     *
+     * To not send unnecessary requests for "temporarily" missing transactions while receiving broadcasts from the
+     * network, we only trigger its logic when the queue exceeds the {@link #REQUESTER_THREAD_ACTIVATION_THRESHOLD}.
+     *
+     * @return lambda for the requester {@link Thread}
+     */
+    private Runnable spawnRequesterThread() {
```
GalRogozinski: Maybe open an issue about doing something a bit more efficient.
Instead of having an infinite while loop maybe you can initiate this thread once the threshold has been exceeded.

src/main/java/com/iota/iri/network/UDPReceiver.java
```diff
@@ -69,7 +69,7 @@ private Runnable spawnReceiverThread() {
 
             while (!shuttingDown.get()) {
 
-                if (((processed + dropped) % 50000 == 0)) {
+                if (((processed + dropped) % 50000 == 49999)) {
```
hmoog: to not spam the console output when there are 0 processed packets

src/main/java/com/iota/iri/storage/rocksDB/RocksDBPersistenceProvider.java
```diff
@@ -336,7 +336,7 @@ public void deleteBatch(Collection<Pair<Indexable, ? extends Class<? extends Per
                     ColumnFamilyHandle handle = classTreeMap.get(entry.hi);
                     writeBatch.remove(handle, keyBytes);
                     ColumnFamilyHandle metadataHandle = metadataReference.get(entry.hi);
-                    if (metadataReference != null) {
+                    if (metadataHandle != null) {
```
hmoog: its just wrong :P

src/main/java/com/iota/iri/controllers/MilestoneViewModel.java
```diff
@@ -143,4 +143,15 @@ public void delete(Tangle tangle) throws Exception {
...
     }
 
+    /**
+     * This method creates a human readable string representation of the milestone.
+     *
+     * It can be used to directly append the milestone in error and debug messages.
+     *
+     * @return human readable string representation of the milestone
+     */
+    @Override
```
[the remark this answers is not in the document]
hmoog: will take a node and add it in a separate PR

src/main/java/com/iota/iri/model/persistables/Transaction.java
hmoog: actually this should in fact be isMilestone - i just kept the old naming scheme (snapshotIndex) O_O

src/main/java/com/iota/iri/controllers/TransactionViewModel.java
```diff
@@ -413,34 +413,34 @@ public void setSnapshot(Tangle tangle, final int index) throws Exception {
...
      * the two types of transactions (normal transactions / milestones) very fast and efficiently without issuing
      * further database queries or even full verifications of the signature. If it is set to true one can for example
      * use the snapshotIndex() method to retrieve the corresponding MilestoneViewModel object.
      *
      * @return true if the transaction is a milestone and false otherwise
      */
-    public boolean isSnapshot() {
-        return transaction.isSnapshot;
+    public boolean isMilestone() {
+        return transaction.isMilestone;
```
GalRogozinski: Java convention is that a getter method of a boolean property starts with the prefix `is` and the boolean property itself doesn't have "is" in the name.

It is not something I made up but a java convention I want to follow.
Static code analyzers may take advantage of such conventions for instance.
So please change the property (field name) to milestone

PR #1096, opened 2018-10-26
Fixed wrong Hash objects used

### LINKED ISSUE #1079: The wrong Hash type is used in several places of IRI
### Bug description
The recently merged hash refactor uses the wrong kind of hashes in several places and should be checked again.

### Expected behaviour
https://github.com/iotaledger/iri/blob/dev/src/main/java/com/iota/iri/Snapshot.java#L86 should be an ADDRESS hash
https://github.com/iotaledger/iri/blob/dev/src/main/java/com/iota/iri/model/StateDiff.java#L25 should be an ADDRESS hash
https://github.com/iotaledger/iri/blob/dev/src/main/java/com/iota/iri/service/API.java#L999 should be a TRANSACTION hash (maybe? didnt examine the code enough to be 100% sure)

### Actual behaviour
The wrong hash types are used in several places.

### What should be done
Better check all the changed lines and verify that are really correct.

kwek20:
# Description

Various Hashes were created from the wrong hash type, this has been fixed.

Fixes #1079 

## Type of change

# How Has This Been Tested?
Manual verify if all uses of the `HashFactory` and `NULL_HASH`

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/service/API.java
+++ b/src/main/java/com/iota/iri/service/API.java
@@ -996,7 +996,7 @@ public class API {
         if (tips == null || tips.size() == 0) {
             hashes = Collections.singletonList(instance.milestoneTracker.latestSolidSubtangleMilestone);
         } else {
-            hashes = tips.stream().map(address -> (HashFactory.ADDRESS.create(address)))
+            hashes = tips.stream().map(address -> (HashFactory.TRANSACTION.create(address)))
                     .collect(Collectors.toCollection(LinkedList::new));
         }
         try {
--- a/src/main/java/com/iota/iri/service/API.java
+++ b/src/main/java/com/iota/iri/service/API.java
@@ -996,7 +996,7 @@ public class API {
         if (tips == null || tips.size() == 0) {
             hashes = Collections.singletonList(instance.milestoneTracker.latestSolidSubtangleMilestone);
         } else {
-            hashes = tips.stream().map(address -> (HashFactory.TRANSACTION.create(address)))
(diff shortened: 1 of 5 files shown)


src/main/java/com/iota/iri/service/API.java
```diff
@@ -996,7 +996,7 @@ private AbstractResponse getBalancesStatement(final List<String> addresses, fina
         if (tips == null || tips.size() == 0) {
             hashes = Collections.singletonList(instance.milestoneTracker.latestSolidSubtangleMilestone);
         } else {
-            hashes = tips.stream().map(address -> (HashFactory.ADDRESS.create(address)))
+            hashes = tips.stream().map(address -> (HashFactory.TRANSACTION.create(address)))
```
gjeee: Would it be an idea to use something different than **address**, eg. **tip** or **txhash**. Using **address** is confusing imo. Maybe i am completely off here but I thought let's just share it.

kwek20: @gjeee For cases like Milestones etc, this might be useful. But when you make certain transactions a `tip` object, you have to revert them again once they are being approved by others. (not specifically in this case, but for example storing transactions somewhere)

For now this PR just changes wrong types, and doesn't do more specific types

gjeee: this is not what I meant, but i see that your latest commit does what I was trying to suggest :-)

kwek20: Oooooh, i see what you mean now :)
Yes i noticed the strange param naming when you quoted it, so fixed it, thanks!

PR #1121, opened 2018-11-06
add "numberOfStaleTransactions" to getNeighbors API call

### LINKED ISSUE #1018: Wrongly labels transactions as 'invalid'
### Bug description
Newly updated nodes labels perfectly valid transactions as 'invalid' if they have too old timestamp for the updated node.
In an environment where users are told that they don't need to upgrade node unless they want to shrink DB-size, it is a wrong assumption that such tx'es are 'invalid'.  They just should not be stored in local db.

### Steps To Reproduce
1. Keep one node on 1.5.3 
2. Upgrade a nbr to 1.5.4
3. Let the 1.5.3-node forward som incoming transactions published prior to last snapshot.

### Expected behaviour
1.5.4-node should just disregard transactions without counting them as 'invalid'

### Actual behaviour
1.5.4-node will mark tx'es as 'invalid', and stats for 1.5.3-nbr will possibly lead to removal of nbr...

alon-e:
# Description

Seperates "numberOfInvalidTransactions" and "numberOfStaleTransactions" in `getNeighbors`

```
{
  "neighbors": [
    {
      "address": "localhost:12345",
      "numberOfAllTransactions": 0,
      "numberOfRandomTransactionRequests": 0,
      "numberOfNewTransactions": 0,
      "numberOfInvalidTransactions": 0,
      "numberOfStaleTransactions": 0,
      "numberOfSentTransactions": 0,
      "connectionType": "tcp"
    }
  ],
  "duration": 1
}
```

Fixes #1018 

## Type of change

- Documentation Fix

# How Has This Been Tested?

_Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration_

- Tested locally

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/network/Neighbor.java
+++ b/src/main/java/com/iota/iri/network/Neighbor.java
@@ -14,6 +14,7 @@ public abstract class Neighbor {
     private long numberOfInvalidTransactions;
     private long randomTransactionRequests;
     private long numberOfSentTransactions;
+    private long numberOfStaleTransactions;
 
     private boolean flagged = false;
     public boolean isFlagged() {
@@ -84,11 +85,15 @@ public abstract class Neighbor {
     public void incInvalidTransactions() {
     	numberOfInvalidTransactions++;
     }
-    
+
+    void incStaleTransactions() {
+        numberOfStaleTransactions++;
+    }
+
     public void incSentTransactions() {
         numberOfSentTransactions++;
     }
-    
+
     public long getNumberOfAllTransactions() {
 		return numberOfAllTransactions;
 	}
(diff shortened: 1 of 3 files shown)


src/main/java/com/iota/iri/network/Neighbor.java
```diff
@@ -84,19 +85,27 @@ void incRandomTransactionRequests() {
     public void incInvalidTransactions() {
     	numberOfInvalidTransactions++;
     }
-    
+
+    void incStaleTransactions() {
```
GalRogozinski: I suppose you didn't add `public` like the rest of the methods becuase you only want classes in the network package to have access.

I am fine with that :-)

jakubcech:
That was quick 🏃

PR #1130, opened 2018-11-08
document converter

alon-e:
# Description

Code comments for the Converter.java class. 

Fixes # 1117

## Type of change
- Documentation Fix

-------

I haven't documented the `Pair<->Trits` methods. I don't see them in use either than testing and frankly didn't follow what each method does.

Issues to open:
1. method naming is not consistent between inverse functions.
2. - " - arguments
3. `pos` & `offset` are used interchangeably

### DIFF
--- a/src/main/java/com/iota/iri/utils/Converter.java
+++ b/src/main/java/com/iota/iri/utils/Converter.java
@@ -3,7 +3,7 @@ package com.iota.iri.utils;
 import java.util.Arrays;
 
 /**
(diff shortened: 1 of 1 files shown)


alon-e:
btw, @jakubcech - Github just added a cool feature to suggest changes: see the +- symbol (ctrl+g)
this allows the PR author to automatically accept and commit the suggested changes in the PR.
would be great to use these in Docs PRs as the changes are often one-liners.

https://blog.github.com/2018-10-16-future-of-software/#suggested-changes-public-beta

PR #1131, opened 2018-11-08
Feat: Introducing several executor services for IRI

hmoog:
# Description

This PR introduces ExecutorServices that will be used for the background workers of IRI (and in particular for local snapshots and its related classes). The introduced ExecutorServices allow the scheduling of Tasks with automated logging capabilities without using "while(!shuttingDown) { ... }" loops by using the well known ScheduledExecutorService interface.

## Type of change

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/utils/thread/BoundedScheduledExecutorService.java
+++ b/src/main/java/com/iota/iri/utils/thread/BoundedScheduledExecutorService.java
@@ -1,0 +1,925 @@
+package com.iota.iri.utils.thread;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.*;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * This class represents a {@link SilentScheduledExecutorService} that accepts only a pre-defined amount of tasks that
+ * can be queued or executed at the same time. All tasks exceeding the defined limit will be ignored (instead of being
+ * queued) by either throwing a {@link RejectedExecutionException} or returning {@code null} depending on the method we
+ * call (non-silent vs silent).<br />
+ * <br />
+ * Whenever a non-recurring task finishes (or a recurring one is cancelled through its {@link Future}), it makes space
+ * for a new task. This is useful for classes like the {@link com.iota.iri.utils.log.interval.IntervalLogger} that want
+ * to delay an action if and only if there is no other delayed action queued already.<br />
+ * <br />
+ * Note: In contrast to other existing implementations like the SizedScheduledExecutorService of the apache package,
+ *       this class is thread-safe and will only allow to spawn and queue the exact amount of tasks defined during its
+ *       creation (since it does not rely on approximate numbers like the queue size).<br />
+ */
+public class BoundedScheduledExecutorService implements SilentScheduledExecutorService, ReportingExecutorService {
+    /**
+     * Holds the maximum amount of tasks that can be submitted for execution.<br />
+     */
+    private final int capacity;
+
+    /**
+     * Holds the underlying {@link ScheduledExecutorService} that manages the Threads in the background.<br />
+     */
+    private final ScheduledExecutorService delegate;
+
+    /**
+     * Holds a set of scheduled tasks tasks that are going to be executed by this
+     * {@link ScheduledExecutorService}.<br />
+     */
+    private final Set<TaskDetails> scheduledTasks = ConcurrentHashMap.newKeySet();
+
+    /**
+     * Thread-safe counter that is used to determine how many tasks were exactly scheduled already.<br />
+     * <br />
+     * Note: Whenever a task finishes, we clean up the used resources and make space for new tasks.<br />
+     */
+    private AtomicInteger tasksQueued = new AtomicInteger(0);
+
+    /**
+     * Creates an executor service that that accepts only a pre-defined amount of tasks that can be queued and run at
+     * the same time.<br />
+     * <br />
+     * All tasks exceeding the defined limit will be ignored (instead of being queued) by either throwing a
+     * {@link RejectedExecutionException} or returning {@code null} depending on the method we call (non-silent vs
+     * silent).<br />
+     *
+     * @param capacity the amount of tasks that can be scheduled simultaneously
+     */
+    public BoundedScheduledExecutorService(int capacity) {
+        this.capacity = capacity;
+
+        delegate = Executors.newScheduledThreadPool(capacity);
+    }
+
+    //region METHODS OF ReportingExecutorService INTERFACE /////////////////////////////////////////////////////////////
+
+    /**
+     * {@inheritDoc}
+     * <br />
+     * It simply adds the task to the internal set of scheduled tasks.<br />
+     */
+    @Override
+    public void onScheduleTask(TaskDetails taskDetails) {
+        scheduledTasks.add(taskDetails);
+    }
+
+    @Override
+    public void onStartTask(TaskDetails taskDetails) {}
+
+    @Override
+    public void onFinishTask(TaskDetails taskDetails, Throwable error) {}
+
+    @Override
+    public void onCancelTask(TaskDetails taskDetails) {}
+
+    /**
+     * {@inheritDoc}
+     * <br />
+     * It frees the reserved resources by decrementing the {@link #tasksQueued} counter and removing the task from the
+     * {@link #scheduledTasks} set.<br />
+     */
+    @Override
+    public void onCompleteTask(TaskDetails taskDetails, Throwable error) {
+        scheduledTasks.remove(taskDetails);
+        tasksQueued.decrementAndGet();
+    }
+
+    //endregion ////////////////////////////////////////////////////////////////////////////////////////////////////////
+
+    //region METHODS OF SilentScheduledExecutorService INTERFACE ///////////////////////////////////////////////////////
+
+    /**
+     * {@inheritDoc}
+     * <br />
+     * Note: Since the {@link ScheduledFuture} returned by this method allows to cancel jobs without their unwinding
+     *       logic being executed, we wrap the returned {@link ScheduledFuture} AND the {@link Runnable} to correctly
+     *       free the resources if its {@link ScheduledFuture#cancel(boolean)} method is called.<br />
+     */
+    @Override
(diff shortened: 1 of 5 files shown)


src/main/java/com/iota/iri/utils/thread/DedicatedScheduledExecutorService.java
```diff
@@ -0,0 +1,555 @@
...
+ * This class represents a {@link ScheduledExecutorService} that is associated with one specific task for which it
+ * provides automatic logging capabilities<br />
+ * <br />
+ * It informs the user about its lifecycle using the logback loggers used by IRI. In addition it offers "silent" methods
+ * of the {@link ScheduledExecutorService} that do not throw {@link Exception}s when we try to start the same task
+ * multiple times. This is handy for implementing the "start" and "shutdown" methods of the background workers of IRI
+ * that would otherwise have to take care of not starting the same task more than once (when trying to be robust against
+ * coding errors or tests that start the same thread multiple times).<br />
+ */
+public class DedicatedScheduledExecutorService extends BoundedScheduledExecutorService {
```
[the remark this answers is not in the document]
hmoog: on purpose

src/main/java/com/iota/iri/utils/thread/DedicatedScheduledExecutorService.java
```diff
@@ -0,0 +1,555 @@
...
+     * <br />
+     * Note: The used logger can be overwritten by providing a different logger in the constructor (to have transparent
+     *       log messages that look like they are coming from a different source).<br />
+     */
+    private static final Logger DEFAULT_LOGGER = LoggerFactory.getLogger(DedicatedScheduledExecutorService.class);
+
+    /**
+     * Holds a reference to the logger that is used to emit messages.<br />
+     */
+    private final Logger logger;
```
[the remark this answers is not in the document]
hmoog: on purpose

src/main/java/com/iota/iri/utils/thread/BoundedScheduledExecutorService.java
```diff
@@ -0,0 +1,931 @@
...
+     * {@link ScheduledExecutorService}.<br />
+     */
+    private final Set<TaskDetails> scheduledTasks = ConcurrentHashMap.newKeySet();
+
+    /**
+     * Thread-safe counter that is used to determine how many tasks were exactly scheduled already.<br />
+     * <br />
+     * Note: Whenever a task finishes, we clean up the used resources and make space for new tasks.<br />
+     */
+    private AtomicInteger tasksQueued = new AtomicInteger(0);
```
GalRogozinski: Perhaps change the name to `tasksCounter` or `tasksQueuedCounter`

I got kinda of confused while reviewing

src/main/java/com/iota/iri/utils/thread/BoundedScheduledExecutorService.java
```diff
@@ -0,0 +1,931 @@
...
+
+    /**
+     * This interface is used to generically describe the lambda that is used to create the unwrapped future from the
+     * delegated {@link ScheduledExecutorService} by passing in the wrapped command.<br />
+     *
+     * @param <RESULT> the kind of future returned by this factory ({@link ScheduledFuture} vs {@link Future})
+     * @param <ARGUMENT> type of the wrapped command that is passed in ({@link Runnable} vs {@link Callable})
+     */
+    @FunctionalInterface
+    private interface FutureFactory<RESULT, ARGUMENT> {
```
GalRogozinski: It is funny, I was so used to only using one letter in type params it never occured to me that we can write a word :laughing: 

Anyhows, :+1: on the clarity

src/main/java/com/iota/iri/utils/thread/BoundedScheduledExecutorService.java
```diff
@@ -0,0 +1,931 @@
...
+     * through their corresponding {@link Future}.<br />
+     *
+     * @param requestedJobCount the amount of tasks that shall be scheduled
+     * @return true if we could reserve the given space and false otherwise
+     */
+    private boolean reserveCapacity(int requestedJobCount) {
+        if (tasksQueued.addAndGet(requestedJobCount) <= capacity) {
+            return true;
+        } else {
+            tasksQueued.addAndGet(-requestedJobCount);
```
GalRogozinski: I don't get why this is here..
I didn't notice you are freeing resources due to this...

src/main/java/com/iota/iri/utils/thread/BoundedScheduledExecutorService.java
```diff
@@ -0,0 +1,931 @@
...
+                            delegate.isShutdown()) {
+
+                        onCompleteTask(taskDetails, error);
+                    } else {
+                        taskDetails.getScheduledForExecution().set(true);
+                    }
+                }
+            }
+
+            return null;
```
GalRogozinski: Are you sure you handled all the cases to prevent later `NullPointerExceptions`?
If the answer is no, maybe it is better to use `Optional`?

src/main/java/com/iota/iri/utils/thread/DedicatedScheduledExecutorService.java
```diff
@@ -0,0 +1,555 @@
...
+     *                   always be logged)
+     * @param logger logback logger that shall be used for the origin of the log messages
+     * @param debug debug flag that indicates if every "run" should be accompanied with a log message
+     */
+    public DedicatedScheduledExecutorService(String threadName, Logger logger, boolean debug) {
+        super(1);
+
+        this.threadName = threadName;
+        this.logger = logger;
+        this.debug = debug;
```
GalRogozinski: Instead of having a separate debug flag can we maybe use `logger.isDebug()` throughout the class logic?

src/main/java/com/iota/iri/utils/thread/ReportingExecutorService.java
```diff
@@ -0,0 +1,58 @@
...
+     */
+    void onCancelTask(TaskDetails taskDetails);
+
+    /**
+     * This method gets called whenever a task completes.<br />
+     * <br />
+     * This can be through either raising an exception, cancelling from the outside or by simply terminating in a normal
+     * manner. For recurring tasks this only gets called once.<br />
+     *
+     * @param taskDetails object containing details about this task
```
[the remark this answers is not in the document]
hmoog: added it

src/main/java/com/iota/iri/utils/thread/ReportingExecutorService.java
```diff
@@ -0,0 +1,58 @@
...
+     */
+    void onStartTask(TaskDetails taskDetails);
+
+    /**
+     * This method gets called whenever a task is finished.<br />
+     * <br />
+     * For recurring tasks it is called multiple times.<br />
+     *
+     * @param taskDetails object containing details about this task
```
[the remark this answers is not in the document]
hmoog: added it

src/main/java/com/iota/iri/utils/thread/BoundedScheduledExecutorService.java
hmoog: renamed it

src/main/java/com/iota/iri/utils/thread/BoundedScheduledExecutorService.java
GalRogozinski: ohh of course,
very obvious the second time I look at it :-)

src/main/java/com/iota/iri/utils/thread/DedicatedScheduledExecutorService.java
GalRogozinski: How do we control the debug flag?
If it is by manually changing in the code it is fine for the dev who works on it, but not if we communicate with a user experiencing problems.
I would think this should be able to activate at least with `logger.isTrace()` so one can see the messages without changing the code.
Would you agree with that?

src/main/java/com/iota/iri/utils/thread/BoundedScheduledExecutorService.java
GalRogozinski: Unless you can explain to me how you are sure there are no `NullPointerExceptions` please do that :-)

src/main/java/com/iota/iri/utils/thread/DedicatedScheduledExecutorService.java
hmoog: It's only for the developer to be able to add extra log messages while developing a new task which allows him to determine if something obvious doesn't work as planned (like the given example of a task not terminating). This should not be controlled by the user.

Exceptions or problems that might arise within the scope of a user interacting with IRI should always be handled by the specific logic of the task. Enabling a system wide debug flag for "all" "Threads" in IRI and printing additional "start / stop" messages for all Threads that use this ExecutorService would in fact make it even harder to debug problems because the console would be spammed with completely useless messages that are not related to the problem (depending on the chosen intervals - every few milliseconds).

src/main/java/com/iota/iri/utils/thread/BoundedScheduledExecutorService.java
hmoog: I don't get it - the only way to retrieve the result is by using future.get() which will already throw a CancellationException if the future was cancelled.

This return statement can only be reached "internally" but the null will never be accessible to any caller.

src/main/java/com/iota/iri/utils/thread/DedicatedScheduledExecutorService.java
GalRogozinski: This is why I offered to do in `trace` mode.
I understand your reasoning though and accept it.

src/main/java/com/iota/iri/utils/thread/BoundedScheduledExecutorService.java
GalRogozinski: Hmm
You create a `Callable` and sometimes you pass it to an executor (`delegate.invokeAny()` for example) as the task to be invoked.

According to the javadocs of `invokeAny` (or any of the methods that try to run the `Callable` tasks), a NullPointerException will be thrown...

hmoog: Yes, if the Callable itself is null but not if the callable return null.

So if I do sth like executorService.schedule...(null);

PR #1135, opened 2018-11-09
Feat: Making IRI use the initialSnapshot and the solidEntryPoints

hmoog:
# Description

This PR contains the changes that are necessary for checking against the solidEntryPoints instead of the NULL_HASH. It does not activate the creation of local snapshots, yet and also doesn't modify the LedgerValidator and MIlestoneTracker to change the "latestSnapshot" instance. The "initialSnapshot" that holds the solidEntryPoints is however read already from the new SnapshotProvider so this activates parts of the new code already. Since the "initialSnapshot" will be the last global snapshot (with the NULL_HASH as its only solidEntryPoint) in the absence of local snapshot files, it is a non-breaking change.

Once this PR is merged there will be a last PR that also enables the use of the "latestSnapshot" instance, which will fully enable the new features.

## Type of change

# How Has This Been Tested?

- All integration tests should succeed

# Checklist:


src/main/java/com/iota/iri/TransactionValidator.java
```diff
@@ -72,11 +69,16 @@ public int getMinWeightMagnitude() {
     }
 
     private boolean hasInvalidTimestamp(TransactionViewModel transactionViewModel) {
+        // ignore invalid timestamps for transactions that where requested by our node while solidifying a milestone
```
GalRogozinski:
```suggestion
        // ignore invalid timestamps for transactions that were requested by our node while solidifying a milestone
```

PR #1145, opened 2018-11-13
Niming weighted edge

stplaydog:


stplaydog:
merge to the wrong place

PR #1161, opened 2018-11-19
tweak tip selection documentation.

### LINKED ISSUE #1119: Service - TipSelection code documentation
### Description
Write code comments for the TipSelection part of the service package. It already has docs in a supposedly good state, but this part of the code is very important to reflect reality.

alon-e:
# Description

tweak tip selection documentation.

Fixes #1119 

## Type of change

- Documentation Fix

### DIFF
--- a/src/main/java/com/iota/iri/service/tipselection/impl/WalkerAlpha.java
+++ b/src/main/java/com/iota/iri/service/tipselection/impl/WalkerAlpha.java
@@ -23,6 +23,10 @@ import java.util.stream.Collectors;
  */
 public class WalkerAlpha implements Walker {
 
+    /**
+    * {@code alpha}: a positive number that controls the randomness of the walk. 
+    * The closer it is to 0, the less bias the random walk will be.
+    */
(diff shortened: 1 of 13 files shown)


src/main/java/com/iota/iri/service/tipselection/impl/WalkerAlpha.java
```diff
@@ -40,10 +49,16 @@ public WalkerAlpha(TailFinder tailFinder, Tangle tangle, MessageQ messageQ, Rand
...
 
+    /**
+     * @return {@link WalkerAlpha#alpha}
+     */
     public double getAlpha() {
         return alpha;
     }
 
+    /**
+     * @param alpha {@link WalkerAlpha#alpha}
```
GalRogozinski: I would add a javadoc to `WalkerAlpha#alpha`
Doesn't even need to be mathematical. Can be something along the lines as:
a positive number that controls the randomness of the walk. The closer it is to 0, the rating of the transaction matters less.

PR #1157, opened 2018-11-18
Document Persistence Provider 

GalRogozinski:
# Description

Document Persistence Provider 
Fixes #1118

## Type of change
- Documentation Fix

PR #1169, opened 2018-11-19
Document Persistable and Indexable

GalRogozinski:
# Description

Document Persistable and Indexable

Fixes #1157 

## Type of change
- Documentation Fix

### DIFF
--- a/src/main/java/com/iota/iri/storage/Persistable.java
+++ b/src/main/java/com/iota/iri/storage/Persistable.java
@@ -2,10 +2,49 @@ package com.iota.iri.storage;
 
 import java.io.Serializable;
 
+/**
+ * Any object that can be serialized and stored in the DB.
+ *
+ * @see Indexable
+ */
 public interface Persistable extends Serializable {
+    /**
+     * Serializes the core fields of the object to bytes that can be stored in a database. Core fields are all the
+     * fields that represent the object in other nodes in the network.
+     *
+     * @return the serialized bytes that represents the core original object
+     */
     byte[] bytes();
+
+    /**
+     * Recreates the core state (populates the fields) of the object by deserializing {@code bytes}.
+     *
+     * @param bytes the serialized version the object
+     */
     void read(byte[] bytes);
+
+    /**
+     * Serializes fields that usually contain data about the object that the node itself calculates
+     * and is not passed to different nodes. This data is supposed to be stored in a different column family/table then
+     * the data serialized by {@link #bytes()}.
+     *
+     * @return the serialized bytes that represents the metadata of the object
+     */
     byte[] metadata();
+
+    /**
+     * Recreates the metadata (populates the relevant fields) of the object by deserializing {@code bytes}.
+     *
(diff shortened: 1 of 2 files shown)


src/main/java/com/iota/iri/storage/Persistable.java
```diff
@@ -2,10 +2,49 @@
 
 import java.io.Serializable;
 
+/**
+ * Any object that can be serialized and stored in the DB.
```
kwek20:
```suggestion
 * Any object that can be serialized and stored in the database.
```

You write database everywhere else, should keep this constant

src/main/java/com/iota/iri/storage/Persistable.java
```diff
@@ -2,10 +2,49 @@
...
+ *
+ * @see Indexable
+ */
 public interface Persistable extends Serializable {
+    /**
+     * Serializes the core fields of the object to bytes that can be stored in a database. Core fields are all the
+     * fields that represent the object in other nodes in the network.
```
kwek20: "the object in other nodes" is quite vague, the other nodes likely also have metadata
```suggestion
     * fields that represent the core data required to construct this object.
```

src/main/java/com/iota/iri/storage/Persistable.java
```diff
@@ -2,10 +2,49 @@
...
+     *
+     * @return the serialized bytes that represents the core original object
+     */
     byte[] bytes();
+
+    /**
+     * Recreates the core state (populates the fields) of the object by deserializing {@code bytes}.
+     *
+     * @param bytes the serialized version the object
```
kwek20:
```suggestion
     * @param bytes the serialized version of the object
```

src/main/java/com/iota/iri/storage/Persistable.java
```diff
@@ -2,10 +2,49 @@
...
+     */
     void read(byte[] bytes);
+
+    /**
+     * Serializes fields that usually contain data about the object that the node itself calculates
+     * and is not passed to different nodes. This data is supposed to be stored in a different column family/table then
```
kwek20:
```suggestion
     * and is not passed to different nodes. This data is stored in a different column family/table then
```
Its not supposed to right? We always do this.

src/main/java/com/iota/iri/storage/Persistable.java
```diff
@@ -2,10 +2,49 @@
...
+
+    /**
+     * Recreates the metadata (populates the relevant fields) of the object by deserializing {@code bytes}.
+     *
+     * @param bytes the serialized bytes that represents the metadata of the object
+     */
     void readMetadata(byte[] bytes);
+
+    /**
+     * Specifies whether the object should be appended to an existing key/index or should it replace the old
```
kwek20:
```suggestion
     * Specifies whether the object should be appended to an existing key/index or should replace the old
```

GalRogozinski:
@kwek20 
no need to capitalize

PR #1174, opened 2018-11-20
Introducing a repair mechanism for corruptions in the ledger database

hmoog:
# Description

This PR implements a repair mechanism for minor corruptions in the IRI database that causes conflicting milestones to be reverted and re-processed without having to reset the whole database.

(In addition it removes some unused imports)

## Detailed Description

If IRI faces database corruptions in the "snapshotIndex" of transactions, it can happen that the same transaction gets processed as being confirmed by two or more milestones and therefore booking its balance multiple times leading to inconsistent balances. This causes the nodes to fall out of sync and report "Skipping negative value for address: ..." in an endless loop. The only way to recover from this problem is to do a --rescan or sometimes even remove the database and start a complete resync.

There are numerous reasons why these corruptions in the snapshotIndex can appear:

- milestones got processed in the wrong order (fixed already but existing databases might still have these corruptions)
- IRI crashes or gets stopped before the modified snapshotIndex was flushed to the database.
- Race conditions between different threads that try to write to the same transaction at the same time (for example solid = true + snapshotIndex = xyz) and therefore overwriting the changes of the other thread. Note: Updating just a single property of the transaction causes the whole transaction to be serialized and written again. **(THIS HAPPENS ALOT DURING TIMES OF HIGH LOADS / NETWORK ACTIVITY)**

## Type of change

# Checklist:


src/main/java/com/iota/iri/service/milestone/impl/MilestoneServiceImpl.java
```diff
@@ -312,5 +367,41 @@ private boolean isMilestoneBundleStructureValid(List<TransactionViewModel> bundl
...
+     * resetting additional milestones in recursive calls.<br />
+     * <br />
+     * It first checks if the desired {@code milestoneIndex} is reachable by this node and then triggers the reset
+     * by:<br />
+     * <br />
+     * 1. resetting the ledger state if it addresses a milestone before the current latest solid milestone<br />
+     * 2. resetting the {@code milestoneIndex} of all transactions that were confirmed by the current milestone<br />
+     * 3. deleting the corresponding {@link StateDiff} entry from the database<br />
+     *
+     * @param index milestone index that shall
```
alon-e:
```suggestion
     * @param index milestone index that shall be reverted
``` 
or ?

src/main/java/com/iota/iri/service/milestone/impl/LatestSolidMilestoneTrackerImpl.java
```diff
@@ -154,6 +173,39 @@ private void latestSolidMilestoneTrackerThread() {
...
+     * <br />
+     * If the application of the milestone fails, we start a repair routine which will revert the milestones preceding
+     * our current milestone and consequently try to reapply them in the next iteration of the {@link
+     * #checkForNewLatestSolidMilestones()} method (until the problem is solved).<br />
+     *
+     * @param milestone the milestone that shall be applied to the ledger state
+     * @throws Exception if anything unexpected goes wrong while applying the milestone to the ledger
+     */
+    private void applySolidMilestoneToLedger(MilestoneViewModel milestone) throws Exception {
+        if (!ledgerService.applyMilestoneToLedger(milestone)) {
```
alon-e: this reverse if statement took me a while to read - better to have the positive (common case) at the beginning IMO.
(i.e. swap)

src/main/java/com/iota/iri/service/milestone/impl/LatestSolidMilestoneTrackerImpl.java
```diff
@@ -154,6 +173,39 @@ private void latestSolidMilestoneTrackerThread() {
...
+
+    /**
+     * Resets the internal variables that are used to keep track of the repair process.<br />
+     * <br />
+     * It gets called whenever we advance to a milestone that has a higher milestone index than the milestone that
+     * initially caused the repair routine to kick in (see {@link #revertPrecedingMilestones(MilestoneViewModel)}.<br />
+     *
+     * @param processedMilestone the milestone that currently gets processed
+     */
+    private void resetRepairBackoffCounterIfProblemSolved(MilestoneViewModel processedMilestone) {
```
alon-e: this name is a bit misleading, it's also the common case, where there was no problem to begin with.

src/main/java/com/iota/iri/service/milestone/impl/LatestSolidMilestoneTrackerImpl.java
```diff
@@ -191,4 +243,32 @@ private void logChange(int prevSolidMilestoneIndex) {
...
+     * reset only the last milestone, then the two last milestones, then the three last milestones (and so on ...) until
+     * the problem was fixed.<br />
+     * <br />
+     * To be able to tell when the problem is fixed and the {@link #repairBackoffCounter} can be reset, we store the
+     * milestone index that caused the problem the first time we call this method.<br />
+     *
+     * @param errorCausingMilestone the milestone that failed to be applied
+     * @throws MilestoneException if we failed to reset the corrupted milestone
+     */
+    private void revertPrecedingMilestones(MilestoneViewModel errorCausingMilestone) throws MilestoneException {
```
alon-e: "Preceding" can be confusing, is it Preceding in the process or in index #?

alon-e: maybe `outOfOrder`?

src/main/java/com/iota/iri/service/milestone/impl/LatestSolidMilestoneTrackerImpl.java
```diff
@@ -191,4 +243,32 @@ private void logChange(int prevSolidMilestoneIndex) {
...
+
+    /**
+     * Tries to actively repair the ledger by reverting the milestones preceding the given milestone.<br />
+     * <br />
+     * It gets called when a milestone could not be applied to the ledger state because of problems like "inconsistent
+     * balances". While this should theoretically never happen (because milestones are by definition "consistent"), it
+     * can still happen because IRI crashed or got stopped in the middle of applying a milestone or if a milestone
+     * was processed in the wrong order.<br />
+     * <br />
+     * Every time we call this method the internal {@link #repairBackoffCounter} gets increased which causes the next
```
alon-e:
```suggestion
     * Every time we call this method the internal {@link #repairBackoffCounter} is incremented which causes the next
```

src/main/java/com/iota/iri/service/milestone/impl/MilestoneServiceImpl.java
```diff
@@ -151,6 +179,17 @@ public MilestoneValidity validateMilestone(TransactionViewModel transactionViewM
...
 
+                                // if we find a NEW milestone that should have been processed before our latest solid
+                                // milestone -> reset the ledger state and check the milestones again
+                                //
+                                // NOTE: this can happen if a new subtangle becomes solid before a previous one while
+                                //       syncing
+                                if (milestoneIndex < snapshotProvider.getLatestSnapshot().getIndex() &&
+                                        milestoneIndex > snapshotProvider.getInitialSnapshot().getIndex()) {
+
+                                     resetCorruptedMilestone(newMilestoneViewModel.index());
```
alon-e: the use of `newMilestoneViewModel.index()` instead of `milestoneIndex` in this line made me look up the value and make sure they are the same -- it's better to use the same var for better readability.

```suggestion
                                     resetCorruptedMilestone(milestoneIndex);
```

src/main/java/com/iota/iri/service/milestone/impl/MilestoneServiceImpl.java
```diff
@@ -95,6 +111,18 @@ public void updateMilestoneIndexOfMilestoneTransactions(Hash milestoneHash, int
...
+     * processedTransactions} with an empty {@link HashSet} which will ensure that we reset all found
+     * transactions.<br />
+     */
+    @Override
+    public void resetCorruptedMilestone(int index) throws MilestoneException {
+        resetCorruptedMilestone(index, new HashSet<>());
+    }
+
     @Override
     public MilestoneValidity validateMilestone(TransactionViewModel transactionViewModel, SpongeFactory.Mode mode,
```
alon-e: we should consider pulling out the `MilestoneVM.store()` from the validator -- a validator should be stateless.

src/main/java/com/iota/iri/service/milestone/impl/MilestoneServiceImpl.java
```diff
@@ -312,5 +367,41 @@ private boolean isMilestoneBundleStructureValid(List<TransactionViewModel> bundl
...
+         log.info("resetting corrupted milestone #" + index);
+         try {
+            MilestoneViewModel milestoneToRepair = MilestoneViewModel.get(tangle, index);
+            if(milestoneToRepair != null) {
+                if(milestoneToRepair.index() <= snapshotProvider.getLatestSnapshot().getIndex()) {
+                    snapshotService.rollBackMilestones(snapshotProvider.getLatestSnapshot(), milestoneToRepair.index());
+                }
+                 updateMilestoneIndexOfMilestoneTransactions(milestoneToRepair.getHash(), milestoneToRepair.index(), 0,
+                        processedTransactions);
+                tangle.delete(StateDiff.class, milestoneToRepair.getHash());
```
alon-e: what happens if we fail before this line is processed? what will happen when IRI is reset?
(i.e. non-atomic change)

src/main/java/com/iota/iri/service/milestone/MilestoneService.java
```diff
@@ -50,6 +50,22 @@ MilestoneValidity validateMilestone(TransactionViewModel transactionViewModel, S
...
+     * rolls back the ledger state to the moment before the milestone was applied.<br />
+     * <br />
+     * This allows us to reprocess the milestone in case of errors where the given milestone could not be applied to the
+     * ledger state. It is for example used by the automatic repair routine of the {@link LatestSolidMilestoneTracker}
+     * (to recover from inconsistencies due to crashes of IRI).<br />
+     * <br />
+     * It recursively resets additional milestones if inconsistencies are found within the resetted milestone (wrong
+     * {@code milestoneIndex}es).<br />
+     *
+     * @param index milestone index that shall
```
alon-e:
```suggestion
     * @param index milestone index that shall be reverted
``` 
or ?

src/main/java/com/iota/iri/service/milestone/impl/LatestSolidMilestoneTrackerImpl.java
hmoog: Both, since milestones get processed in the order that they are issued.

outOfOrder is wrong - we reset the milestones preceding the errorCausingMilestone - during this process we might identify which milestone was processed out of order (or marked the wrong transactions) but the milestones we reset here are only the ones that were issued before the once that failed to get applied.

src/main/java/com/iota/iri/service/milestone/impl/LatestSolidMilestoneTrackerImpl.java
hmoog: changed it and introduces some additional utility methods for better readability

src/main/java/com/iota/iri/service/milestone/impl/LatestSolidMilestoneTrackerImpl.java
hmoog: changed it and introduces some additional utility methods for better readability

src/main/java/com/iota/iri/service/milestone/impl/LatestSolidMilestoneTrackerImpl.java
hmoog: changed it and introduces some additional utility methods for better readability

src/main/java/com/iota/iri/service/milestone/impl/MilestoneServiceImpl.java
hmoog: That's old legacy code that just got copied to the new class - refactoring everything was not part of the scope of the PR. The whole method can and should be refactored (also in relation to the cyclomatic complexity).

src/main/java/com/iota/iri/service/milestone/impl/MilestoneServiceImpl.java
hmoog: It will lead to an inconsistency which can also happen during the application of milestones (in contrast to reverting them).

At the moment this will be fixed by the repair mechanism. The repair will just be triggered again if the milestone is still inconsistent (after a IRI restart).

src/main/java/com/iota/iri/service/milestone/impl/MilestoneServiceImpl.java
alon-e: sure - that was more of a comment, will open an issue.

src/main/java/com/iota/iri/service/milestone/impl/LatestSolidMilestoneTrackerImpl.java
```diff
@@ -191,4 +265,32 @@ private void logChange(int prevSolidMilestoneIndex) {
...
+     * @throws MilestoneException if we failed to reset the corrupted milestone
+     */
+    private void repairCorruptedMilestone(MilestoneViewModel errorCausingMilestone) throws MilestoneException {
+        if(repairBackoffCounter++ == 0) {
+            errorCausingMilestoneIndex = errorCausingMilestone.index();
+        }
+        for (int i = errorCausingMilestone.index(); i > errorCausingMilestone.index() - repairBackoffCounter; i--) {
+            milestoneService.resetCorruptedMilestone(i);
+        }
+    }
```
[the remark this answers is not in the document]
hmoog: lets say 100 is corrupt because it didnt get its snapshotIndex set correctly and 101 also approves some of the transactions that were taken into account for the balances of 100 already. 101 can therefore not be applied.

We will first reset 101 and try to reapply it - if that fails we reset 101 and 100 and try to reapply both. If that fails we try to reapply 101, 100 and 99 and try to reapply the three of them.

Depending on which transaction didnt get its snapshotIndex correctly set, we might need to go a few milestones back to "find" the one that wasnt processed correctly. sometimes milestones 101 would reference a "broken" tx from milestone 97 for example.

PR #1178, opened 2018-11-21
Introducing a dedicated Transaction Requester Background Worker

hmoog:
# Description

This PR introduces a dedicated background worker that tries to work through the request queue by sending random tips along the requested transactions. This massively increases the sync speed of new nodes that would otherwise be limited to requesting in the same rate as new transactions are received.

Note: To reduce the overhead for the node we only trigger this worker if the request queue gets bigger than the "REQUESTER_THREAD_ACTIVATION_THRESHOLD". Otherwise we rely on the processing of the queue due to normal outgoing traffic like transactions that get relayed by our node.

## Type of change

# Checklist:


GalRogozinski:
The fix to the broadcast test is not on this branch so it still fails on buildkite

PR #1184, opened 2018-11-22
Hotfix for the solidification bug of the Local Snapshots branch

hmoog:
# Description

While refactoring the code we introduced a small but severe bug where the generateBalanceDiff method of the LedgerService traversed one milestone too far, booking balances twice.

In addition it fixes the shutdown() order (IRI crashed before - not being able to shutdown all instances in time because some of the legacy shutdown-methods take a really long time to terminate). The shutdown-methods of the new classes terminate instantly.

## Type of change

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/service/ledger/impl/LedgerServiceImpl.java
+++ b/src/main/java/com/iota/iri/service/ledger/impl/LedgerServiceImpl.java
@@ -115,8 +115,7 @@ public class LedgerServiceImpl implements LedgerService {
             return true;
         }
         Set<Hash> visitedHashes = new HashSet<>(approvedHashes);
-        Map<Hash, Long> currentState = generateBalanceDiff(visitedHashes, tip,
-                snapshotProvider.getLatestSnapshot().getIndex());
+        Map<Hash, Long> currentState = generateBalanceDiff(visitedHashes, tip);
         if (currentState == null) {
             return false;
         }
@@ -135,8 +134,8 @@ public class LedgerServiceImpl implements LedgerService {
     }
 
     @Override
-    public Map<Hash, Long> generateBalanceDiff(Set<Hash> visitedTransactions, Hash milestoneHash,
-            int latestSolidMilestoneIndex) throws LedgerException {
+    public Map<Hash, Long> generateBalanceDiff(Set<Hash> visitedTransactions, Hash transactionHash) throws
+            LedgerException {
 
         Map<Hash, Long> state = new HashMap<>();
         Set<Hash> countedTx = new HashSet<>();
@@ -146,17 +145,16 @@ public class LedgerServiceImpl implements LedgerService {
             countedTx.add(solidEntryPointHash);
         });
 
-        final Queue<Hash> nonAnalyzedTransactions = new LinkedList<>(Collections.singleton(milestoneHash));
+        final Queue<Hash> nonAnalyzedTransactions = new LinkedList<>(Collections.singleton(transactionHash));
         Hash transactionPointer;
         while ((transactionPointer = nonAnalyzedTransactions.poll()) != null) {
             if (visitedTransactions.add(transactionPointer)) {
                 try {
                     final TransactionViewModel transactionViewModel = TransactionViewModel.fromHash(tangle,
                             transactionPointer);
-
-                    if (milestoneService.transactionBelongsToMilestone(transactionViewModel,
-                            latestSolidMilestoneIndex + 1)) {
-
+                    // only take transactions into account that have not been confirmed, yet (that are not included in
+                    // the ledger state)
+                    if (!milestoneService.isTransactionConfirmed(transactionViewModel)) {
                         if (transactionViewModel.getType() == TransactionViewModel.PREFILLED_SLOT) {
                             return null;
                         } else {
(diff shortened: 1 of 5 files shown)


src/main/java/com/iota/iri/service/ledger/impl/LedgerServiceImpl.java
```diff
@@ -150,12 +150,12 @@ public boolean isBalanceDiffConsistent(Set<Hash> approvedHashes, Map<Hash, Long>
...
         while ((transactionPointer = nonAnalyzedTransactions.poll()) != null) {
             if (visitedTransactions.add(transactionPointer)) {
-
                 try {
                     final TransactionViewModel transactionViewModel = TransactionViewModel.fromHash(tangle,
                             transactionPointer);
 
-                    if (milestoneService.transactionBelongsToMilestone(transactionViewModel, milestoneIndex)) {
+                    if (milestoneService.transactionBelongsToMilestone(transactionViewModel,
+                            latestSolidMilestoneIndex + 1)) {
```
GalRogozinski: I just have a little question:
In `MilestoneServiceImpl`:
```
 @Override
    public boolean transactionBelongsToMilestone(TransactionViewModel transaction, int milestoneIndex) {
        return transaction.snapshotIndex() == 0 || transaction.snapshotIndex() >= milestoneIndex;
    }
```

If transaction belongs to a milestone shouldn't it be  `transaction.snapshotIndex() <= milestoneIndex`?

GalRogozinski: I think what you want to do after you fix `transactionBelongsToMilestone` is do:
`if (!milestoneService.transactionBelongsToMilestone(transactionViewModel,
                            latestSolidMilestoneIndex))` (note the **!**)

hmoog: Nope its not <= milestoneIndex because that would address ALL transactions that happened before the given index.

Lets say I want to retrieve all TX that belong to the milestone with index 10. Your equasion would return true for all tx with snapshotIndex 10, 9, 8, 7, 6, 5 ,4 ,3, 2 ,1 ... but we only want the ones that belong to exactly 10.

What we do here is pretty much the following.

Transactions that belong to milestone 10 are the ones that:

1. either have not been marked by a previous milestones yet ( == 0)
2. already carry the correct snapshot index ( == 10)
3. have mistakenly been marked by a future milestone ( > 10)

This can then be merged to == 0 || >= 10

This is utility method is only used while iterating the approovees from a milestone and it therefore needs to cover these 3 cases.

GalRogozinski: Wait one second... In the old `LedgerValidator` we could have start traversing from an area that is non confirmed if I am not mistaken and calculate the balance (`getLatestDiff`). How do we do it now?

In the old version condition (3) is not accurate. In the new `LedgerService` it is since you start traversing from a `milestoneHash`.

GalRogozinski: I see actually now. `milestoneHash` is a misleading name.

GalRogozinski: Your logic works the problem is that definition (3) is not exactly correct in `LedgerServiceImpl`.
It is correct the way you use it in `MilestoneServiceImpl`...

We need to better define what this method does. In `LedgerServiceImpl` it is possible that you have a transaction that approved (indirectly) the latest milestone and in the meanwhile another thread assigned it a larger snapshot value.

src/main/java/com/iota/iri/service/ledger/impl/LedgerServiceImpl.java
```diff
@@ -146,17 +145,16 @@ public boolean isBalanceDiffConsistent(Set<Hash> approvedHashes, Map<Hash, Long>
...
-
                 try {
                     final TransactionViewModel transactionViewModel = TransactionViewModel.fromHash(tangle,
                             transactionPointer);
-
-                    if (milestoneService.transactionBelongsToMilestone(transactionViewModel, milestoneIndex)) {
-
+                    // only take transactions into account that have not been confirmed, yet (that are not included in
+                    // the ledger state)
+                    if (!milestoneService.isTransactionConfirmed(transactionViewModel)) {
```
GalRogozinski: I think we overlooked something :-(
This code is different than before.

Before you were fed from the parameters a `milestoneIndex` that could never change.
Now it can change in the background...

hmoog: it was always called with snapshotProvider.getLatestSnapshot().getIndex() because thats what the method was for - generate the stateDiff compared to the confirmed part of the tangle.

We now fix this value in the little wrapper

    @Override
    public boolean isTransactionConfirmed(TransactionViewModel transaction) {
        return isTransactionConfirmed(transaction, snapshotProvider.getLatestSnapshot().getIndex());
    }

It doesnt matter if the wrapper fixes the value or the caller. The underlying isTransactionConfirmed(TransactionViewModel, milestoneIndex) will work with a fixed value that can not magically change anymore

hmoog: Now I got you -> fixed 💃

PR #1186, opened 2018-11-22
make storeMessage store transactions

### LINKED ISSUE #1185: `storeMessageStatement` actually only broadcasts - it does not store
### Bug description
the API call `storeMessageStatement` actually only broadcasts - it does not store. despite it's name

### Steps To Reproduce
1. run an isolated node w/ testnet flag
2. call storeMessageStatement
3. see that no transaction is stored to the DB

### Expected behaviour
store :)
### Actual behaviour
broadcast

alon-e:
# Description

Fixes #1185

## Type of change

# How Has This Been Tested?

- running an isolated node
- running a local network

### DIFF
--- a/src/main/java/com/iota/iri/service/API.java
+++ b/src/main/java/com/iota/iri/service/API.java
@@ -1727,6 +1727,7 @@ public class API {
     }
 
    /**
+     *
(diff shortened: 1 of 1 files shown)


alon-e:
done.

PR #1198, opened 2018-11-26
Perf: Limit the amount of milestones that need to be analyzed when generating the solid entry points

hmoog:
# Description

This PR introduces a method to limit the milestones that have to be analyzed when generating the solid entry points.

Since the snapshot creation is a "continuous" process where we build upon the information gathered during the creation of previous snapshots, we do not need to analyze all previous milestones but can rely on slowly gathering the missing information over time. While this might lead to a situation where the very first snapshots taken by a node might generate snapshot files that can not reliably be used by other nodes to sync it is still a reasonable trade-off to reduce the load on the nodes. We just assume that anybody who wants to share his snapshots with the community as a way to bootstrap new nodes will run his snapshot enabled node for a few hours before sharing his files (this is a problem in very rare edge cases when having back-referencing transactions anyway).

It massively decreases the required time to migrate from a non-snapshotted database to a snapshotted one.

## Type of change

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/service/snapshot/impl/SnapshotServiceImpl.java
+++ b/src/main/java/com/iota/iri/service/snapshot/impl/SnapshotServiceImpl.java
@@ -43,6 +43,21 @@ public class SnapshotServiceImpl implements SnapshotService {
     private static final Logger log = LoggerFactory.getLogger(SnapshotServiceImpl.class);
 
     /**
+     * Holds a limit for the amount of milestones we go back in time when generating the solid entry points (to speed up
+     * the snapshot creation).<br />
+     * <br />
+     * Note: Since the snapshot creation is a "continuous" process where we build upon the information gathered during
+     *       the creation of previous snapshots, we do not need to analyze all previous milestones but can rely on
+     *       slowly gathering the missing information over time. While this might lead to a situation where the very
+     *       first snapshots taken by a node might generate snapshot files that can not reliably be used by other nodes
+     *       to sync it is still a reasonable trade-off to reduce the load on the nodes. We just assume that anybody who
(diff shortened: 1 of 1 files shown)


src/main/java/com/iota/iri/service/snapshot/impl/SnapshotServiceImpl.java
```diff
@@ -42,6 +42,21 @@
...
+     * Note: Since the snapshot creation is a "continuous" process where we build upon the information gathered during
+     *       the creation of previous snapshots, we do not need to analyze all previous milestones but can rely on
+     *       slowly gathering the missing information over time. While this might lead to a situation where the very
+     *       first snapshots taken by a node might generate snapshot files that can not reliably be used by other nodes
+     *       to sync it is still a reasonable trade-off to reduce the load on the nodes. We just assume that anybody who
+     *       wants to share his snapshots with the community as a way to bootstrap new nodes will run his snapshot
+     *       enabled node for a few hours before sharing his files (this is a problem in very rare edge cases when
+     *       having back-referencing transactions anyway).<br />
+     */
+    private static final int OUTER_SHELL_SIZE = 100;
```
GalRogozinski: Maybe this should be a configuration?

GalRogozinski: Actually whether it should be a configured value or not it is a product question so @jakubcech should be involved.

jakubcech: I don't think this should be made configurable from the get-go. There's already quite a lot to configure and wrap your brain around with different intervals and limits, and this seems like a reasonable thing to have set to a default value and possibly revisit in the future if we see a need for configurability.

PR #1199, opened 2018-11-28
Refactor: Removed grace time from solid entry points generation

hmoog:
# Description

This PR removes the grace time from the solid entry points generation. It was added in the past to work around the back referencing transaction issue. Since the issue can not be solved by just keeping the solid entry points alive for a little longer time and needs to be addressed separately (we will use cuckoo filters for this later on), it can and should be removed.

Especially in scenarios where we have a lot of TPS and a side tangle (like it happened today) it can slow the snapshot generation noticable and disrupt the operation of the node.

Note: This PR also changes the info message when the sendqueue is full to being a debug message because it can otherwise spam the CLI output making it very hard to debug problems and see important messages of the console output.

## Type of change

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/network/TCPNeighbor.java
+++ b/src/main/java/com/iota/iri/network/TCPNeighbor.java
@@ -14,12 +14,12 @@ import java.util.concurrent.TimeUnit;
 /**
  * Created by paul on 4/15/17.
  */
- 
+
  /**
- * This class Extends {@link Neighbor} base class with TCP specific functionality. 
+ * This class Extends {@link Neighbor} base class with TCP specific functionality.
  * It keeps reference of Source and Sink while maintaining a sendQueue for keeping
- * outgoing requests. 
- * 
+ * outgoing requests.
+ *
  */
 public class TCPNeighbor extends Neighbor {
     private static final Logger log = LoggerFactory.getLogger(Neighbor.class);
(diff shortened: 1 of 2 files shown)


src/main/java/com/iota/iri/network/TCPNeighbor.java
```diff
@@ -93,7 +93,7 @@ public void send(DatagramPacket packet) {
         synchronized (sendQueue) {
             if (sendQueue.remainingCapacity() == 0) {
                 sendQueue.poll();
-                log.info("Sendqueue full...dropped 1 tx");
+                log.debug("Sendqueue full...dropped 1 tx");
```
GalRogozinski: I found this info statement very useful while trying to peer nodes and noticing right away that there is a problem.
I agree it is very annoying the way it can flood the console.

Maybe use the `IntervalLogger`?

GalRogozinski:
>Since the issue can not be solved by just keeping the solid entry points alive for a little longer time and needs to be addressed separately (we will use cuckoo filters for this later on), it can and should be removed

Maybe we do this change when we start applying the cuckoo filter?

src/main/java/com/iota/iri/network/TCPNeighbor.java
hmoog: We already have the connection message for this case. In case a node drops the connection or doesnt accept packets (like with NELSON) you will not see anything in your cli.

PR #1203, opened 2018-11-29
Fixed #541: Will add configuration parameter for remote trusted api hosts

### LINKED ISSUE #541: add trusted host config field
Hello everyone

it could be nice have a new config field to pass a trusted host for remote api

For example:
--remote-api-trusted mynode.iota.org

legacycode:
# Description

Will add remote trusted api hosts with following configuration parameter. You can specify an ip or dns address in a comma separated string:

`--remote --remote-trusted-api-hosts 10.0.0.1,www.iota.org`

Default, if parameter is not provided: [localhost/127.0.0.1]

Fixes #541 

## Type of change

# How Has This Been Tested?

## First test:

1. Start iri with following parameter: `--testnet -p 14265 --remote`
2. Visit api url from another host ip address
3. Create a post request with body `{"command": "addNeighbors", "uris": ["udp://8.8.8.8:14265", "udp://8.8.8.5:14265"]}` and send it

Expected result: 401 Unauthorized
```
{
"error": "COMMAND addNeighbors is not available on this node",
"duration": 16
}
```

## Second test:

1. Start iri with following parameter: `--testnet -p 14265 --remote --remote-trusted-api-hosts 10.0.0.10`
2. Replace `10.0.0.10` with your client hosts ip address
3. Visit api url from another host ip address
4. Create a post request with body `{"command": "addNeighbors", "uris": ["udp://8.8.8.8:14265", "udp://8.8.8.5:14265"]}` and send it

Expected result: 200 OK
```
{
"addedNeighbors": 2,
"duration": 20
}

```
# Checklist:


src/main/java/com/iota/iri/IRI.java
```diff
@@ -109,6 +109,9 @@ private static void configureLogging() {
         public static void main(String [] args) throws Exception {
             IotaConfig config = createConfiguration(args);
             log.info("Welcome to {} {}", config.isTestnet() ? TESTNET_NAME : MAINNET_NAME, VERSION);
+            config.getRemoteTrustedApiHosts().forEach(h->{
```
legacycode: I added the logging output to be sure the hosts are parsed correctly.

GalRogozinski:
Thanks,
All non local snapshot PRs will be handled and merged after Christmas.

src/main/java/com/iota/iri/conf/BaseIotaConfig.java
```diff
@@ -695,6 +721,8 @@ protected void setPowThreads(int powThreads) {
         int API_PORT = 14265;
         String API_HOST = "localhost";
         List<String> REMOTE_LIMIT_API = IotaUtils.createImmutableList("addNeighbors", "getNeighbors", "removeNeighbors", "attachToTangle", "interruptAttachingToTangle");
+        InetAddress REMOTE_LIMIT_API_DEFAULT_HOST = InetAddress.getLoopbackAddress();
+        List<InetAddress> REMOTE_LIMIT_API_HOSTS = Arrays.asList(REMOTE_LIMIT_API_DEFAULT_HOST);
```
GalRogozinski: Use `IotaUtils.createImmutableList()` please

This is a change request

src/main/java/com/iota/iri/conf/converters/InetAddressConverter.java
```diff
@@ -0,0 +1,28 @@
...
+    @Override
+    public InetAddress convert(String address) {
+        try {
+            return InetAddress.getByName(address.trim());
+           } catch (UnknownHostException e) {
+            log.error("Invalid value ({}) for paramater: {}", address.trim(), e.getMessage());
+        }
+        return null;
+    }
+}
```
GalRogozinski: Hmm,

When you use JCommander without setters you will need this class.
Why not just put `InetAddress.getByName(address.trim())` in the setter?
This is just more boiler plate code... So please remove :-)

You can also rethrow the `UnknowHostException` as a `ParameterException`.
If the user enters a bad value it was probably by mistake and it is best for IRI not to start to give the user the opportunity to correct his mistake.

By the way the reason we are using setter is because in the future I think that most configs should have some validation/conversion and it nicer to do in the setter then in an external class. Of course this is my opinion and you can disagree with me on that...

src/test/java/com/iota/iri/conf/ConfigTest.java
```diff
@@ -199,6 +212,13 @@ public void testIniParsingMainnet() throws Exception {
         Assert.assertEquals("PORT", 17000, iotaConfig.getPort());
         Assert.assertEquals("NEIGHBORS", Arrays.asList("udp://neighbor1", "neighbor", "tcp://neighbor2"),
                 iotaConfig.getNeighbors());
+
+        List<InetAddress> expectedTrustedApiHosts = new ArrayList<>();
+        expectedTrustedApiHosts.add(InetAddress.getByName("192.168.0.55"));
+        expectedTrustedApiHosts.add(InetAddress.getByName("10.0.0.10"));
+        expectedTrustedApiHosts.add(BaseIotaConfig.Defaults.REMOTE_LIMIT_API_DEFAULT_HOST);
+        Assert.assertEquals("REMOTE_TRUSTED_API_HOSTS", expectedTrustedApiHosts, iotaConfig.getRemoteTrustedApiHosts());
+
```
GalRogozinski: :+1: For adding unit tests :-)

I think you can maybe create the list in one line with `IotaUtils` or `Arrays.asList()`.
However, if you think your way is neater then you don't have to change :-)

GalRogozinski:
@legacycode 
You have a conflict so I can't merge

src/main/java/com/iota/iri/conf/BaseIotaConfig.java
```diff
@@ -162,6 +167,27 @@ protected void setRemoteLimitApi(String remoteLimitApi) {
...
+
+    @JsonProperty
+    @Parameter(names = {"--remote-trusted-api-hosts"}, description = APIConfig.Descriptions.REMOTE_TRUSTED_API_HOSTS)
+    public void setRemoteTrustedApiHosts(String remoteTrustedApiHosts) {
+        InetAddressConverter converter = new InetAddressConverter();
+        List<String> addresses = IotaUtils.splitStringToImmutableList(remoteTrustedApiHosts, SPLIT_STRING_TO_LIST_REGEX);
+        List<InetAddress> inetAddresses = new ArrayList<>();
+        for (String address : addresses) {
+            inetAddresses.add(converter.convert(address));
+        }
```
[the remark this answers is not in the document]
legacycode: Good idea, done.

src/main/java/com/iota/iri/conf/BaseIotaConfig.java
legacycode: Good idea, done.

src/main/java/com/iota/iri/conf/converters/InetAddressConverter.java
GalRogozinski: > In my opinion BaseIotaConfig.java should only be an entity class and validation and conversion logic should be moved into JCommander converters and validators.

I don't feel like it is such a pain. It is fine to add logic in setters. You can also gain re-usability with private methods.
You can claim that the `BaseIotaConfig` class is getting very big so it may be better to do it your way (and you have a good point). I don't see it as much of a priority to be honest or a big issue.

>  also would like to drop/move the support of Jackson configuration file and migrate to JCommander configuration files provided by the @syntax

I agree that if we can drop a dependency then it is a good idea

PR #1204, opened 2018-11-29
Increase rescan performance of old milestones after IRI restart

hmoog:
# Description

IRI always scans all potential milestone candidates after a restart. Depending on the database size this can take a really long time (hours in the worst case). If we have a local snapshot we can always immediately discard the transactions prior to the snapshot point because the node doesn't need them anymore. This increases the restart performance of IRI massively (few seconds).

## Type of change

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/service/milestone/impl/LatestMilestoneTrackerImpl.java
+++ b/src/main/java/com/iota/iri/service/milestone/impl/LatestMilestoneTrackerImpl.java
@@ -11,7 +11,6 @@ import com.iota.iri.service.milestone.LatestMilestoneTracker;
 import com.iota.iri.service.milestone.MilestoneException;
 import com.iota.iri.service.milestone.MilestoneService;
 import com.iota.iri.service.milestone.MilestoneSolidifier;
-import com.iota.iri.service.milestone.MilestoneValidity;
 import com.iota.iri.service.snapshot.Snapshot;
 import com.iota.iri.service.snapshot.SnapshotProvider;
 import com.iota.iri.storage.Tangle;
@@ -25,9 +24,6 @@ import java.util.Deque;
 import java.util.HashSet;
 import java.util.Set;
 import java.util.concurrent.TimeUnit;
-
-import static com.iota.iri.service.milestone.MilestoneValidity.INCOMPLETE;
-import static com.iota.iri.service.milestone.MilestoneValidity.INVALID;
 
 /**
  * Creates a tracker that automatically detects new milestones by incorporating a background worker that periodically
@@ -187,9 +183,9 @@ public class LatestMilestoneTrackerImpl implements LatestMilestoneTracker {
     }
 
     @Override
-    public MilestoneValidity analyzeMilestoneCandidate(Hash transactionHash) throws MilestoneException {
+    public boolean processMilestoneCandidate(Hash transactionHash) throws MilestoneException {
         try {
-            return analyzeMilestoneCandidate(TransactionViewModel.fromHash(tangle, transactionHash));
+            return processMilestoneCandidate(TransactionViewModel.fromHash(tangle, transactionHash));
         } catch (Exception e) {
             throw new MilestoneException("unexpected error while analyzing the transaction " + transactionHash, e);
         }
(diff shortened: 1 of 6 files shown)


src/main/java/com/iota/iri/service/milestone/impl/LatestMilestoneTrackerImpl.java
```diff
@@ -209,8 +208,13 @@ public MilestoneValidity analyzeMilestoneCandidate(TransactionViewModel transact
                     transaction.getCurrentIndex() == 0) {
 
                 int milestoneIndex = milestoneService.getMilestoneIndex(transaction);
+                if (milestoneIndex <= snapshotProvider.getInitialSnapshot().getIndex()) {
+                    return INVALID;
+                }
```
GalRogozinski: I think Alon's beef was that the `validateMilestone` depended on the milestone Index. On which he was correct, we don't want other potential callers of this method to be blocked by the index.

 He didn't say anything about your `IRRELEVANT` enum. I think that returning an INVALID on a vaild milestone will be very misleading to anyone in the future (it is at least misleading for me). I would bring back this enum value.

GalRogozinski: Actually I have a better idea... Instead of adding another enum to validity (IRRELEVANT) doesn't make much sense, `analyzeMilestoneCandidate` can return a boolean: `true` if successfully analyzed, `false` otherwise.

If the `milestoneIndex` is below the snapshot then return `true` and add a little comment: "has already been analyzed if it is below the snapshot" (feel free to change the wording).

I think this makes the most sense.

alon-e: I agree w/ @GalRogozinski that returning `INVALID` may not be the best answer. other than that lgtm.

hmoog: i changed it accordingly

src/main/java/com/iota/iri/service/milestone/impl/LatestMilestoneTrackerImpl.java
```diff
@@ -203,14 +198,19 @@ public MilestoneValidity analyzeMilestoneCandidate(Hash transactionHash) throws
...
 
                 int milestoneIndex = milestoneService.getMilestoneIndex(transaction);
 
-                switch (milestoneService.validateMilestone(transaction, SpongeFactory.Mode.CURLP27, 1)) {
+                // if the milestone is older than our ledger start point: we already processed it in the past
+                if (milestoneIndex <= snapshotProvider.getInitialSnapshot().getIndex()) {
+                    return true;
+                }
+
+                switch (milestoneService.validateMilestone(transaction, milestoneIndex, SpongeFactory.Mode.CURLP27, 1)) {
```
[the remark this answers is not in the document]
GalRogozinski: Let's try not to ignore codacy.

Even thought the way you did it is less verbose, I think it is still clearer and more readable to have the default case inside the switch brackets. You can also add a little comment:
`//We can consider the milestone candidate processed and move on w/o farther action`

src/main/java/com/iota/iri/service/milestone/impl/LatestMilestoneTrackerImpl.java
```diff
@@ -203,14 +198,19 @@ public MilestoneValidity analyzeMilestoneCandidate(Hash transactionHash) throws
      * {@link MilestoneSolidifier} that takes care of requesting the missing parts of the milestone bundle.<br />
      */
     @Override
-    public MilestoneValidity analyzeMilestoneCandidate(TransactionViewModel transaction) throws MilestoneException {
+    public boolean processMilestoneCandidate(TransactionViewModel transaction) throws MilestoneException {
```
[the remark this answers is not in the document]
GalRogozinski: :+1:  on the name change

GalRogozinski: +1 on the name change

src/main/java/com/iota/iri/service/milestone/impl/LatestMilestoneTrackerImpl.java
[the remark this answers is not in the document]
hmoog: i think this is really unnecessary here since the switch is used as a faster / easier to read version of an if then else if then but if you think it helps i can add an empty default with that comment .. :P

PR #1228, opened 2018-12-10
Feat: Disable snapshot logic completely if disabled in the config

hmoog:
# Description

This PR completely disables the creation of the snapshot related managers and classes if the features are disabled in the config.

## Type of change

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/service/snapshot/impl/SnapshotServiceImpl.java
+++ b/src/main/java/com/iota/iri/service/snapshot/impl/SnapshotServiceImpl.java
@@ -433,7 +433,7 @@ public class SnapshotServiceImpl implements SnapshotService {
                 try {
                     // only clean up if the corresponding milestone transaction was cleaned up already -> otherwise
                     // let the MilestonePrunerJob do this
-                    if (TransactionViewModel.fromHash(tangle, transactionHash).getType() ==
+                    if (transactionPruner != null && TransactionViewModel.fromHash(tangle, transactionHash).getType() ==
                             TransactionViewModel.PREFILLED_SLOT) {
 
                         transactionPruner.addJob(new UnconfirmedSubtanglePrunerJob(transactionHash));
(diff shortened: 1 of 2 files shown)


src/main/java/com/iota/iri/service/snapshot/impl/SnapshotServiceImpl.java
```diff
@@ -433,7 +433,7 @@ private void cleanupExpiredSolidEntryPoints(Tangle tangle, Map<Hash, Integer> ol
                 try {
                     // only clean up if the corresponding milestone transaction was cleaned up already -> otherwise
                     // let the MilestonePrunerJob do this
-                    if (TransactionViewModel.fromHash(tangle, transactionHash).getType() ==
+                    if (transactionPruner != null && TransactionViewModel.fromHash(tangle, transactionHash).getType() ==
```
GalRogozinski: There is no point doing the null check here
You can do it right at the beginning of the method...

src/main/java/com/iota/iri/service/snapshot/impl/SnapshotServiceImpl.java
```diff
@@ -464,7 +464,7 @@ private void cleanupOldData(SnapshotConfig config, TransactionPruner transaction
         int startingIndex = config.getMilestoneStartIndex() + 1;
 
         try {
-            if (targetIndex >= startingIndex) {
+            if (transactionPruner != null && targetIndex >= startingIndex) {
```
GalRogozinski: ditto

src/main/java/com/iota/iri/service/snapshot/impl/SnapshotServiceImpl.java
hmoog: changed it

src/main/java/com/iota/iri/service/snapshot/impl/SnapshotServiceImpl.java
hmoog: changed it

PR #1234, opened 2018-12-12
IRI can read global snapshot files from current working directory

GalRogozinski:
# Description

In order to share global state more easily with the community without performing a release, we want to be able to load snapshot data from the current working directory.
With the changes here we will first search for files in the current working directory before we try to load them from the jar.

This current PR itself won't suffice because we still need to load metadata about the signature (i.e. merkle tree leaf used). A later PR will solve this.

## Type of change

# How Has This Been Tested?

Was ran from jar locally and tested with files in current working directory.

****


GalRogozinski:
The PR has failed because of how testnet loads snapshot files.
The solution will be another PR in which we load the snapshot files to mainnet in a similar way to which we load them to testnet.

GalRogozinski:
In the end apparently it was best to add the fix to the current PR

Issue #1229, opened 2018-12-10
getRandomSolidTipHash returns non-solid tip


alongalky:
### Bug description
The `TipsViewModel::getRandomSolidTipHash` function returns a non-solid tip if a solid one is not found. This is probably a bug.

### Code
https://github.com/iotaledger/iri/blob/5883633a06312602c4a2439906d7ade49ed7f2f4/src/main/java/com/iota/iri/controllers/TipsViewModel.java#L99-L104

### Expected behaviour
`getRandomSolidTipHash` should return a solid tip or fail.

PR #1235, opened 2018-12-13
Fix: Node requested old transactions forever

hmoog:
# Description

Because of a bug in the checkSolidity function and a problem with the bottom->top solidifier (missing cuckoo filter) the node started requesting very old transactions and never stopped which at some point overwhelmed the requester (and increased the resource consumption).

## Type of change

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/TransactionValidator.java
+++ b/src/main/java/com/iota/iri/TransactionValidator.java
@@ -260,8 +260,8 @@ public class TransactionValidator {
                 }
 
                 final TransactionViewModel transaction = fromHash(tangle, hashPointer);
-                if(!transaction.isSolid()) {
-                    if (transaction.getType() == PREFILLED_SLOT && !snapshotProvider.getInitialSnapshot().hasSolidEntryPoint(hashPointer)) {
+                if(!transaction.isSolid() && !snapshotProvider.getInitialSnapshot().hasSolidEntryPoint(hashPointer)) {
+                    if (transaction.getType() == PREFILLED_SLOT) {
                         solid = false;
 
                         if (!transactionRequester.isTransactionRequested(hashPointer, milestone)) {
@@ -436,12 +436,14 @@ public class TransactionValidator {
      * @throws Exception if we encounter an error while requesting a transaction
      */
     private boolean checkApproovee(TransactionViewModel approovee) throws Exception {
-        if(approovee.getType() == PREFILLED_SLOT) {
-            transactionRequester.requestTransaction(approovee.getHash(), false);
-            return false;
-        }
         if(snapshotProvider.getInitialSnapshot().hasSolidEntryPoint(approovee.getHash())) {
             return true;
+        }
+        if(approovee.getType() == PREFILLED_SLOT) {
+            // don't solidify from the bottom until cuckoo filters can identify where we deleted -> otherwise we will
+            // continue requesting old transactions forever
+            //transactionRequester.requestTransaction(approovee.getHash(), false);
+            return false;
         }
         return approovee.isSolid();
     }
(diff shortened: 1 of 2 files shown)


src/main/java/com/iota/iri/network/impl/TransactionRequesterWorkerImpl.java
```diff
@@ -147,6 +147,9 @@ public void shutdown() {
      */
     private TransactionViewModel getTransactionToSendWithRequest() throws Exception {
         Hash tip = tipsViewModel.getRandomSolidTipHash();
+        if (tip == null) {
+            tip = tipsViewModel.getRandomNonSolidTipHash();
+        }
```
GalRogozinski: Why did you create this change?
You think that due to local snapshots, tips that should be solid won't solidify?

I think usually we don't want to send "spam" tips that approve non-existent transactions.

hmoog: Fix: allow the transactionRequester to also send non-solid tips

Allow the transactionRequester to also send non-solid tips - otherwise we might have problems getting the tips solid (if we can not send any requests).

alon-e: @hmoog this should not make a difference, as `getRandomSolidTipHash` will call `getRandomNonSolidTipHash` if it has no tips
see: https://github.com/iotaledger/iri/issues/1229

src/main/java/com/iota/iri/TransactionValidator.java
```diff
@@ -436,13 +436,15 @@ private boolean quickSetSolid(final TransactionViewModel transactionViewModel) t
...
-            transactionRequester.requestTransaction(approovee.getHash(), false);
-            return false;
-        }
         if(snapshotProvider.getInitialSnapshot().hasSolidEntryPoint(approovee.getHash())) {
             return true;
         }
+        if(approovee.getType() == PREFILLED_SLOT) {
+            // don't solidify from the bottom until cuckoo filters can identify where we deleted -> otherwise we will
+            // continue requesting old transactions forever
+            //transactionRequester.requestTransaction(approovee.getHash(), false);
```
alon-e: this change was made to prevent false positives - transactions that we think are missing, but have deliberately been deleted.

but it seems this will create false negatives - where actual missing transactions (small holes) will not be requested and thought of as deleted.

hmoog: That is correct @alon-e but they will be requested by the checkSolidity calls from the top. So as long as we don't have a way to determine where to stop the solidification (the cuckoo filters PR prepared and merged), this will break nodes that have local snapshots enabled. If we solidify from the top we stop whenever we reach a solid entry point.

But if we happen to receive a transaction that is "below" a solid entry point - this logic will continue requesting everything below it until we reach the genesis.

src/main/java/com/iota/iri/network/impl/TransactionRequesterWorkerImpl.java
hmoog: Oh, you are right it shouldn't make a difference since getRandomSolidTipHash() has an implementation that doesnt correctly reflects its name. Given the fact that we already have an issue to "fix" this behavior i would anyway keep this change in place to make it still work once getRandomSolidTipHash gets fixed.

src/main/java/com/iota/iri/TransactionValidator.java
```diff
@@ -436,13 +436,15 @@ private boolean quickSetSolid(final TransactionViewModel transactionViewModel) t
      * @throws Exception if we encounter an error while requesting a transaction
      */
     private boolean checkApproovee(TransactionViewModel approovee) throws Exception {
-        if(approovee.getType() == PREFILLED_SLOT) {
```
alon-e: this order swap! perfect -- the NULL_HASH also returns `PREFILLED_SLOT` so up until this fix it would not mark TXs pointing to Genesis as solid.

PR #1159, opened 2018-11-18
Document configurations

GalRogozinski:
# Description

Document configurations

## Type of change

- Documentation Fix

Issue #930, opened 2018-08-12
Boolean configuration flags have undefined behavior


GalRogozinski:
### Bug description
This is a follow up on #910.
The following boolean flags that are *true by default* and have to be set to false like so:
`--dns-refresher false`
`--dns-resolution false`

However other flags that are *false by default* are set to true w.o specifying value, e.g.
`--rescan`

### Expected behavior
Either use a disabling flag:
`--disable-dns-refresher` 

**Or enforce passing a value:**
 `--rescan true`

Edit:
We need to give a good error message if we go with option num. 2 and if the flag is used without a passing value. 

We need to understand what and how we break things and document it properly.

jakubcech:
I'm for enforcing passing a value

jakubcech:
Check whether this breaks the docker file.

jakubcech:
#1047

karimodm:
For what solution approach are we settling on? I am prone to avoid any ambiguity and always go for arity >0, "enforce passing a value: --rescan true" approach.

jakubcech:
Yes, we agreed on arity > 0 aka enforce passing a value. :)

jakubcech:
This still in review or were there changes to be made here @kwek20?

GalRogozinski:
We still need to fix aloe tests and travis.
I want to fix aloe tests first, and then when it passes we will fix Travis

kwek20:
@GalRogozinski All fixed, see the linked PRs in #1295

PR #1243, opened 2018-12-18
SolidificationConfig for the tipsolidifier

### LINKED ISSUE #1242: Make snapshot solidifier configurable
### Description
We need a configuration for the tipsolidifier thread. It allows you to completely disable the tipsolidifier thread. 

We should only create the tip solidifier if the configuration for it is set to enabled. 

### Motivation
We need users to be able to control the load on the node, as the tip solidifier can take up quite a lot of resources, taking down the node in edge cases. 

### Requirements
- [ ] A user is able to enable a or disable a new configuration option. 
- [ ] When the option is set to `false`, the `tipSolidifier` isn't started on the node. 
- [ ] When the option is set to `true`, the `tipSolidifier` is started on the node.
- [ ] The default option is `true` - `tipSolidifier` enabled.

kwek20:
# Description
This PR adds the ability for node owners to disable tip solidification.
The reason behind it is that it potentially takes a lot of resources, so you should be able to disable it.
 If it is enabled, it will now show up in the `getNodeInfo['features']` list as "tipSolidification".

Fixes #1242 

## Type of change

# How Has This Been Tested?
- Unit tests
- getNodeInfo, correctly enabled/disabled in features list

### DIFF
--- a/src/main/java/com/iota/iri/Iota.java
+++ b/src/main/java/com/iota/iri/Iota.java
@@ -152,7 +152,7 @@ public class Iota {
                 latestMilestoneTracker, messageQ, configuration);
         replicator = new Replicator(node, configuration);
         udpReceiver = new UDPReceiver(node, configuration);
-        tipsSolidifier = new TipsSolidifier(tangle, transactionValidator, tipsViewModel);
+        tipsSolidifier = new TipsSolidifier(tangle, transactionValidator, tipsViewModel, configuration);
         tipsSelector = createTipSelector(configuration);
 
         injectDependencies();
--- a/src/main/java/com/iota/iri/conf/BaseIotaConfig.java
+++ b/src/main/java/com/iota/iri/conf/BaseIotaConfig.java
@@ -710,11 +710,6 @@ public abstract class BaseIotaConfig implements IotaConfig {
     protected void setTipSolidifierEnabled(boolean tipSolidifierEnabled) {
         this.tipSolidifierEnabled = tipSolidifierEnabled;
     }
-
-    @Parameter(names = "--disable-tip-solidifier", description = SolidificationConfig.Descriptions.TIP_SOLIDIFIER)
-    protected void setTipSolidifierEnabled() {
-        this.tipSolidifierEnabled = false;
-    }
     
     @Override
     public int getBelowMaxDepthTransactionLimit() {
--- a/src/main/java/com/iota/iri/conf/BaseIotaConfig.java
+++ b/src/main/java/com/iota/iri/conf/BaseIotaConfig.java
@@ -705,13 +705,13 @@ public abstract class BaseIotaConfig implements IotaConfig {
     }
 
     @JsonProperty
-    @Parameter(names = "--tip-solidifier", description = SolidificationConfig.Descriptions.TIP_SOLIDIFIER_ENABLED, 
+    @Parameter(names = "--tip-solidifier", description = SolidificationConfig.Descriptions.TIP_SOLIDIFIER, 
         arity = 1)
     protected void setTipSolidifierEnabled(boolean tipSolidifierEnabled) {
         this.tipSolidifierEnabled = tipSolidifierEnabled;
     }
 
-    @Parameter(names = "--disable-tip-solidifier", description = SolidificationConfig.Descriptions.TIP_SOLIDIFIER_ENABLED)
+    @Parameter(names = "--disable-tip-solidifier", description = SolidificationConfig.Descriptions.TIP_SOLIDIFIER)
     protected void setTipSolidifierEnabled() {
         this.tipSolidifierEnabled = false;
     }
--- a/src/main/java/com/iota/iri/conf/BaseIotaConfig.java
+++ b/src/main/java/com/iota/iri/conf/BaseIotaConfig.java
@@ -80,7 +80,10 @@ public abstract class BaseIotaConfig implements IotaConfig {
     protected int maxDepth = Defaults.MAX_DEPTH;
     protected double alpha = Defaults.ALPHA;
     private int maxAnalyzedTransactions = Defaults.MAX_ANALYZED_TXS;
-
+    
+    //Tip Solidification
+    protected boolean tipSolidifierEnabled = Defaults.TIP_SOLIDIFIER_ENABLED;
+    
     //PearlDiver
     protected int powThreads = Defaults.POW_THREADS;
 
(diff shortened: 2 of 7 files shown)


src/main/java/com/iota/iri/conf/BaseIotaConfig.java
```diff
@@ -667,6 +670,17 @@ public double getAlpha() {
...
         this.alpha = alpha;
     }
+    
+    @Override
+    public boolean isTipSolidifierEnabled() {
+        return tipSolidifierEnabled;
+    }
+
+    @JsonProperty
+    @Parameter(names = "--tip-solidifier-enabled", description = SolidificationConfig.Descriptions.TIP_SOLIDIFIER_ENABLED)
```
GalRogozinski: add `arity` please

look at other boolean properties and #930

src/main/java/com/iota/iri/conf/SolidificationConfig.java
```diff
@@ -0,0 +1,14 @@
+package com.iota.iri.conf;
+
+public interface SolidificationConfig extends Config {
+
+    /**
+     * @return {@value Descriptions#TIP_SOLIDIFIER_ENABLED}
+     */
+    boolean isTipSolidifierEnabled();
+    
+    interface Descriptions {
```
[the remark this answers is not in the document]
GalRogozinski: I personally find this less important but we did add a rule for it so lets do it

src/main/java/com/iota/iri/service/TipsSolidifier.java
```diff
@@ -36,7 +45,7 @@ public void init() {
                     scanTipsForSolidity();
                     if (log.isDebugEnabled()) {
                         long now = System.currentTimeMillis();
-                        if ((now - lastTime) > 10000L) {
+                        if ((now - lastTime) > getLogDelay()) {
```
GalRogozinski: why not just extract to a static constant?

src/main/java/com/iota/iri/service/TipsSolidifier.java
```diff
@@ -45,7 +54,7 @@ public void init() {
                     log.error("Error during solidity scan : {}", e);
                 }
                 try {
-                    Thread.sleep(RESCAN_TX_TO_REQUEST_INTERVAL);
+                    Thread.sleep(getRescanInterval());
```
GalRogozinski: Not neccessary

src/main/java/com/iota/iri/conf/BaseIotaConfig.java
```diff
@@ -745,6 +759,9 @@ protected void setPowThreads(int powThreads) {
         //TipSel
         int MAX_DEPTH = 15;
         double ALPHA = 0.001d;
+        
+        //Tip solidification
+        boolean TIP_SOLIDIFIER_ENABLED = true;
```
GalRogozinski: @jakubcech 
When you wrote the requirements why did you decide that it should be true?
IIRC we decided this thread consumes resources and might cause crashes and better be turned off.

jakubcech: Thinking was that people can turn it off if they experience problems. I mentioned that after submitting. I'm for making it false by default if it makes sense (pending investigation iiuc).

GalRogozinski: @hmoog 
Said it works fine for him

However if we don't have enough time to test for the change (community testers) we can leave it true and give a tip to the community to turn it off

jakubcech: ^ Agree

src/main/java/com/iota/iri/service/TipsSolidifier.java
kwek20: Maybe one day we want to make this type of logging configurable.
+ magic values, they hurt my eyes!

src/main/java/com/iota/iri/service/TipsSolidifier.java
kwek20: Also made for the possibility of configuration later

src/main/java/com/iota/iri/Iota.java
```diff
@@ -106,7 +106,7 @@ public Iota(IotaConfig configuration) throws IOException {
         replicator = new Replicator(node, configuration);
         udpReceiver = new UDPReceiver(node, configuration);
         ledgerValidator = new LedgerValidator(tangle, milestoneTracker, transactionRequester, messageQ);
-        tipsSolidifier = new TipsSolidifier(tangle, transactionValidator, tipsViewModel);
+        tipsSolidifier = new TipsSolidifier(tangle, transactionValidator, tipsViewModel, configuration);
```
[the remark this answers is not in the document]
kwek20: I thought about it, but that would mean we would check at 3 different places in Iota.java for config enabled/disabled. If we want to do that for more classes in the future, it will be ugly/unreadable.

For memory, the 3 instances passed are created anyways, only the logger and the class itself gets loaded. I dont know numbers but i think thats a discardable amount of ram.

src/main/java/com/iota/iri/conf/BaseIotaConfig.java
jakubcech: Should be just `--tip-solidifier` and accept `true` and `false` for enabling and disabling explicitly. 

This is done as:
@Parameter(names = "--tip-solidifier-enabled", description = SolidificationConfig.Descriptions.TIP_SOLIDIFIER_ENABLED, arity = 1)

src/main/java/com/iota/iri/service/TipsSolidifier.java
GalRogozinski: For those intervals we need to use the new `IntervalLogger` Hans introduced (not now of course).

Anyhows, some people are thinking that having too many configurations is bad. If we make it configurable, we will have to do `config.getLogDelay()`.

I think for just extract it to a constant. It makes the magic value leaner on the eye.

src/main/java/com/iota/iri/service/TipsSolidifier.java
GalRogozinski: No need to configure this. What is needed is to remove this ugly sleep and use a ScheduledExecutor.

Revert please.

src/main/java/com/iota/iri/conf/SolidificationConfig.java
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * 
+ * Configurations that should be used for the tip solidification process, 
+ * You can also completely disable the process.
+ */
+public interface SolidificationConfig extends Config {
+
+    /**
+     * @return {@value Descriptions#TIP_SOLIDIFIER_ENABLED}
```
GalRogozinski: Can you also add default value like in:
https://github.com/iotaledger/iri/pull/1159 (a PR that will be merged after LS is released)

src/main/java/com/iota/iri/Iota.java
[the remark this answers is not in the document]
GalRogozinski: My thinking was that later on we will move to a nicer design using dependency injections (I resisted this in the beginning but Hans convinced me). In this new design we will not create objects needlessly (hopefully :grimacing: ). So I wanted the logic to be like what we want to achieve.

I already told Hans to do the same on his classes.

I see and understand your point but still, let's get the logic correct

Issue #973, opened 2018-09-05
Test parallel calls to GTTA 


jakubcech:
### Description
Test calling GTTA on a single node in parallel and see how long it takes for the individual calls to return a response. We should enable parallelism (by disabling synchronization, as done here: #1245 ) on GTTA and see how it performs compared to a normal node.

**This test already exists in the Tangle tools repo already, where we have jMeter spammer. So this just needs some minor change-up:** https://github.com/iotaledger/tangle-tools/tree/master/Jmeter-Spammer 

Are we threadsafe?
Are milestones still being updated as the calls come in?
Should we only allow parallelism up to a certain amount of active requests? So the node doesn't get DOSed. 

**This being the first nightly test, part of the scope is establishing with @geminoz how we run and deploy these tests on Spamnet (or other network).**

### Motivation
Figure out if we need to invest in making changes to IRI to support better parallelization, CW caching or tip caching in the future to accommodate faster responses to GTTA that do not overload the nodes. 

### Requirements
- [ ] Use jMeter to simulate different workers. 
- [ ] Make a 1000 calls with the same depth at the same time and log the time it takes to response to all of them - with synchronizations. 
- [ ] Make a 1000 calls with the same depth at the same time and log the time it takes to response to all of them - without synchronizations.
- Run jMeter via the command line. 
- The test is also run against manually against the #1245 PR.

### Open Questions (optional)
- [ ] Where do we upload the results?

We can change the numbers later on to establish where is a good cut-off for parallel active requests.

jakubcech:
Hi @marktwtn, happy if you take a look at it and post your results here. Please let me know if anything's unclear with regards to the description. :) 👍

marktwtn:
I was wondering about the question.
```
Are milestones still being updated as the calls come in?
```
Which milestone is preferred in the question?
`latestMilestone`, `latestSolidSubtangleMilestoneIndex` or both?

marktwtn:
The experiment result of making a 1000 calls with the same depth at the same time and logging the time it takes to response to all of them:

The black dots in the charts are the response time.

IRI 1.5.3 - with synchronization
![gtta parallel test on testnet - 1 5 3](https://user-images.githubusercontent.com/6748200/50194636-bfc69e80-0376-11e9-812c-27c423e8d378.png)

Maximum response time: 23731 ms
Throughput: 2,373.699 / minute

IRI 1.5.5 - without synchronization
[...]

PR #1245, opened 2018-12-19
Parallelize GTTA API

marktwtn:
# Description

The `synchronized` keyword has been removed for improving the response time and throughput of the GTTA requests.
The pull request is based on the experiment result of issue #973.

IRI 1.5.3 - with synchronization
gtta parallel test on testnet - 1 5 3

Maximum response time: 23731 ms
Throughput: 2,373.699 / minute

IRI 1.5.5 - without synchronization
gtta parallel test on testnet - no sync

Maximum response time: 3694 ms
Throughput: 15,011.258 / minute

## Type of change

- Performance

# How Has This Been Tested?

- Use `jMeter` to send a 1000 GTTA calls with the same depth(15) at the same time and log the time it takes to response to all of them.

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/service/API.java
+++ b/src/main/java/com/iota/iri/service/API.java
@@ -833,7 +833,7 @@ public class API {
       * @return {@link com.iota.iri.service.dto.GetTransactionsToApproveResponse}
       * @throws Exception When tip selection has failed. Currently caught and returned as an {@link ErrorResponse}.
       **/
-    private synchronized AbstractResponse getTransactionsToApproveStatement(int depth, Optional<Hash> reference) throws Exception {
+    private AbstractResponse getTransactionsToApproveStatement(int depth, Optional<Hash> reference) throws Exception {
         if (depth < 0 || depth > instance.configuration.getMaxDepth()) {
             return ErrorResponse.create("Invalid depth input");
         }


jserv:
Please revise the Git commit messages. Reference: https://chris.beams.io/posts/git-commit/
At least motivations should be addressed.

jserv:
@jakubcech, can you review this PR?

GalRogozinski:
The reason we are blocking multiple calls to GTTA is to prevent DOS on the node.
During the random walks entails sig validation which is a heavy operation.

According to #973 no DOS happened since @marktwtn claimed that the latest solid milestone updated correctly. Still IF needs to validate that as a proper regression test that will always run on the network.

Before we merge I want a nightly test that does:
We need to set up on mainnet or spamnet (@geminoz this could be a great test to do on spamnet) a node without `synchronized` and a control node with `synchronized`. We need to setup GTTA spammers (preferrably JMeter which will report us stats) which will work on both nodes. Lastly, we ascertain that neither nodes start to fall out of sync.

In other words I want #973 implemented as a nightly test as a prerequisite to merge.

DyrellC:
I can confirm that parallel calls for GTTA with a high enough volume of calls can cause OOM errors fairly quickly on the node. When they are sent in batches it's not too bad, but in the event you send 1000 at once and your node has a low allocation of memory, it's likely you'll run into those OOM errors. 

image

image

PR #1248, opened 2018-12-20
Set minimum values for LS

### LINKED ISSUE #1217: Set default value and validations for local snapshot config values
### Description
We lowered the minimum validation value of LOCAL_SNAPSHOTS_DEPTH to 50 for testing and haven't set any validation for LOCAL_SNAPSHOTS_PRUNING_DELAY. We want to set DEPTH back to 100 and set up validation for PRUNING_DELAY before release.

### Motivation
Sensible minimum values that keep the node in sync before we implement the cuckoo filter. 

### Requirements
- [ ] LOCAL_SNAPSHOTS_PRUNING_DELAY default value == 50,000
- [ ] LOCAL_SNAPSHOTS_PRUNING_DELAY minimum value that passes validation = 50,000
- [ ] LOCAL_SNAPSHOTS_DEPTH default value = 100
- [ ] LOCAL_SNAPSHOTS_DEPTH minimum value that passes validation = 100

kwek20:
# Description

Sets minimum values to pruning delay and depth for local snapshots

Fixes #1217

## Type of change

# How Has This Been Tested?
Unit tests

### DIFF
--- a/src/main/java/com/iota/iri/conf/BaseIotaConfig.java
+++ b/src/main/java/com/iota/iri/conf/BaseIotaConfig.java
@@ -490,9 +490,10 @@ public abstract class BaseIotaConfig implements IotaConfig {
     @Parameter(names = {"--local-snapshots-pruning-delay"}, description =
             SnapshotConfig.Descriptions.LOCAL_SNAPSHOTS_PRUNING_DELAY)
     protected void setLocalSnapshotsPruningDelay(int localSnapshotsPruningDelay) {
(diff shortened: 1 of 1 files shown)


DyrellC:
@jakubcech  So we don't want to have the option to store less than 30 days of transactions for LS nodes? I've had several nodes running on much less and the DB's have been very small as a result, and the nodes have had no problem operating longer term. Is this a minimum that may change in the future to accommodate the smaller devices that may not want to be storing a full 30 days?

PR #1249, opened 2018-12-21
Merge spentaddresses

hmoog:
# Description

Introduces a SpentAddressesProvider which allows us to keep track of the spent addresses even when the transactions are deleted

## Type of change


src/main/java/com/iota/iri/service/spentaddresses/impl/SpentAddressesServiceImpl.java
```diff
@@ -0,0 +1,125 @@
...
+                    );
+                }
+            }
+        } catch (Exception e) {
+            throw new SpentAddressesException(e);
+        }
+
+        for (Hash address : addressesToCheck) {
+            if (wasAddressSpentFrom(address)) {
+                spentAddressesProvider.addAddress(address);
```
GalRogozinski: Probably it will be better to `filter` and do batch save

src/main/java/com/iota/iri/service/spentaddresses/impl/SpentAddressesProviderImpl.java
```diff
@@ -0,0 +1,90 @@
...
+        try {
+            tangle.save(new SpentAddress(), addressHash);
+        } catch (Exception e) {
+            throw new SpentAddressesException(e);
+        }
+    }
+
+    @Override
+    public void writeSpentAddressesToDisk(String basePath) {
+        // TODO: iterate over db and dump file
```
GalRogozinski: I understand that the current solution will be always to clear and rewrite the former file.
Fine for now but we will need a solution that is append only for future versions.

Pay attention @jakubcech

PR #1255, opened 2018-12-30
Fix unwanted dns reverse resolving in dns refresher

georgmittendorfer:
# Description

The dns refresher does a reverse lookup. If the reverse lookup is successful, neighbors that were added with their IP will be resolved via their DNS name after the lookup. 

By turning off the reverse lookup only neighbors that are added by domain name get periodically checked and updated in the dns resolver thread.

The fix was quite easy by replacing the `getHostName` method that does a reverse lookup with `getHostString` that doesn't do a reverse lookup (see JDK javadocs). This is only a minor change that shouldn't have any side effects.

Fixes # (issue) - no issue submitted

## Type of change

# How Has This Been Tested?

As the dns refresher code is not easily unit-testable, therefore I did an end to end test for UDP neighbors by running a patched node only. There might be a similar issue in `ReplicatorSourceProcessor` but as I do not use TCP I did not look into it.

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/network/Node.java
+++ b/src/main/java/com/iota/iri/network/Node.java
@@ -186,7 +186,7 @@ public class Node {
 
                     try {
                         neighbors.forEach(n -> {
-                            final String hostname = n.getAddress().getHostName();
+                            final String hostname = n.getAddress().getHostString();
                             checkIp(hostname).ifPresent(ip -> {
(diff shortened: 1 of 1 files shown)


codepleb:
Looking forward to this PR to be merged, since I currently experience sideeffects on my project (https://github.com/codepleb/inschpektor).

GalRogozinski:
Tested in debug mode to ascertain that there is no reverse lookup for IPs

PR #1256, opened 2019-01-03
Feat: Write snapshot files to temp files first

hmoog:
# Description

Since a node can crash during the snapshot creation we write the new snapshot files to temp files first. And only move them atomically if the writing succeeds. This way a node can not get corrupted if it dies during the creation of the snapshot files.

## Type of change

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/service/snapshot/impl/SnapshotProviderImpl.java
+++ b/src/main/java/com/iota/iri/service/snapshot/impl/SnapshotProviderImpl.java
@@ -9,6 +9,7 @@ import com.iota.iri.service.snapshot.*;
 import java.io.*;
 import java.nio.file.Files;
 import java.nio.file.Paths;
+import java.nio.file.StandardCopyOption;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.stream.Stream;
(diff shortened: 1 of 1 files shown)


src/main/java/com/iota/iri/service/snapshot/impl/SnapshotProviderImpl.java
```diff
@@ -118,15 +119,38 @@ public Snapshot getLatestSnapshot() {
...
-            writeSnapshotMetaDataToDisk(snapshot, basePath + ".snapshot.meta");
+            tmpStateFile = File.createTempFile("iri.", ".snapshot.state");
+            tmpMetaFile = File.createTempFile("iri.", ".snapshot.meta");
+
+            writeSnapshotStateToDisk(snapshot, tmpStateFile.getAbsolutePath());
+            writeSnapshotMetaDataToDisk(snapshot, tmpMetaFile.getAbsolutePath());
+
+            Files.move(Paths.get(tmpStateFile.getAbsolutePath()), Paths.get(basePath + ".snapshot.state"),
+                    StandardCopyOption.ATOMIC_MOVE);
+            Files.move(Paths.get(tmpMetaFile.getAbsolutePath()), Paths.get(basePath + ".snapshot.meta"),
```
alon-e: if for some reason this fails (or takes a long time) you'll be in a mid-state where the state was updated but meta not.
is this an issue? would the opposite order be better/worse? can this be done atomically as well? (move both at once).

src/main/java/com/iota/iri/service/snapshot/impl/SnapshotProviderImpl.java
```diff
@@ -118,15 +119,38 @@ public Snapshot getLatestSnapshot() {
...
-            writeSnapshotStateToDisk(snapshot, basePath + ".snapshot.state");
-            writeSnapshotMetaDataToDisk(snapshot, basePath + ".snapshot.meta");
+            tmpStateFile = File.createTempFile("iri.", ".snapshot.state");
+            tmpMetaFile = File.createTempFile("iri.", ".snapshot.meta");
+
+            writeSnapshotStateToDisk(snapshot, tmpStateFile.getAbsolutePath());
+            writeSnapshotMetaDataToDisk(snapshot, tmpMetaFile.getAbsolutePath());
+
+            Files.move(Paths.get(tmpStateFile.getAbsolutePath()), Paths.get(basePath + ".snapshot.state"),
+                    StandardCopyOption.ATOMIC_MOVE);
```
alon-e: don't you need `REPLACE_EXISTING` as well?

src/main/java/com/iota/iri/service/snapshot/impl/SnapshotProviderImpl.java
hmoog: Usually the move should not "move" but only "rename" the file, which should be a matter of milliseconds, but I guess you are right. Especially since we create the temp files in the temp directory it could happen that the file actually gets "copied" if they reside on different file systems.

Maybe it is better to create the temp files in the same path as the target files and just append them with a .tmp extension so the rename can happen very fast.

src/main/java/com/iota/iri/service/snapshot/impl/SnapshotProviderImpl.java
hmoog: REPLACE_EXISTING is ignored if ATOMIC_MOVE is provided and it is plattform dependent if it replaces existing files (which is a bad thing i guess). 

Maybe the best way to solve this is to:

1. write new ".tmp" files in the same folder
2. rename old files to ".bkp"
3. rename ".tmp" files to their correct location
4. delete ".bkp" files

This way we would always have the backup files prior to the failed snapshot which could be renamed manually by the noder operator of this procedure ever fails.

hmoog: Maybe even always keep the bkp files instead of deleting them so node operators have a way to recover from really odd edge cases.

And just REPLACE_EXISTING the ".bkp" files while renaming them.

PR #1263, opened 2019-01-09
Spent Addresses should be persisted in a seperate db instance

### LINKED ISSUE #1262: Spent Addresses by Local Snasphots should be stored using RocksDb
### Description
Currently after local snapshots and pruning we save the spent addresses data to text files.
We should persist it in the db instead, but in seperate from tangle data.

### Motivation
Spent addresses files are expected to grow large and become hard to manage.
Using text files may also open a door for race conditions that can be handled gracefully by the db

### Requirements
1. The data should be written to a seperate db folder that is not mainnetdb
2. Confirmed spent addresses should be persisted everytime we do a local snapshot
3. Unconfirmed spent addresses should be persisted everytime we prune

GalRogozinski:
# Description

Persists spent addresses to a separate db instance

Fixes #1262 

## Type of change

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/Iota.java
+++ b/src/main/java/com/iota/iri/Iota.java
@@ -30,13 +30,12 @@ import com.iota.iri.storage.Persistable;
 import com.iota.iri.storage.Tangle;
 import com.iota.iri.storage.ZmqPublishProvider;
 import com.iota.iri.storage.rocksDB.RocksDBPersistenceProvider;
-import com.iota.iri.utils.IotaUtils;
 import com.iota.iri.utils.Pair;
 import com.iota.iri.zmq.MessageQ;
 
 import java.security.SecureRandom;
-import java.util.Arrays;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
 
 import org.apache.commons.lang3.NotImplementedException;
@@ -76,6 +75,18 @@ import org.slf4j.LoggerFactory;
  */
 public class Iota {
     private static final Logger log = LoggerFactory.getLogger(Iota.class);
+    public static final HashMap<String, Class<? extends Persistable>> COLUMN_FAMILIES =
+            new LinkedHashMap<String, Class<? extends Persistable>>() {{
+        put("transaction", Transaction.class);
+        put("milestone", Milestone.class);
+        put("stateDiff", StateDiff.class);
+        put("address", Address.class);
+        put("approvee", Approvee.class);
+        put("bundle", Bundle.class);
+        put("obsoleteTag", ObsoleteTag.class);
+        put("tag", Tag.class);
+    }};
+    public static final Pair<String, Class<? extends Persistable>> METADATA_COLUMN_FAMILY = new Pair<>("transaction-metadata", Transaction.class);
 
     public final SpentAddressesProviderImpl spentAddressesProvider;
 
@@ -289,17 +300,8 @@ public class Iota {
                         configuration.getDbPath(),
                         configuration.getDbLogPath(),
                         configuration.getDbCacheSize(),
-                        new HashMap<String, Class<? extends  Persistable>>() {{
-                            put("transaction", Transaction.class);
-                            put("milestone", Milestone.class);
-                            put("stateDiff", StateDiff.class);
-                            put("address", Address.class);
-                            put("approvee", Approvee.class);
-                            put("bundle", Bundle.class);
-                            put("obsoleteTag", ObsoleteTag.class);
-                            put("tag", Tag.class);
-                        }},
-                        new Pair<>("transaction-metadata", Transaction.class))
+                        COLUMN_FAMILIES,
+                        METADATA_COLUMN_FAMILY)
                 );
                 break;
             }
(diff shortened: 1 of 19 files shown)


GalRogozinski:
@DyrellC 

Please test this.
Look for the following:
1. false positives
2. confirmed spent from addresses before/after snapshot
3. unconfirmed spent from addresses before/after snapshot
4. confirmed spent from addresses before/after pruning
5. unconfirmed spent from addresses before/after pruning

src/main/java/com/iota/iri/service/spentaddresses/impl/SpentAddressesProviderImpl.java
```diff
@@ -10,19 +10,19 @@
...
-import org.apache.commons.io.FileUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * 
+ *
  * Implementation of <tt>SpentAddressesProvider</tt>.
  * Addresses are saved/found on the {@link Tangle}.
  * The addresses will be written to a file called {@value #SNAPSHOT_SPENTADDRESSES_FILE}.
```
kwek20: You removed this file but forgot this javadoc

src/main/java/com/iota/iri/Iota.java
```diff
@@ -71,6 +75,18 @@
  */
 public class Iota {
     private static final Logger log = LoggerFactory.getLogger(Iota.class);
+    public static final HashMap<String, Class<? extends Persistable>> COLUMN_FAMILIES =
```
kwek20: This seems off compared to the rest of the class.

I would rather see 2 separate enums for this instead of a "random" map in the main class.
The main class is now imported on multiple places just for these 2 family maps.

src/main/java/com/iota/iri/service/spentaddresses/impl/SpentAddressesProviderImpl.java
```diff
@@ -85,26 +76,16 @@ private void readPreviousEpochsSpentAddresses() {
                 readSpentAddressesFromStream(
                         SpentAddressesProviderImpl.class.getResourceAsStream(previousEpochsSpentAddressesFile));
             } catch (SpentAddressesException e) {
-                log.error("failed to read spent addresses from " + previousEpochsSpentAddressesFile, e);
+                log.error("Failed to read spent addresses from " + previousEpochsSpentAddressesFile, e);
```
kwek20: We might want to throw the error here, and not catch it in the constructor.  Pretty important if were missing/failing to read this file

src/main/java/com/iota/iri/storage/rocksDB/RocksDBPersistenceProvider.java
```diff
@@ -29,58 +28,39 @@
...
 
     private RocksDB db;
     // DBOptions is only used in initDB(). However, it is closeable - so we keep a reference for shutdown.
     private DBOptions options;
     private BloomFilter bloomFilter;
     private boolean available;
 
-    public RocksDBPersistenceProvider(String dbPath, String logPath, int cacheSize) {
+    public RocksDBPersistenceProvider(String dbPath, String logPath, int cacheSize,
+                                      Map<String, Class<? extends Persistable>> columnFamilies,
```
kwek20: .. Continue from Iota.java: 
And then replace this part by reading all enum fields & class value, which we add to the map

src/main/java/com/iota/iri/Iota.java
GalRogozinski: The main class should only be used to inject dependencies.
I definitely agree with you that this detail doesn't belong here. It should belong to the `Tangle` class that needs to define what column families do we use.

GalRogozinski: Another problem is that I used `HashMap` instead of `Map`.

Also I didn't use an immutable type...
In java 9 we will get `Map.of()`
I think I will leave it as it is now, and push soon for an upgrade to Java 11 :-)

src/main/java/com/iota/iri/service/spentaddresses/impl/SpentAddressesProviderImpl.java
GalRogozinski: :+1:

src/main/java/com/iota/iri/service/spentaddresses/impl/SpentAddressesProviderImpl.java
GalRogozinski: It was like this before. I didn't catch Hans on that...
Good job :-)

GalRogozinski:
@kwek20 
>Im not sure about add/save change, Thats kind of implementation specific.
`Provider` is a confusing name. It should be named `Repository`. From here you understand that the point of this class is to always persist data.

DyrellC:
Spent address objects are continuing to persist correctly in the new db. Once the transactions are pruned, the address is injected into the spentAddresses db (Checked by searching the new db for addresses that were spent from following the pruning process). Attempts to double spend are rejected as expected, and the rejected transaction addresses are not persisted (no false positives found). While transactions are still present in the original db, validation catches attempts to spend from addresses that have already been spent from, and any chains of transactions that would lead to inconsistent ledger states are rejected, ensuring db integrity.

PR #1266, opened 2019-01-10
Don't start node if local snapshot is loaded but there is no spent ad…

### LINKED ISSUE #1260: Node should only start with the local snapshot spent address file present
### Description
Node should only start up successfully when the file that persists and updates spent addresses for an LS node is present. This should be a simple start-up check that returns an error. 

### Requirements
With file
1. I run a node and the spent addresses file the LS node updates is present.
2. The node starts as usual.

Without file
1. I run a node and the spent addresses file the LS node updates is present.
2. The node fails to run, stating that file <file name>.<file extension> is missing in <folder.

GalRogozinski:
# Description
Ensures that we have spent addresses data before we load a local snapshot

Fixes #1260 

## Type of change
- Bug fix

# How Has This Been Tested?

Isolated node has been run with and without spent address db to ensure that the node exits

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/service/spentaddresses/SpentAddressesProvider.java
+++ b/src/main/java/com/iota/iri/service/spentaddresses/SpentAddressesProvider.java
@@ -8,6 +8,15 @@ import java.util.Collection;
  * Find, mark and store spent addresses
  */
 public interface SpentAddressesProvider {
+    /**
+     * Folder where we store spent address data
+     */
+    String SPENT_ADDRESSES_DB = "spent-addresses-db";
+
+    /**
+     * Folder where we store spent address logs
+     */
+    String SPENT_ADDRESSES_LOG = "spent-addresses-log";
     
     /**
      * Checks if this address hash has been spent from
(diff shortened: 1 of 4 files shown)


GalRogozinski:
Note that this may break local unit tests if you have snapshot files in your CWD but no spent address db

src/main/java/com/iota/iri/service/spentaddresses/SpentAddressesProvider.java
```diff
@@ -8,6 +8,15 @@
  * Find, mark and store spent addresses
  */
 public interface SpentAddressesProvider {
+    /**
+     * Folder where we store spent address data
+     */
+    String SPENT_ADDRESSES_DB = "spent-addresses-db";
```
alon-e: see #1268

GalRogozinski: yeah, I don't want to make it configurable for this release

I want to raise #1268 in a team meeting and decide on the best way to approach it.
Worst case we will add it on a later version

PR #1277, opened 2019-01-14
Regression: Add machine2 tests (Transaction tests) 

DyrellC:
# Description
This PR is for adding machine2 [Transaction] tests from https://github.com/iotaledger/iri-regression/issues/31 and https://github.com/iotaledger/iri-regression/issues/32. It also contains a rework of the transaction logic for ease of writing future tests.

## Type of change

# How Has This Been Tested?
These tests were run locally on locally hosted nodes on versions 1.5.5, 1.5.6 and the newly release 1.6.0 with each passing successfully. The tests were also run using the `ciglue.sh` to make sure that the kubernetes clusters were properly passing as well. 

# Checklist:


python-regression/tests/features/steps/transaction_steps.py
```diff
@@ -129,3 +139,58 @@ def make_transaction(node, arg_list):
...
     pool.fetch_results(transaction_results[0], 30)
+
+
+@step(r'"(\d+)" transactions are issued on "([^"]+)" with:')
+def issue_multiple_transactions(step, num_transactions, node):
+    transaction_hashes = []
+    stored_values = ()
+    for iteration in range(int(num_transactions)):
+        if iteration == 0:
+            stored_values = deepcopy(step.hashes)
```
GalRogozinski: why deep copy?
Where are you changing `stored_values`?

python-regression/tests/features/steps/transaction_steps.py
```diff
@@ -129,3 +139,58 @@ def make_transaction(node, arg_list):
...
+
+@step(r'a milestone is issued with index (\d+) and referencing a hash from "([^"]+)"')
+def issue_a_milestone_with_reference(step, index, static_variable):
+    """
+    This method issues a milestone with a given index, and stores that milestone hash in staticVals.py
+
+    :param index: The index of the milestone you are issuing
+    """
+    node = world.config['nodeId']
+    address = getattr(static, "TEST_BLOWBALL_COO")
```
GalRogozinski: Is this the best way to get static vals in Python?
I know that if static was a `class` you could do static.TEST_BLOWBALL_COO?

python-regression/tests/features/steps/transaction_steps.py
```diff
@@ -129,3 +139,58 @@ def make_transaction(node, arg_list):
...
+
+    api = api_utils.prepare_api_call(node)
+
+    transactions = getattr(static, static_variable)
```
GalRogozinski: When I think of static variables I think of something that shouldn't be changed

Use `world` please

python-regression/util/milestone_logic/milestones.py
```diff
@@ -0,0 +1,43 @@
...
+        address=Address(address),
+        value=0
+    )
+
+    bundle = ProposedBundle()
+    bundle.add_transaction(txn1)
+    bundle.add_transaction(txn2)
+
+    bundle[0]._legacy_tag = Tag(converter.int_to_trytestring(index, 9))
+    bundle[1]._legacy_tag = Tag(converter.int_to_trytestring(index, 9))
```
GalRogozinski: Doesn't it suffice that only the tail tx will have the index?

python-regression/util/test_logic/api_test_logic.py
```diff
@@ -84,33 +84,49 @@ def prepare_options(args, option_list):
...
+            elif arg_type == "configValue":
+                node = fetch_config('nodeId')
+                value = world.config[value][node]
+
+            elif arg_type == "configList":
+                node = fetch_config('nodeId')
+                value = [world.config[value][node]]
+
+            if key != 'seed':
+                option_list[key] = value
```
GalRogozinski: This will grow big in time,
we need to think about it

python-regression/util/transaction_bundle_logic/bundle_logic.py
```diff
@@ -0,0 +1,22 @@
+from iota import BundleHash, Fragment
+from iota.crypto import HASH_LENGTH
+from iota.crypto.kerl import Kerl
+
+
+def finalize(bundle):
```
GalRogozinski: just remind me why you didn't use PYOTA's finalize

python-regression/util/transaction_bundle_logic/transaction_logic.py
```diff
@@ -25,26 +26,37 @@ def create_transaction_bundle(address, tag, value):
...
+            transfers=[transaction],
+            inputs=[inputs['inputs'][0]],
+            change_address=Address(getattr(static, "TEST_EMPTY_ADDRESS"))
+        )
+    else:
+        prepared_transaction = api.prepare_transfer(
+            transfers=[transaction]
+        )
 
     gtta = api.get_transactions_to_approve(depth=3)
```
GalRogozinski: just that you know, if you want to make sure you won't be getting tips that may conflict with your reference you can pass the reference as an arg to GTTA.

Not sure if this is relevant here

python-regression/tests/features/steps/transaction_steps.py
```diff
@@ -129,3 +139,58 @@ def make_transaction(node, arg_list):
...
+            for index, value in enumerate(stored_values):
+                stored_value_list[index] = value
+
+        for arg_index, arg in enumerate(step.hashes):
+            if arg['keys'] == "seed" and (arg['type'] == "staticList" or arg['type'] == "ignore"):
+                if arg['type'] != stored_value_list[arg_index]['type']:
+                    arg['type'] = stored_value_list[arg_index]['type']
+
+                if arg['values'] != stored_value_list[arg_index]['values']:
+                    arg['values'] = stored_value_list[arg_index]['values']
```
GalRogozinski: Why is this needed?

python-regression/tests/features/steps/transaction_steps.py
```diff
@@ -129,3 +139,58 @@ def make_transaction(node, arg_list):
...
+
+                new_address = getattr(static, arg['values'])
+                arg['values'] = new_address[iteration]
+        logger.info('Sending Transaction {}'.format(iteration + 1))
+        transaction = generate_transaction_and_attach(step, node)
+        transaction_hash = Transaction.from_tryte_string(transaction['trytes'][0]).hash
+        transaction_hashes.append(transaction_hash)
+
+    setattr(static, "ATTACHED_TRANSACTIONS", transaction_hashes)
```
GalRogozinski: Why aren't you using `world`?
It was added into aloe to make you feel comfy with changing global params

It is also easy to reset on hooks

python-regression/tests/features/steps/transaction_steps.py
```diff
@@ -22,26 +24,36 @@ def generate_transaction_and_attach(step, node):
     arg_list = step.hashes
```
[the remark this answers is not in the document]
DyrellC: It was mostly for readability on my part since I don't particularly like `step.hashes`, the name doesn't really define it's purpose in any way, but i've removed it here since it is unnecessary

python-regression/tests/features/steps/transaction_steps.py
```diff
@@ -22,26 +24,36 @@ def generate_transaction_and_attach(step, node):
...
+
+    for arg in arg_list:
+        if arg['keys'] == 'seed' and arg['type'] == 'staticList':
+            seed = arg['values']
+            arg['type'] = 'ignore'
+    if seed != "":
+        api = api_utils.prepare_api_call(node, seed=seed)
+        is_value_transaction = True
+    else:
+        api = api_utils.prepare_api_call(node)
```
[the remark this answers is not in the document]
DyrellC: This is in combination with the code from the issue multiple transactions step, but you're right it should be encapsulated and placed in utility format for easier readability. The same goes for the logic in the issue multiple transactions step, to make it easier to understand why the logic is there and necessary.

python-regression/util/transaction_bundle_logic/bundle_logic.py
DyrellC: If we use the PYOTA finalize it modifies the obsolete tag value and leads to unpredictable milestone indexes

python-regression/util/test_logic/api_test_logic.py
DyrellC: Yeah I agree, it'll grow too big too quickly, so i'll be separating the if/elifs into function calls and do a dictionary call list the same way I prepared the api calls. This will make it more readable and will make expanding it much easier.

python-regression/tests/features/machine2/2_transaction_tests.feature
```diff
@@ -0,0 +1,66 @@
...
+        |address                |TEST_ADDRESS               |staticValue    |
+        |value                  |0                          |int            |
+        |tag                    |ZERO9VALUE                 |string         |
+
+        And "findTransactions" is called on "nodeA" with:
+        |keys                   |values                     |type           |
+        |tags                   |ZERO9VALUE                 |list           |
+
+        Then the response for "findTransactions" on "nodeA" is stored in the static value "FOUND_TRANSACTIONS"
+        #In the default DB, the current index is 50. The next milestone issued should be 51.
```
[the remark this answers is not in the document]
DyrellC: The step that this comment was referring to has been removed now and the logic rearranged for pulling the `findTransactions` results for the `getInclusionStates` step. There will be a further breakdown in the API Util logic as well since the list of fetching logic is starting to become erroneous.

python-regression/tests/features/steps/api_test_steps.py
```diff
@@ -160,20 +161,31 @@ def spam_call(step, api_call, num_tests, node):
 
     # See if call will be made on one node or all
     api_utils.assign_nodes(node, nodes)
-    node = world.config['nodeId']
+    call_node = world.config['nodeId']
```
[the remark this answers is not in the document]
DyrellC: In this one, it's pulling the call node from the precious steps not assigning it, because the input for this call takes in an `all_nodes` argument instead of specifying a specific one. The `call_node` is the node reference that the results of the call will be stored in for reference from any preceding steps that might want to access them.

python-regression/tests/features/steps/transaction_steps.py
```diff
@@ -137,20 +122,24 @@ def make_transaction(node, arg_list):
...
+        if arg['keys'] == "seed" and arg['type'] == "staticList":
+            seed_value = arg['values']
+            seed_type = arg['type']
 
-        if iteration == 0:
-            args_copy = api_utils.duplicate_arguments(step.hashes)
+    for iteration in range(int(num_transactions)):
+        seed = ""
+        if seed_value != "" and seed_type == "staticList":
+            seed = getattr(static, seed_value)[iteration]
```
GalRogozinski: my python and aloe has grown weak...

What is this `static` object?

python-regression/util/test_logic/api_test_logic.py
```diff
@@ -47,12 +47,12 @@ def check_neighbors(step, node):
     for i in response:
         expected_neighbors = step.hashes
         if type(response[i]) != int:
-            for x in range(len(response[i])):    
+            for x in range(len(response[i])):
                 if expected_neighbors[0]['neighbors'] == response[i][x]['address']:
                     contains_neighbor[0] = True
                 if expected_neighbors[1]['neighbors'] == response[i][x]['address']:
                     contains_neighbor[1] = True
```
GalRogozinski: This function is only for 2 neighbors?

python-regression/tests/features/machine2/2_transaction_tests.feature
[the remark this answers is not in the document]
GalRogozinski: That `findTransactions` call looks like an unnecessary middle step...

If you wanted to add it as an extra assertion that the transactions were issued it should be explicitly written (and then under the hood it can the current function + another function that does the assertion)

However, If it is here because it is part of the implementation then:
1. You don't need it... you can just use the transactions created in the first step
2. Regardless of 1, implementation steps shouldn't be here

python-regression/tests/features/steps/api_test_steps.py
```diff
@@ -160,20 +161,31 @@ def spam_call(step, api_call, num_tests, node):
...
-    def run_call(node, api):
+    def run_call(node, args):
         logger.debug('Running Thread on {}'.format(node))
-        response = api.get_transactions_to_approve(depth=3)
+        api = args['api']
+        options = args['options']
+        response = api_utils.fetch_call(api_call, api, options)
         return response
 
     args = nodes
```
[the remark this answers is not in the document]
GalRogozinski: This line still looks ugly to me...

Why not do `for current_node in nodes:` in the next line?

python-regression/tests/features/steps/transaction_steps.py
DyrellC: This is the `static_values.py` file imported as `static`. It's fetching the seed from the list of seeds stored in `static_values.py`

python-regression/util/test_logic/api_test_logic.py
DyrellC: Currently yes, I'm going to do a separate PR to replace the neighbor logic, but it seemed inappropriate to tag it on to this PR.

python-regression/tests/features/machine2/2_transaction_tests.feature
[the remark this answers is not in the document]
DyrellC: I made an adjustment to the transaction logic to forego the need for the `findTransactions` steps. The step was there to provide a storage point for the transactions to be pulled for reference in the milestone step, but the new change makes that unnecessary.

python-regression/tests/features/steps/api_test_steps.py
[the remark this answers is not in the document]
DyrellC: I had intended to do this already, sorry for missing that, it's changed now. It was a remnant of some old logic that I changed before this PR.

python-regression/tests/features/steps/transaction_steps.py
DyrellC: This is the `static_values.py` file imported as `static`

PR #1286, opened 2019-01-16
Fix: added a buffer to LSManager

### LINKED ISSUE #1209: Transition buffer between LOCAL_SNAPSHOTS_INTERVAL_UNSYNCED and LOCAL_SNAPSHOTS_INTERVAL_SYNCED
### Description
The transition between `LOCAL_SNAPSHOTS_INTERVAL_UNSYNCED` and `LOCAL_SNAPSHOTS_INTERVAL_SYNCED` is now immediate in snapshotMilestone == latestMilestone. We could implement a buffer of 1-3 milestones in between the transition instead. 
UNSYNCED -> SYNCED transition. Maybe a "buffer" of 1-3 milestones would be better instead of snapshotMilestone == latestMilestone

**We agreed on a buffer of 5.**

kwek20:
# Description
To prevent jumping back and forth in and out of sync, there is now a buffer in between.
Only when the latest milestone and latest snapshot differ more than this number, we fall out of sync

Fixes #1209

## Type of change

# How Has This Been Tested?
Not yet tested, will request a node as soon as possible.
submitting the PR to start possible discussion about the implementation

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/service/snapshot/impl/LocalSnapshotManagerImpl.java
+++ b/src/main/java/com/iota/iri/service/snapshot/impl/LocalSnapshotManagerImpl.java
@@ -26,6 +26,12 @@ public class LocalSnapshotManagerImpl implements LocalSnapshotManager {
      * due.
      */
     private static final int LOCAL_SNAPSHOT_RESCAN_INTERVAL = 10000;
+    
+    /**
+     * To prevent jumping back and forth in and out of sync, there is a buffer in in between.
+     * Only when the latest milestone and latest snapshot differ more than this number, we fall out of sync
+     */
+    private static final int LOCAL_SNAPSHOT_SYNC_BUFFER = 5;
 
     /**
      * Logger for this class allowing us to dump debug and status messages.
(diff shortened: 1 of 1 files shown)


src/main/java/com/iota/iri/service/snapshot/impl/LocalSnapshotManagerImpl.java
```diff
@@ -51,6 +57,11 @@
      * Configuration with important snapshot related parameters.
      */
     private SnapshotConfig config;
+    
+    /**
+     * If this node is currently seen as in sync
+     */
+    private boolean isInSync;
```
GalRogozinski: I don't like this field:
1. It just states that this class takes more memory
2. You can easily do without it and I don't understand what value it adds
3. For this class it isn't a good example, but in general in multi threaded environments you don't want to have these kind of things since 2 different threads may modify this field.

src/main/java/com/iota/iri/service/snapshot/impl/LocalSnapshotManagerImpl.java
```diff
@@ -136,4 +146,46 @@ private void monitorThread(LatestMilestoneTracker latestMilestoneTracker) {
...
+        int latestSnapshot = snapshotProvider.getLatestSnapshot().getIndex();
+        
+        // If we are out of sync, only a full sync will get us in
+        if (!isInSync && latestIndex == latestSnapshot) {
+            isInSync = true;
+        
+        // When we are in sync, only dropping below the buffer gets us out of sync 
+        } else if (latestSnapshot < latestIndex - LOCAL_SNAPSHOT_SYNC_BUFFER) {
+            isInSync = false;
+        }
```
GalRogozinski: This could be made simpler:
```
if (latestSnapshot < latestIndex - LOCAL_SNAPSHOT_SYNC_BUFFER) {
   return false;
}
 else {
    return true;
}
```

kwek20: You cant right? Because then instead of boucing at `latestSnapshot` and `latestIndex`, it will bounce at `latestIndex - BUFFEr` and `latestsnapshot`

PR #1309, opened 2019-01-29
Added check for genisis transaction

### LINKED ISSUE #1306: Node stops syncing
### Bug description
We stop requesting transactions and the node doesn't synce

### IRI version
1.6.0

### Steps To Reproduce
1. Download a recent DB
2. Start syncing

### Expected behaviour
The node should sync

### Actual behaviour
It stops requesting transactions.
Keeps on logging solidifying milestones

kwek20:
# Description
If the node does not have tips (during startup/syncing for example), we send the genesis transaction.
Due to this being all 9's, its type is `PREFILLED_SLOT`, therefor we never actually broadcast the request.

Fixes #1306 

## Type of change

# How Has This Been Tested?
Running nodes from downloaded db dont sync, unless this extra check is done

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/network/impl/TransactionRequesterWorkerImpl.java
+++ b/src/main/java/com/iota/iri/network/impl/TransactionRequesterWorkerImpl.java
@@ -109,7 +109,8 @@ public class TransactionRequesterWorkerImpl implements TransactionRequesterWorke
         try {
             if (transactionRequester.numberOfTransactionsToRequest() >= REQUESTER_THREAD_ACTIVATION_THRESHOLD) {
                 TransactionViewModel transaction = getTransactionToSendWithRequest();
-                if (transaction != null && transaction.getType() != TransactionViewModel.PREFILLED_SLOT) {
+                if (transaction != null && (transaction.getType() != TransactionViewModel.PREFILLED_SLOT
+                        || transaction.getHash().equals(Hash.NULL_HASH))) {
                     for (Neighbor neighbor : node.getNeighbors()) {
                         try {
                             // automatically adds the hash of a requested transaction when sending a packet


src/main/java/com/iota/iri/network/impl/TransactionRequesterWorkerImpl.java
```diff
@@ -109,7 +109,8 @@ public void processRequestQueue() {
         try {
             if (transactionRequester.numberOfTransactionsToRequest() >= REQUESTER_THREAD_ACTIVATION_THRESHOLD) {
                 TransactionViewModel transaction = getTransactionToSendWithRequest();
-                if (transaction != null && transaction.getType() != TransactionViewModel.PREFILLED_SLOT) {
+                if (transaction != null && (transaction.getType() != TransactionViewModel.PREFILLED_SLOT
+                        || transaction.getHash().equals(Hash.NULL_HASH))) {
```
GalRogozinski: Note that usually we would do `Hash.Null_HASH.equals(transaction.getHash())` to avoid null pointer exceptions.

However, `transaction.getHash()` shouldn't be null and we don't want to be masking bugs. So this is actually good :-)

kwek20: Totally intended! :angel: !

GalRogozinski:
@kwek20 
I approved but maybe add a unit test that checks this very specific change
If you feel like adding other unit tests that don't concern your change do it in another PR so we don't delay the merger of this one

src/test/java/com/iota/iri/network/impl/TransactionRequesterWorkerImplTest.java
```diff
@@ -0,0 +1,142 @@
...
+    
+    private static Tangle tangle;
+    private static Node node;
+    private static TipsViewModel tvm;
+    
+    private static TransactionRequester requester;
+    private static TransactionRequesterWorkerImpl worker;
+    
+    @Rule 
+    public MockitoRule mockitoRule = MockitoJUnit.rule();
```
GalRogozinski: I always used the `MockitoJunitRunner`
You made me google around and I found this:
https://github.com/mockito/mockito/issues/391

So I guess now using rule will be the new standard in the code :+1:

src/test/java/com/iota/iri/network/impl/TransactionRequesterWorkerImplTest.java
```diff
@@ -0,0 +1,142 @@
...
+    
+    @BeforeClass
+    public static void setUp(){
+        requester = mock(TransactionRequester.class);
+        requester.init(0); //Dont drop anything
```
GalRogozinski: This line isn't doing anything because you are using a mock.
Perhaps you wanted to use a `spy` here?
Or use the actual object with mocked dependencies?

src/test/java/com/iota/iri/network/impl/TransactionRequesterWorkerImplTest.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    @Before
+    public void before() throws NoSuchFieldException, SecurityException, Exception {
+        tangle = mock(Tangle.class);
```
GalRogozinski: If you used the `mockitoRule`m why not use mock annotations like `@mock`?
Just looks nicer imo

src/test/java/com/iota/iri/network/impl/TransactionRequesterWorkerImplTest.java
```diff
@@ -0,0 +1,142 @@
...
+        tvm = mock(TipsViewModel.class);
```
GalRogozinski: `tvm` makes me think of TransactionViewModel :sweat_smile: 

Maybe call it `tipsVm` or something similar?

src/test/java/com/iota/iri/network/impl/TransactionRequesterWorkerImplTest.java
```diff
@@ -0,0 +1,142 @@
...
+        
+        worker = new TransactionRequesterWorkerImpl();
+        
+        node = mock(Node.class);
+        when(node.getNeighbors()).thenReturn(new LinkedList<>());
+        
+        //Set start queue to 0 for mocking
+        setFinalStatic(worker.getClass().getDeclaredField("REQUESTER_THREAD_ACTIVATION_THRESHOLD"), 0);
```
GalRogozinski: I actually don't like that because you are changing the behavior of the class to fit your tests.
I would rather to see a loop creating 50 dummy transactions. 

The reason is that if someone would change the value of the constant the test will pick it up easily

This is a change request

src/test/java/com/iota/iri/network/impl/TransactionRequesterWorkerImplTest.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+    
+    @AfterClass
+    public static void tearDown() throws Exception {
+        worker.shutdown();
+    }
+    
+    @Test
+    public void processRequestQueueTest() throws Exception {
+        TransactionRequesterWorkerImpl fakeWorker = mock(TransactionRequesterWorkerImpl.class);
```
GalRogozinski: Since this is the class under test, why are you mocking it?
I think you can (should) mock all the dependencies and instantiate the real object.

I see you are using `thenCallRealMethod()` which makes it kind of ok, but still...
This is a change request

src/test/java/com/iota/iri/network/impl/TransactionRequesterWorkerImplTest.java
```diff
@@ -0,0 +1,142 @@
...
+        when(fakeWorker.getTransactionToSendWithRequest())
+            .thenReturn(TVMRandomNull)
+            .thenReturn(TVMRandomNotNull)
+            .thenReturn(TVMAll9Null)
+            .thenReturn(TVMAll9NotNull);
+        
+        when(fakeWorker.init(any(), any(), any(), any())).thenCallRealMethod();
+        fakeWorker.init(tangle, requester, tvm, node);
+        
+        when(fakeWorker.isValidTransaction(any())).thenCallRealMethod();
```
GalRogozinski: hmm once you use the real impl you can return the scope of the method back to private I think :-)

src/test/java/com/iota/iri/network/impl/TransactionRequesterWorkerImplTest.java
```diff
@@ -0,0 +1,142 @@
...
+    @Test
+    public void validTipToAddTest() throws Exception {
+       assertTrue(worker.isValidTransaction(TVMRandomNull));
+       assertTrue(worker.isValidTransaction(TVMRandomNotNull));
+       assertTrue(worker.isValidTransaction(TVMAll9Null));
+       
+       assertFalse(worker.isValidTransaction(TVMAll9NotNull));
+    }
+    
+    private static Transaction getRandomTransaction(Random seed) {
```
GalRogozinski: `TransactionTestUtils`

src/test/java/com/iota/iri/network/impl/TransactionRequesterWorkerImplTest.java
```diff
@@ -0,0 +1,142 @@
...
+        Transaction transaction = TransactionViewModelTest.getRandomTransaction(seed);
+        
+        transaction.type = TransactionViewModel.FILLED_SLOT;
+        transaction.parsed = true;
+        return transaction;
+    }
+    
+    private static Transaction get9Transaction() {
```
GalRogozinski: `TransactionTestUtils`

src/test/java/com/iota/iri/network/impl/TransactionRequesterWorkerImplTest.java
```diff
@@ -0,0 +1,142 @@
...
+import com.iota.iri.network.TransactionRequester;
+import com.iota.iri.network.impl.TransactionRequesterWorkerImpl;
+import com.iota.iri.storage.Tangle;
+import com.iota.iri.utils.Converter;
+
+public class TransactionRequesterWorkerImplTest {
+    
+    //Good
+    private static final TransactionViewModel TVMRandomNull = new TransactionViewModel(
+            getRandomTransaction(new Random()), Hash.NULL_HASH);
```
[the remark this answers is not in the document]
kwek20: Ah yes! Thats why i had mocked the class under test, i couldnt get the tangle mock to work

src/test/java/com/iota/iri/network/impl/TransactionRequesterWorkerImplTest.java
```diff
@@ -0,0 +1,147 @@
...
+    @Mock(answer = Answers.RETURNS_DEEP_STUBS)
+    private static SnapshotProvider snapshotProvider;
+    
+    @Mock
+    private static MessageQ messageQ;
+    
+    private static TransactionRequester requester;
+    private static TransactionRequesterWorkerImpl worker;
+
+    private Tangle tangle;
```
GalRogozinski: I see you are using the `TangleMockUtils`

How come this doesn't have a mock?

src/test/java/com/iota/iri/network/impl/TransactionRequesterWorkerImplTest.java
```diff
@@ -0,0 +1,147 @@
...
+    public void validTipToAddTest() throws Exception {
+       assertTrue("Null transaction hash should always be accepted", worker.isValidTransaction(TVMRandomNull));
+       assertTrue("Not null transaction hash should always be accepted", worker.isValidTransaction(TVMRandomNotNull));
+       assertTrue("Null transaction hash should always be accepted", worker.isValidTransaction(TVMAll9Null));
+       
+       assertFalse("All 9s transaction should not be accepted", worker.isValidTransaction(TVMAll9NotNull));
+    }
+    
+    private void fillRequester() throws Exception {
+        for (int i=0; i< TransactionRequesterWorkerImpl.REQUESTER_THREAD_ACTIVATION_THRESHOLD; i++) {
```
GalRogozinski: I am trying to think...
Sometimes in unit tests we don't like to use the class constants becuase if a dev changes them, the test changes automatically. Sometimes we want to protect the code from devs casually changing constants.

However here it is ok, because my main worry is if someone decreases the constant, and hardcoding won't help.
So no need to change :-)

src/test/java/com/iota/iri/network/impl/TransactionRequesterWorkerImplTest.java
```diff
@@ -0,0 +1,153 @@
...
+    private Tangle tangle;
+    
+    @Mock
+    private Node node;
+    
+    @Mock
+    private TipsViewModel tipsVM;
+
+    @Before
+    public void before() throws NoSuchFieldException, SecurityException, Exception {
```
GalRogozinski: No need to declare `throws` :-)
When you commit there should be an option in the IDE to perform code analysis

src/test/java/com/iota/iri/network/impl/TransactionRequesterWorkerImplTest.java
```diff
@@ -0,0 +1,153 @@
...
+        requester = new TransactionRequester(tangle, snapshotProvider, messageQ);
+        
+        worker = new TransactionRequesterWorkerImpl();
+        worker.init(tangle, requester, tipsVM, node);
+    }
+    
+    @After
+    public void tearDown() throws Exception {
```
GalRogozinski: ditto

PR #1311, opened 2019-01-29
Change: Do not request one transaction at a time when checking for solidity

### LINKED ISSUE #1310: Request as much transactions as possible as part of `checkSolidity`
### Description
When traversing a subtangle to test for a transaction solidity in `TransactionValidator::checkSolidity` we only request the first `PREFILLED_SLOT` transaction we encounter.
Instead of stopping upon the first absent transaction, we should continue exploring the subtangle to possibly add more of them to the requester's queue.

### Motivation
Moar syncing speed!

Yes.

karimodm:
# Description

This change will make IRI request bigger portions of Tangle while exploring transactions for solidity.

Fixes #1310 

## Type of change

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/TransactionValidator.java
+++ b/src/main/java/com/iota/iri/TransactionValidator.java
@@ -266,7 +266,7 @@ public class TransactionValidator {
 
                         if (!transactionRequester.isTransactionRequested(hashPointer, milestone)) {
                             transactionRequester.requestTransaction(hashPointer, milestone);
-                            break;
+                            continue;
                         }
                     } else {
                         nonAnalyzedTransactions.offer(transaction.getTrunkTransactionHash());


karimodm:
>  it seems that the 10 milestones list is kind of ineffective
Yeah, I've been experimenting with different size values of the milestone solidification pool; although the changes were not that significant as far as I could experience.

Anyhow this change refers to the Tangle traversing mechanism when checking solidity of either a milestone or a regular transaction. Instead of requesting a single branch as a result of a single `checkSolidity` call, this change would make more than one PREFILLED_SLOT end up in the requester queue, before the next call happens: I've experienced an improvement in solidification times.

GalRogozinski:
After Discussing with @jakubcech we decided that this should be in 1.6.2 and not 1.6.1
We first want to add the stabilizing fixes and see that the system stabilizes and then add the improvements.

@karimodm if you can please add a unit test for the specific change it will be great!

src/main/java/com/iota/iri/TransactionValidator.java
```diff
@@ -266,7 +266,7 @@ public boolean checkSolidity(Hash hash, boolean milestone, int maxProcessedTrans
 
                         if (!transactionRequester.isTransactionRequested(hashPointer, milestone)) {
                             transactionRequester.requestTransaction(hashPointer, milestone);
-                            break;
+                            continue;
```
alon-e: not sure if the following suggestion makes sense:
if it's a milestone, then `continue`, else `break`.

the reason being:
this mechanism can be used to fill the request queue with arbitrary data faster.
```
0
| \
0  ?
0
| \
0  ?
0
| \
0  ?
...
```
in the current code the first `?` would be requested and stop (and if this `?` does not exist the rest will never be requested, with this change all the historical `?`s will be requested (up until the `maxProccessedTransactions` is reached).

I'm not sure this is such an issue, but my proposal would not change behaviour for random transactions, but acheive the improvement for syncing nodes.

alon-e: not sure if the following suggestion makes sense:
if it's a milestone, then `continue`, else `break`.

the reason being:
this mechanism can be used to fill the request queue with arbitrary data faster.
```
0
| \
0  ?
0
| \
0  ?
0
| \
0  ?
...
```
in the current code the first `?` would be requested and stop (and if this `?` does not exist the rest will never be requested, with this change all the historical `?`s will be requested (up until the `maxProccessedTransactions` is reached).

I'm not sure this is such an issue, but my proposal would not change behavior for random transactions but achieve the improvement for syncing nodes.

GalRogozinski: I think Alon's point is that all branches that are not confirmed by a milestone are "arbitrary data" and we don't want to have priority time or resources on them. The requester queue can fill up with sidetangle requests and thus hinder the solidification of "milestone branches".

However, if the `TipsSolidifier` is not enabled we are not supposed to be solidifying on "arbitrary data", and if it is enabled then maybe it is because the operator cares about "arbitrary data"?

Bottom line, due to the `TipsSolidifier` activation config I am with @karimodm.

alon-e: good point @karimodm, still this will take time to fill (1 tx per scan). I still think that this can be avoided by separating the cases; it will reduce the amount of change.

However, I don't insist on this, if @GalRogozinski thinks it's fine then I'm OK.

karimodm: > still this will take time to fill (1 tx per scan)
Doesn't the `continue` here make the queue fill faster in any of the cases?

@GalRogozinski true on arbitrary data, this change combined with the requester freshness should fend off those scenarios though.

PR #1322, opened 2019-02-05
refactor milestone signature parameters to be configurable

alon-e:
# Description

Refactor of milestone signatures to allow easy configuration of a different Coordinator.
with higher Merkle tree depth.

Fixes # (issue)

## Type of change

# How Has This Been Tested?

- Run locally on Mainnet and Testnet (testing all new flags)
* giving `--testnet-coordinator-signature-mode` a bad value, like `KERM` would give the following error:
```
02/05 14:15:31.993 [main] ERROR com.iota.iri.IRI$IRILauncher - There was a problem parsing commandline arguments: Invalid value for --testnet-coordinator-signature-mode parameter. Allowed values:[CURLP81, CURLP27, KERL]
```


src/main/java/com/iota/iri/service/milestone/impl/MilestoneServiceImpl.java
```diff
@@ -160,10 +188,10 @@ public void resetCorruptedMilestone(int index) throws MilestoneException {
     }
 
     @Override
-    public MilestoneValidity validateMilestone(TransactionViewModel transactionViewModel, int milestoneIndex,
-            SpongeFactory.Mode mode, int securityLevel) throws MilestoneException {
+    public MilestoneValidity validateMilestone(TransactionViewModel transactionViewModel, int milestoneIndex)
+            throws MilestoneException {
 
-        if (milestoneIndex < 0 || milestoneIndex >= 0x200000) {
+        if (milestoneIndex < 0 || milestoneIndex >= maxMilestoneIndex) {
```
alon-e: this will allow higher depths of trees. (and tighter validation for lower trees)

src/main/java/com/iota/iri/conf/TestnetConfig.java
```diff
@@ -53,6 +57,39 @@ protected void setDontValidateTestnetMilestoneSig(boolean dontValidateTestnetMil
...
+        this.coordinatorSecurityLevel = coordinatorSecurityLevel;
+    }
+
+    @Override
+    public SpongeFactory.Mode getCoordinatorSignatureMode() {
+        return coordinatorSignatureMode;
+    }
+
+    @JsonProperty("COORDINATOR_SIGNATURE_MODE")
+    @Parameter(names = "--testnet-coordinator-signature-mode", description = MilestoneConfig.Descriptions.COORDINATOR_SIGNATURE_MODE)
```
alon-e: I originally added a converter and validator, a la compass
https://github.com/iotaledger/compass/blob/3aaf05763cded4a229cc23f8e9cd3d53256948db/compass/conf/BaseConfiguration.java#L49

but the default validation is really good (see PR::test)

src/main/java/com/iota/iri/conf/MilestoneConfig.java
```diff
@@ -15,9 +17,28 @@
      */
     boolean isDontValidateTestnetMilestoneSig();
 
+    /**
+     * @return {@value Descriptions#NUMBER_OF_KEYS_IN_A_MILESTONE}
+     */
+    int getNumberOfKeysInMilestone();
```
GalRogozinski: Maybe add default value to the description like in https://github.com/iotaledger/iri/pull/1159?

src/main/java/com/iota/iri/service/milestone/impl/MilestoneServiceImpl.java
```diff
@@ -65,9 +65,31 @@
...
+     */
+    private int numberOfKeysInMilestone;
+
+    /**
+     * Specifies if signatures should be validated.<br />
      */
-    private ConsensusConfig config;
+    private boolean skipValidation;
+
+    private int maxMilestoneIndex;
```
GalRogozinski: Why did you decide to extract those from the config object and not just use the getters like before.
In my opinion this just saves lots of lines of code...

I think the nicer thing to do is to have `HashFactory.ADDRESS.create(config.getCoordinator())`.
I suppose `maxMilestoneIndex = 1 << config.getNumberOfKeysInMilestone();` can also be done as an automatically calculated config in my opinion.

The thing is that if we do this change here, we will have to do the same design in the rest of the code.

So what I am asking for you is either:
1. Explain why you think it is better to not use config getters. In this case then I insist that at least the fields will be `final`. According to DDD services should be stateless. I say that at the very least they should be immutable. (The `Hash` field won't be immutable, as you well know, but we will solve it)
2. Move everything back to the config and use getters (what I prefer). There is a reason why the setters are not visible

karimodm: It may be interesting to have a MilestoneService instance that can validate milestones coming from another coo than the one configured (wink, wink ...)

src/main/java/com/iota/iri/service/milestone/impl/MilestoneServiceImpl.java
```diff
@@ -94,7 +116,13 @@ public MilestoneServiceImpl init(Tangle tangle, SnapshotProvider snapshotProvide
         this.snapshotProvider = snapshotProvider;
         this.snapshotService = snapshotService;
         this.messageQ = messageQ;
-        this.config = config;
+
+        coordinatorAddress = HashFactory.ADDRESS.create(config.getCoordinator());
+        coordinatorSignatureMode = config.getCoordinatorSignatureMode();
+        coordinatorSecurityLevel = config.getCoordinatorSecurityLevel();
+        numberOfKeysInMilestone = config.getNumberOfKeysInMilestone();
+        skipValidation = (config.isTestnet() && config.isDontValidateTestnetMilestoneSig());
```
karimodm: Avoids very very ugly line

src/main/java/com/iota/iri/service/milestone/impl/MilestoneServiceImpl.java
alon-e: The reason is that these parameters are fetched, constructed and calculated often. I think it's not efficient to do that, but I get the point of not storing state.

Anyways, I'm not sure I'll be available this week to make the change, if you find time to do it great, if not I'll update when I get to it.

GalRogozinski: I wouldn't expect this to be the bottleneck in any scenario.
Doing a `get` should be really fast
I would expect that the HotSpots compiler to optimize such inefficiencies

alon-e:
great changes! Approved.

PR #1326, opened 2019-02-07
Use xxHash instead of SHA-256 on incoming transaction hashes cache

### LINKED ISSUE #1171: Use faster Hash function in "getBytesDigest"
### Bug description
It's not a bug at all, but consumes a lot of CPU to use SHA256 to get a digest of the incoming data.
Maybe we should use a more suitable hash function here?
I've choosen xxhash for that purpose in Hercules. Collisions should not be a problem here.

https://github.com/iotaledger/iri/blob/94377986bda39fd5ab0b5bc94231a42510484af6/src/main/java/com/iota/iri/network/Node.java#L757

viossat:
# Description

Use xxHash instead of SHA-256 on incoming transaction hashes cache to reduce CPU load and allows higher TPS throughput.

Fixes #1171

## Type of change

# Checklist:


muXxer:
Nice one! :)

GalRogozinski:
We need to do some tests to see how this improves the cpu load

If @viossat or @muXxer did their own tests I will be happy if they upload the results here

muXxer:
I only made some benchmarks in golang, sorry. But here are my results for golang:
```
Running tool: /usr/local/go/bin/go test -benchmem -run=^$ -bench . -coverprofile=/tmp/vscode-gohKcwB0/go-code-cover

goos: linux
goarch: amd64
BenchmarkMD5-4               10000000           154 ns/op           0 B/op           0 allocs/op
BenchmarkSHA256-4            10000000           229 ns/op           0 B/op           0 allocs/op
BenchmarkHighwayHash-4       30000000          41.3 ns/op           0 B/op           0 allocs/op
BenchmarkXXHash-4           100000000          10.4 ns/op           0 B/op           0 allocs/op
PASS```

229ns vs 10.4ns on my machine

This is called more often than Curl, depending on the amount of neighbors you have.

**Comment by GalRogozinski** [comment]
After 1.6.1 will be out we will check if this should be merged

**Comment by viossat** [comment]
Here [a quick benchmark](https://gist.github.com/viossat/f2f57ae5458d9c125ee7fd613abc3876) between SHA-256 and two xxHash libs ([LZ4](https://github.com/lz4/lz4-java/tree/master/src/java/net/jpountz/xxhash) and [Zero-Allocation-Hashing](https://github.com/OpenHFT/Zero-Allocation-Hashing)) hashing the same million of 1653-byte arrays. The PR implements the last one. Please note the `net.jpountz.xxhash` lib seems the fastest below 1 million.

| hash function | time (ms) |
| ------------- | ------------- |
| java.security.MessageDigest ("SHA-256") | 7204 |
| net.jpountz.xxhash.XXHash32 | 260 |
| net.jpountz.xxhash.XXHash64 | 180 |
| net.openhft.hashing.LongHashFunction | 163 |

That means a 5-neighbors node at 100 TPS would only save about 1 second every 5 minutes.

**Comment by muXxer** [comment]
At higher rate this makes more impact
```>>> (7.204/1000000.0)*5000*60*5
10.806000000000001
>>> (0.163/1000000.0)*5000*60*5
0.24450000000000002
```
Thats already a difference of 10.56s every 5 Minutes with 1000 TPS and 5 neighbors.
:wen: 1000 TPS eoy? :)

GalRogozinski:
I want to finish #1328 first so that we can track the overall improvement

PR #1327, opened 2019-02-07
Fix unknown IXI command response

viossat:
# Description

Calling an unknown command of a known IXI returns `{"duration":0}`. This change will return the standard error message `Command [Foo.bar] is unknown`.

## Type of change

# Checklist:

### DIFF
--- a/src/test/java/com/iota/iri/IXITest.java
+++ b/src/test/java/com/iota/iri/IXITest.java
@@ -1,5 +1,6 @@
 package com.iota.iri;
 
+import com.iota.iri.service.CallableRequest;
 import com.iota.iri.service.dto.AbstractResponse;
 import com.iota.iri.service.dto.ErrorResponse;
 import org.hamcrest.CoreMatchers;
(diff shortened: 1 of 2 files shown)


GalRogozinski:
@viossat 
change looks good

You want to add a unit test for this specific change?

src/test/java/com/iota/iri/IXITest.java
```diff
@@ -27,6 +32,11 @@ public static void setUp() throws Exception {
         ixiDir.create();
         ixi = new IXI();
         ixi.init(ixiDir.getRoot().getAbsolutePath());
+
+        Field ixiApiField = ixi.getClass().getDeclaredField("ixiAPI");
+        ixiApiField.setAccessible(true);
+        Map<String, Map<String, CallableRequest<AbstractResponse>>> ixiAPI = (Map<String, Map<String, CallableRequest<AbstractResponse>>>) ixiApiField.get(ixi);
```
GalRogozinski: Can you just add a new line after `=` since the line is too long.
I will need to make adjustments so that our static analysis tools will catch those things.

After that change I will merge and squash since the fix and the test are :100:

PR #1332, opened 2019-02-10
Fix: Reduce db access and cpu usage needed to persist spent addresses

### LINKED ISSUE #1321: Save spent addresses after validating a bundle
### Description
Currently we persist spent addresses while pruning.
Instead we want to persist the spent address as soon as we validate a spent bundle

### Motivation
To save resources. No need to reload txs from db. No need to run bundle validator again.
We want to piggyback on existing mechanism.

### Requirements
1. After the bundle validator is invoked, we persist a spent from address
2. Remove the other paths from the code where we persist spent addresses

GalRogozinski:
# Description

1. While we are applying a milestone, persist validated spent addresses in the db from confirmed transactions.
2. Stop persisting them while taking a local snapshot
3. Adds unit tests while doing neccessary refactoring to make the code more testable

Fixes #1321

## Type of change

# How Has This Been Tested?

Unit test added
Canary have been ran on the mainnet

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/service/ledger/impl/LedgerServiceImpl.java
+++ b/src/main/java/com/iota/iri/service/ledger/impl/LedgerServiceImpl.java
@@ -48,6 +48,8 @@ public class LedgerServiceImpl implements LedgerService {
 
     private SpentAddressesProvider spentAddressesProvider;
 
+    private BundleValidator bundleValidator;
+
     /**
      * Initializes the instance and registers its dependencies.<br />
      * <br />
@@ -67,13 +69,15 @@ public class LedgerServiceImpl implements LedgerService {
      * @return the initialized instance itself to allow chaining
      */
     public LedgerServiceImpl init(Tangle tangle, SnapshotProvider snapshotProvider, SnapshotService snapshotService,
-            MilestoneService milestoneService, SpentAddressesProvider spentAddressesProvider) {
+            MilestoneService milestoneService, SpentAddressesProvider spentAddressesProvider,
+                                  BundleValidator bundleValidator) {
 
         this.tangle = tangle;
         this.snapshotProvider = snapshotProvider;
         this.snapshotService = snapshotService;
         this.milestoneService = milestoneService;
         this.spentAddressesProvider = spentAddressesProvider;
+        this.bundleValidator = bundleValidator;
 
         return this;
     }
@@ -182,7 +186,7 @@ public class LedgerServiceImpl implements LedgerService {
                             if (transactionViewModel.getCurrentIndex() == 0) {
                                 boolean validBundle = false;
 
-                                final List<List<TransactionViewModel>> bundleTransactions = BundleValidator.validate(
+                                final List<List<TransactionViewModel>> bundleTransactions = bundleValidator.validate(
                                         tangle, snapshotProvider.getInitialSnapshot(), transactionViewModel.getHash());
 
                                 for (final List<TransactionViewModel> bundleTransactionViewModels : bundleTransactions) {
--- a/src/main/java/com/iota/iri/service/ledger/impl/LedgerServiceImpl.java
+++ b/src/main/java/com/iota/iri/service/ledger/impl/LedgerServiceImpl.java
@@ -8,6 +8,7 @@ import com.iota.iri.model.Hash;
 import com.iota.iri.service.ledger.LedgerException;
 import com.iota.iri.service.ledger.LedgerService;
 import com.iota.iri.service.milestone.MilestoneService;
+import com.iota.iri.service.snapshot.Snapshot;
 import com.iota.iri.service.snapshot.SnapshotException;
 import com.iota.iri.service.snapshot.SnapshotProvider;
 import com.iota.iri.service.snapshot.SnapshotService;
@@ -17,7 +18,6 @@ import com.iota.iri.service.spentaddresses.SpentAddressesProvider;
 import com.iota.iri.storage.Tangle;
 
 import java.util.*;
-import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 /**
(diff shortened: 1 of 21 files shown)


src/main/java/com/iota/iri/service/ledger/impl/LedgerServiceImpl.java
```diff
@@ -173,11 +186,15 @@ public boolean isBalanceDiffConsistent(Set<Hash> approvedHashes, Map<Hash, Long>
...
 
-                                final List<List<TransactionViewModel>> bundleTransactions = BundleValidator.validate(
+                                final List<List<TransactionViewModel>> bundleTransactions = bundleValidator.validate(
                                         tangle, snapshotProvider.getInitialSnapshot(), transactionViewModel.getHash());
 
                                 for (final List<TransactionViewModel> bundleTransactionViewModels : bundleTransactions) {
 
+                                    //ISSUE 1008: generateBalanceDiff should be refactored so we don't have those hidden
+                                    // concerns
+                                    persistValidatedSpentAddresses(bundleTransactionViewModels);
```
alon-e: this appears before the `BundleValidator.isInconsistent(bundleTransactionViewModels), etc.` - it seems you are adding an address even if the signature is invalid/has no funds? (I'm looking in GH, not an IDE so I might be mistaken)

src/main/java/com/iota/iri/service/ledger/impl/LedgerServiceImpl.java
```diff
@@ -217,6 +234,14 @@ public boolean isBalanceDiffConsistent(Set<Hash> approvedHashes, Map<Hash, Long>
         return state;
     }
 
+    private void persistValidatedSpentAddresses(List<TransactionViewModel> bundleTransactionViewModels) throws SpentAddressesException {
+        List<Hash> spentAddresses = bundleTransactionViewModels.stream()
+                .filter(tx -> tx.value() < 0)
+                .map(TransactionViewModel::getAddressHash)
```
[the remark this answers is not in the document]
GalRogozinski: Agree,
I will create a SingleThreadExecutor that will perform this operation

src/main/java/com/iota/iri/service/ledger/impl/LedgerServiceImpl.java
```diff
@@ -153,7 +164,9 @@ public boolean isBalanceDiffConsistent(Set<Hash> approvedHashes, Map<Hash, Long>
         Map<Hash, Long> state = new HashMap<>();
         Set<Hash> countedTx = new HashSet<>();
 
-        snapshotProvider.getInitialSnapshot().getSolidEntryPoints().keySet().forEach(solidEntryPointHash -> {
+        Snapshot initialSnapshot = snapshotProvider.getInitialSnapshot();
```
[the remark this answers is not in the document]
GalRogozinski: This change was done as part of the following commit: https://github.com/iotaledger/iri/pull/1332/commits/faa01934606eb9990d58b66a49d44a2f6714d52b - "make the code more debuggable" (w/o changing logic).

What happened is that during a unit test I got a `NullPointerException` while mocking, but it was hard to understand what was null w/o entering debug mode.
So I quickly changed the code to make it easier to work with for the next person.
I added this change in a specialized commit. It doesn't affect any logic.

src/main/java/com/iota/iri/service/ledger/impl/LedgerServiceImpl.java
GalRogozinski: It is indeed before `BundleValidator.isInconsistent(bundleTransactionViewModels)`. However, it is after `bundleValidator.validate()` which does signature validation. So I believe it is the correct place to put the method.

GalRogozinski:
> How do you plan on tackling the transition?

_Scenario 1: an operator has a node with a correct `persisted-spent-addresses db`_:
Everytime we apply a milestone we persist spent addresses. So we are always ahead of the local snapshot, which is good imho. If someone gives this db to a friend we can be sure that it is not missing any confirmed addresses. Even ones above the LS :-)

_Scenario 2: an operator is syncing from a db_:
While pruning didn't happen the spent-addresses can be found using the old mechanism that relies on transaction metadata. Once we prune a confirmed transaction there will be a check to see if it is missing:
https://github.com/iotaledger/iri/blob/151a1bba19b0cb2dee907ca1274e480266820162/src/main/java/com/iota/iri/service/transactionpruning/jobs/MilestonePrunerJob.java#L273-L279

Note #1338 that tries to optimize this code.


Am I missing something?

src/main/java/com/iota/iri/service/ledger/impl/LedgerServiceImpl.java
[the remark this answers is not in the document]
alon-e: I was referencing the function "isBalanceDiffConsistent", I just cannot comment on the function declaration.

Gal & I talked and He is aware that it's triggered in every bundle validation.

src/main/java/com/iota/iri/service/ledger/impl/LedgerServiceImpl.java
```diff
@@ -173,11 +184,16 @@ public boolean isBalanceDiffConsistent(Set<Hash> approvedHashes, Map<Hash, Long>
...
 
-                                final List<List<TransactionViewModel>> bundleTransactions = BundleValidator.validate(
+                                final List<List<TransactionViewModel>> bundleTransactions = bundleValidator.validate(
                                         tangle, snapshotProvider.getInitialSnapshot(), transactionViewModel.getHash());
 
                                 for (final List<TransactionViewModel> bundleTransactionViewModels : bundleTransactions) {
 
+                                    //ISSUE 1008: generateBalanceDiff should be refactored so we don't have those hidden
+                                    // concerns
+                                    spentAddressesService
```
alon-e: small caveat, you will also scan addresses that are not spent for sure.
for example data transactions. if probably worth first filtering only negative values and if none empty submit.

alon-e:
>Am I missing something?
I was missing the fact that we make sure again it's persisted in the pruning job.
from my understanding that `warning` should show when ppl transition to 1.6.1 if so it's worth adding to the release notes.

src/main/java/com/iota/iri/service/ledger/impl/LedgerServiceImpl.java
GalRogozinski: Yup,
We are currently submitting empty sets after filtering. There is no logical fallacy but it does result in creating more objects thus increasing GC pressure...

I would say that this is a point we need to pay attention to in general with the Data Providers

PR #1347, opened 2019-02-17
Fix: log inconsistent addresses in debug mode

GalRogozinski:
# Description

Change log level to debug when reporting on an inconsistent address

## Type of change

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/service/snapshot/impl/SnapshotStateImpl.java
+++ b/src/main/java/com/iota/iri/service/snapshot/impl/SnapshotStateImpl.java
@@ -23,6 +23,7 @@ import java.util.stream.Collectors;
  * Implements the basic contract of the {@link SnapshotState} interface.
  */
 public class SnapshotStateImpl implements SnapshotState {
+
     /**
      * Logger for this class (used to emit debug messages).
      */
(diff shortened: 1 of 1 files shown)


src/main/java/com/iota/iri/service/snapshot/impl/SnapshotStateImpl.java
```diff
@@ -166,7 +166,9 @@ public boolean equals(Object obj) {
         HashMap<Hash, Long> result = new HashMap<>();
         balances.forEach((key, value) -> {
             if (value < 0) {
-                log.info("negative value for address " + key + ": " + value);
+                if (log.isDebugEnabled()) {
```
kwek20: Whats the reason for enclosing it in the debugEnabled? This is generally used to prevent expensive method calls from running when the output of that call is only used in logging.

But since we dont do that, and log to debug now, it seems kind of double

GalRogozinski: `"negative value for address " + key + ": " + value`

Here we are concatenating strings and creating new objects which won't be used if we are not in debug mode

kwek20: Id rather see this then:
`log.debug("negative value for address {}: {}", key, value);`

GalRogozinski: Yup this is the correct way to do it

kwek20: @GalRogozinski You still have the isDebugEnabled check now :)

GalRogozinski: fixed

Issue #1091, opened 2018-10-23
Write missing unit tests for local snapshots integration


jakubcech:
### Description
We need to make sure the new local snapshot code is covered with unit tests. 

### Motivation
Unit tests are part of DOD and the current implementation of LS is missing in that respect. E.g.,

 - add unit tests for the DAGHelper
 - add more descriptive JavaDoc for DAGHelper
 - add unit tests for the introduced config vars
- ...

PR #1363, opened 2019-02-27
Test: Updated TransactionTestUtils, its usage, and javadoc

kwek20:
# Description

Changes to unit tests required for the added unit tests in #1091 

## Type of change

# Checklist:


src/test/java/com/iota/iri/TransactionTestUtils.java
```diff
@@ -138,9 +157,9 @@ public static TransactionViewModel createTransactionWithTrunkAndBranch(String tr
      * @param branch The branch transaction hash
      * @return The transaction trits
      */
-    public static byte[] getTransactionWithTrunkAndBranch(Hash trunk, Hash branch) {
-        byte[] trits = new byte[TransactionViewModel.TRINARY_SIZE];
-        return getTransactionTritsWithTrunkAndBranch(trits, trunk, branch);
+    public static byte[] getRandomTransactionWithTrunkAndBranch(Hash trunk, Hash branch) {
```
karimodm: Maybe worth specifying that this method, in contrast to the `create*` methods, do not calculate Tx's hash, it will be random / not correct instead. Is this intended and accounted for?
Also: since we are changing the naming: from its name, I would expect that this method returns a transaction object, it returns bytes instead. Same for `getTransactionWithTrunkAndBranchTrits`.

src/test/java/com/iota/iri/TransactionTestUtils.java
```diff
@@ -213,6 +232,23 @@ public static Transaction get9Transaction() {
 
         return buildTransaction(trits);
```
karimodm: Implicit casting! `buildTransaction` returns Transaction while this method `byte[]`. Typing should be fixed here and other occurrences. I know you did not write this line but it should be fixed.

karimodm:
Looks good, but some more fixes might be good since we are at it.

src/main/java/com/iota/iri/network/TransactionRequester.java
```diff
@@ -91,7 +91,7 @@ public void requestTransaction(Hash hash, boolean milestone) throws Exception {
      *
      * It used when the queue capacity is reached, and new transactions would be dropped as a result.
      */
-    // @VisibleForTesting
+    //Package Private For Testing
```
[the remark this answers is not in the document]
kwek20: @karimodm What do you mean? If you're talking about the message, pretty sure @GalRogozinski told me to name it like this.

src/test/java/com/iota/iri/TransactionTestUtils.java
kwek20: Im not sure what you mean here, maybe its from old code?

src/main/java/com/iota/iri/network/TransactionRequester.java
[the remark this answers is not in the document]
kwek20: Ah talked with Gal, were going with the annotation now!

PR #1367, opened 2019-03-05
Fix #1276: Return ip address instead of hostname in getNeighbors

### LINKED ISSUE #1276: Neighbors IP addresses are always resolved to their domain name
### Bug description
`getNeighbors` api call will return the domain name of neighbors even though node was configured with their IP address

### IRI version
1.6.0

### Hardware Spec
On what hardware is the node running on?

### Steps To Reproduce
1. In `iota.ini` add to `NEIGHBORS` some valid node IP (e.g. `tcp://1.2.3.4:15600`)
2. Run the node and call `getNeighbors` via api (`curl http://localhost:14265 \
  -X POST \
  -H 'Content-Type: application/json' \
  -H 'X-IOTA-API-Version: 1' \
  -d '{"command": "getNeighbors"}'`)

### Expected behaviour
In the response one should see the IP returned for the new neignbor (`1.2.3.4:15600`)

### Actual behaviour
One sees the resolved domain name (`my.test.node.com:15600`)

legacycode:
# Description

Will return the ip address on api call to getNeighbors.

Fixes #1276 

## Type of change

# How Has This Been Tested?

- Created unit tests in GetNeighborsResponseTest
- Start iri with <code>--testnet -n tcp://1.2.3.4:15600</code>
- <code>curl http://localhost:14265 -X POST -H 'Content-Type: application/json' -H 'X-IOTA-API-Version: 1' -d '{"command": "getNeighbors"}'</code>

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/service/dto/GetNeighborsResponse.java
+++ b/src/main/java/com/iota/iri/service/dto/GetNeighborsResponse.java
@@ -57,7 +57,7 @@ public class GetNeighborsResponse extends AbstractResponse {
      * 
      */
     @SuppressWarnings("unused")
-    private static class Neighbor {
+    public static class Neighbor {
 
         private String address;
         private long numberOfAllTransactions;
@@ -138,7 +138,7 @@ public class GetNeighborsResponse extends AbstractResponse {
          * @param n the neighbor currently connected to this node
          * @return a new instance of {@link GetNeighborsResponse.Neighbor}
          */
-        static Neighbor createFrom(com.iota.iri.network.Neighbor n) {
+        public static Neighbor createFrom(com.iota.iri.network.Neighbor n) {
             Neighbor ne = new Neighbor();
             int port = n.getPort();
             ne.address = n.getAddress().getAddress().getHostAddress() + ":" + port;
--- a/src/main/java/com/iota/iri/service/dto/GetNeighborsResponse.java
+++ b/src/main/java/com/iota/iri/service/dto/GetNeighborsResponse.java
@@ -60,13 +60,13 @@ public class GetNeighborsResponse extends AbstractResponse {
     static class Neighbor {
 
         private String address;
-        long numberOfAllTransactions;
-        long numberOfRandomTransactionRequests;
-        long numberOfNewTransactions;
-        long numberOfInvalidTransactions;
-        long numberOfStaleTransactions;
-        long numberOfSentTransactions;
-        String connectionType;
+        private long numberOfAllTransactions;
+        private long numberOfRandomTransactionRequests;
+        private long numberOfNewTransactions;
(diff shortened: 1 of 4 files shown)


src/main/java/com/iota/iri/service/dto/GetNeighborsResponse.java
```diff
@@ -25,7 +23,7 @@
      *     <li>numberOfSentTransactions</li>
      *     <li>numberOfStaleTransactions</li>
      * </ol>
-     * @see {@link com.iota.iri.service.dto.GetNeighborsResponse.Neighbor}
+     * @see com.iota.iri.service.dto.GetNeighborsResponse.Neighbor
```
GalRogozinski: @kwek20 
Can you verify that your doclet works properly with those changes?

src/main/java/com/iota/iri/service/dto/GetNeighborsResponse.java
```diff
@@ -58,16 +56,17 @@ public static AbstractResponse create(final List<com.iota.iri.network.Neighbor>
...
-                numberOfStaleTransactions,
-                numberOfSentTransactions;
-        public String connectionType;
+        long numberOfAllTransactions;
+        long numberOfRandomTransactionRequests;
+        long numberOfNewTransactions;
+        long numberOfInvalidTransactions;
+        long numberOfStaleTransactions;
+        long numberOfSentTransactions;
+        String connectionType;
```
GalRogozinski: If you decided to fix the modifiers why did you choose `default` visibility and not `private`?

legacycode: good point. fixed. can you add the rule to the local pmd file as well?

georgmittendorfer:
I examined behaviour of the getNeighbor call: with UDP neighbors it only returns the IP, if the neighbor got added with IP and it returns the domain name, if the neighbor was added with domain name. So it works properly for UDP neighbors in my opinion.

What is the benefit of returning the neighbor information not in the way the neighbors got added (domain for neighbors added by domain and IP for neighbors added with IP)? For neighbors with dynamic IP this will return different IPs whenever the IP changes.

Therefore I assume that the problem is related to the TCP stack. The change above will only cosmetically change the output and return IP addresses for neighbors that got added with domain name, too. But it will not solve the root cause for the unwanted dns reverse resolving.

See https://github.com/iotaledger/iri/pull/1255 for the fix in the UDP stack (it did not handle the TCP stack).

src/main/java/com/iota/iri/service/dto/GetNeighborsResponse.java
kwek20: Confirmed this is OK! Its still seen as a class link because every `@see` is a link

PR #1368, opened 2019-03-09
create ISSInPlace.digest test case for cIRI as reference

ed2k:
# Description

Add a unit test for ISSInPlace.digest. This function is used in bundle validation. It also serves the purpose of reference test implementation for cIRI.

## Type of change

# Checklist:

### DIFF
--- a/src/test/java/com/iota/iri/crypto/ISSTest.java
+++ b/src/test/java/com/iota/iri/crypto/ISSTest.java
@@ -78,6 +78,28 @@ public class ISSTest {
         }
     }
 
+    @Test
+    public void sigDigestISSInPlace() throws Exception {
+        int index = 0;
+        int nof = 1;
+        SpongeFactory.Mode[] modes = {SpongeFactory.Mode.KERL};
+        Hash[] hashes = {HashFactory.ADDRESS.create("DQMAVYJTIIQUCDHRIZAPYKFCFBTHNDAMHLYEEIEBSR9REQBBVUQOSDWFSGPTZFNQQTVMQPVUDHDTEZVSW")};
+        for (int i=0; i < modes.length; i++) {
+            SpongeFactory.Mode mode = modes[i];
+            byte[] seedTrits = Converter.allocateTritsForTrytes(seed.length());
+            Converter.trits(seed, seedTrits, 0);
+
+            byte[] subseed = ISS.subseed(mode, seedTrits, index);
+            byte[] key = ISS.key(mode, subseed, nof);
+            byte[] digest = new byte[Curl.HASH_LENGTH];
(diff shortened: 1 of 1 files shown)


src/test/java/com/iota/iri/crypto/ISSTest.java
```diff
@@ -78,6 +78,28 @@ public void addressGenerationISS() throws Exception {
         }
     }
 
+    @Test
+    public void sigDigestISSInPlace() throws Exception {
+        int index = 0;
+        int nof = 1;
+        SpongeFactory.Mode[] modes = {SpongeFactory.Mode.KERL};
+        Hash[] hashes = {HashFactory.ADDRESS.create("DQMAVYJTIIQUCDHRIZAPYKFCFBTHNDAMHLYEEIEBSR9REQBBVUQOSDWFSGPTZFNQQTVMQPVUDHDTEZVSW")};
```
GalRogozinski: Can you split it to two lines after the `=`?
Our static code analysis will pick this up later on

src/test/java/com/iota/iri/crypto/ISSTest.java
```diff
@@ -78,6 +78,28 @@ public void addressGenerationISS() throws Exception {
...
+            Converter.trits(seed, seedTrits, 0);
+
+            byte[] subseed = ISS.subseed(mode, seedTrits, index);
+            byte[] key = ISS.key(mode, subseed, nof);
+            byte[] digest = new byte[Curl.HASH_LENGTH];
+            byte[] sig = Arrays.copyOfRange(key, 0, key.length);
+            ISSInPlace.digest(mode, sig, 0, key, 0, digest);
+            byte[] address = ISS.address(mode, digest);
+            Hash addressTrytes = HashFactory.ADDRESS.create(address);
+            assertEquals(hashes[i].toString(), addressTrytes.toString());
```
GalRogozinski: 1. Can you please add a message in the first parameter? i.e.  "Address is not as expected"
2. The `toString()` is extraneous, `equals()` is defined :-)

src/test/java/com/iota/iri/crypto/ISSTest.java
ed2k: not sure I understand the reason, the static code analysis will think it is too wide?

src/test/java/com/iota/iri/crypto/ISSTest.java
ed2k: which function's first parameter? 
will remove toString(), again, the existing function also used toString(), should be removed for consistency.

kwek20: @ed2k every `assert()` function accepts a first parameter with a message to display when the test fails. Thats the one @GalRogozinski meant.

PR #1369, opened 2019-03-12
Bugfix: inconsistency in arrivalTime units

### LINKED ISSUE #766: `setArrivalTime` publishes both in seconds and milliseconds
`setArrivalTime` publishes both in seconds and milliseconds

### Expected behaviour
arrivalTime should be published only as milliseconds since epoch

### Actual behavior
arrivalTime is published  as both seconds and milliseconds since epoch 

source 1: [src/main/java/com/iota/iri/network/Node.java](https://github.com/iotaledger/iri/blob/d59de9fbaa2432a7123d9a4d65a1555241fcd934/src/main/java/com/iota/iri/network/Node.java#L392)
source 2: [src/main/java/com/iota/iri/service/API.java](https://github.com/iotaledger/iri/blob/bc6b943368695ab266db31347c1752ba79dca846/src/main/java/com/iota/iri/service/API.java#L660)

No.

alon-e:
# Description
When storing a transaction via the API, the arrival timestamp is stored in Seconds, while when gossiped it's stored with an arrival timestamp in Milliseconds
Fixes #766 

## Type of change

### DIFF
--- a/src/main/java/com/iota/iri/service/API.java
+++ b/src/main/java/com/iota/iri/service/API.java
@@ -855,7 +855,7 @@ public class API {
         for (final TransactionViewModel transactionViewModel : elements) {
             //store transactions
             if(transactionViewModel.store(instance.tangle, instance.snapshotProvider.getInitialSnapshot())) {
-                transactionViewModel.setArrivalTime(System.currentTimeMillis() / 1000L);
+                transactionViewModel.setArrivalTime(System.currentTimeMillis());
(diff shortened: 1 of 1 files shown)


GalRogozinski:
Need to fix
https://github.com/iotaledger/iri/issues/766#issuecomment-472048212

jakubcech:
This seems to have been fixed the other way around in #1458?

GalRogozinski:
Yes as described by the expected behavior of #766

PR #1407, opened 2019-04-14
Fix: fetch correct arrival time while traversing dags for orphaned transactions.

GalRogozinski:
# Description
While traversing the transactions approvers we use the arrival time of the original root transaction instead of the `currentTransaction`.

## Type of change

# How Has This Been Tested?

A unit test needs to be added

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/service/snapshot/impl/SnapshotServiceImpl.java
+++ b/src/main/java/com/iota/iri/service/snapshot/impl/SnapshotServiceImpl.java
@@ -527,11 +527,6 @@ public class SnapshotServiceImpl implements SnapshotService {
     private boolean isOrphaned(Tangle tangle, TransactionViewModel transaction,
             TransactionViewModel referenceTransaction, Set<Hash> processedTransactions) throws SnapshotException {
 
-        long arrivalTime = transaction.getArrivalTime() / 1000L;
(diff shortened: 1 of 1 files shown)


src/main/java/com/iota/iri/conf/BaseIotaConfig.java
```diff
@@ -958,7 +958,7 @@ protected void setPowThreads(int powThreads) {
         String PREVIOUS_EPOCHS_SPENT_ADDRESSES_TXT =
                 "/previousEpochsSpentAddresses1.txt /previousEpochsSpentAddresses2.txt " +
                         "/previousEpochsSpentAddresses3.txt";
-        long GLOBAL_SNAPSHOT_TIME = 1554901200;
+        long GLOBAL_SNAPSHOT_TIME = 1554904800;
```
kwek20: Whats the reason for this change?

GalRogozinski: The `dev` branch was not in sync for some reason...
This is a change from the release of 1.7.0

GalRogozinski:
@kwek20 
Due to your comment I looked again, it seems the whole initial check is superfluous. When we traverse we start from the initial hash anyhows

PR #1418, opened 2019-04-24
Feature: Stop solidifying when asked for pruned transaction 

### LINKED ISSUE #1371: Stop solidifying when asking for previously pruned transactions
### Description
While solidifying transactions we might start requesting old transactions that have been pruned.
We want to stop once we recognize this.

### Motivation
Optimization

### Requirements
1. When we check solidity and we encounter a tx that is in the Cuckoo filter we set a `deletedScore` to 1.
We don't persist the tx but save it in local memory
2. We start to recursively request its parents and if they are also in the cuckoo filter increment `deletedProbability`.
3. If we reach a score of 10 we abort. 
4. If we find one parent that is not in the cuckoo filter we assume that the rest are false positives.

### Open Questions (optional)
Do we need to clear the cuckoo filter of very old txs in order to decrease false positive rates?

kwek20:
# Description
When we receive a transaction hash which is in the PrunedTransactionProvider we add the hash to the PrunedTransactionVerifier, and stop the rest of the receiving transaction process.

These transactions are used to prove that a transaction hash is pruned before. We deem 10 parents of the original transaction, which are pruned, enough safety to say it was pruned.

Once it is proven, the transaction hash stays in memory marked as pruned until a restart.

When one of the parents of a requested transaction is NOT pruned, the entire chain is proven to be false positives. Chain is discarded and initial transaction is marked as "false positive" to prevent the tx requesting to start over again. Just like the pruned state, this false positive state does not persist upon restarts.

Fixes #1371 

## Type of change

# How Has This Been Tested?
Need a review over the implementation before testing further

# Checklist:


GalRogozinski:
Need to properly prioritize the merging of this PR

PR #1419, opened 2019-04-24
Fix: Made neighbors return ip if possible, otherwise host string

kwek20:
# Description

GetNeighborsResponseTest.getNeighbors was broken as it returned test.test:8888 instead of ip:port.
No. more!

## Type of change

# Checklist:


GalRogozinski:
I don't understand...
If "GetNeighborsResponseTest.getNeighbors was broken" then:
1. Why wasn't the unit test changed?
2. I still don't understand why is the unit test broken... Can you give an example of an input that will trigger the bug before your fix

kwek20:
Hm some change must have carried over to dev when i switched, cleaned up and its no longer there.
Closing

PR #1437, opened 2019-05-06
Fix: fixes spent states of addresses not getting persisted on tx pruning

luca-moser:
# Description

If a node only boots up with a database (without spent-addresses-db) and the node has LS pruning enabled, then the node will prune everything below the given thresholds but not persist the spent state of the corresponding addresses.

This commit fixes the issue by replacing the spentAddressesService's wasAddressSpentFrom() with a simple check against the persistence layer (spentAddressesProvider.containsAddress()) in order to ensure, that the spent state is correctly persisted. Previously the check would always return true, making it so that the address was never persisted as spent.

Signed-off-by: Luca Moser <moser.luca@gmail.com>

## Type of change

# How Has This Been Tested?

In the ICC network with a node with a very short pruning delay which correctly reported the "genesis" address which held the 2.7Pi as spent, even after pruning the given transaction. (the transaction itself wasn't available anymore via `getTrytes`).

# Checklist:


src/main/java/com/iota/iri/service/transactionpruning/async/AsyncTransactionPruner.java
```diff
@@ -67,6 +68,11 @@
      */
     private SpentAddressesService spentAddressesService;
 
+    /**
+     * Used to check whether an address is already persisted in the persistence layer.
+     */
```
GalRogozinski: I just want to let you know my opinion.
These types of comments could have simply been placed only on the `SpentAddressProvider` class file and that's it. Via an IDE you could read it from all classes that use it as dependency.

Right now putting such a comment in every place we use the `SpentAddressProvider` is kind of repetitive, and pretty much violates DRY (do not repeat yourself). If we want to change the comment we have to change it in multiple places.

The reason we have those is because Hans, who did a great job, really liked to have them everywhere.

Having said that, since currently the entire class looks like this, it is perfectly ok that you added the comment and no need to remove it now.

PR #1451, opened 2019-05-13
Feature: Improved CW Calculation

### LINKED ISSUE #1428: Test the DaF CW calculation against the current one
### Description 
We should experiment with the other implementation of the CW algo and compare how it behaves compared to the one we currently have.

Paul may have tried and discarded the implementation. Any other issues than the recursion point? 

CW = the size of the subgraph 

### Requirements
- [ ] Implement the changed CW calculation - for each tx in a subgraph we perform a scan BFS or DFS and udpdate the CW for the root transaction.  
- [ ] DFS can be without recursion if that is an issue. 
- [ ] Benchmark it against the current one

kwek20:
# Description
Updated the CW calculation to be up to 3x faster.
Changing the Transforming map to a regular HashMap made another small improvement.
Could not find a clear increase in memory usage as no OOM occured during any testing, even with `-Xmx512M`

Fixes #1428

## Type of change
- Improvement to CW calc leading to faster GTTA.

# How Has This Been Tested?
Tests ran, benchmarked, network tests with GTTA

** DEPTH 3**
1000*5 threads

*hashId*:

Cache: 
summary =   5000 in 00:19:06 =    4.4/s Avg:  1013 Min:   379 Max: 24359 Err:     0 (0.00%)
summary =   5000 in 00:18:35 =    4.5/s Avg:   986 Min:   356 Max:  3006 Err:     0 (0.00%)

*dev*: 
summary =   5000 in 00:35:36 =    2.3/s Avg:  1866 Min:   612 Max:  5439 Err:     0 (0.00%)

*Just Hash*:
summary =   5000 in 00:18:03 =    4.6/s Avg:   951 Min:   289 Max:  2891 Err:     0 (0.00%)
summary =   5000 in 00:18:04 =    4.6/s Avg:   953 Min:   271 Max:  2676 Err:     0 (0.00%)

-- All HashId Removed
** DEPTH 15**
100*10 threads, stoppped at 83
summary =     83 in 00:16:50 =    0.1/s Avg: 113670 Min:  9931 Max: 234565 Err:     0 (0.00%)

** DEPTH 15**
10*20 threads
summary =    200 in 00:28:11 =    0.1/s Avg: 154674 Min:  8683 Max: 339657 Err:     0 (0.00%)

# Checklist:


GalRogozinski:
I did some benchmark on the `dev` tip-selection with my laptop and a devnet db:
```    ┌───────────────────────┬─────────┬────────┬────┬────┬─────────┬────────┬──────────┬───────────────┬───────────┬──────────┐
  │         Label         │    #    │ Averag │ Mi │ Ma │  Std.   │ Error  │ Throughp │   Received    │   Sent    │   Avg.   │
  │                       │ Samples │   e    │ n  │ x  │  Dev.   │   %    │    ut    │    KB/sec     │  KB/sec   │  Bytes   │
  ├───────────────────────┼─────────┼────────┼────┼────┼─────────┼────────┼──────────┼───────────────┼───────────┼──────────┤
  │ GetTransactionsToAppr │  5000   │  156   │ 38 │ 73 │  87.50  │ 0.000% │ 28.17568 │     11.60     │   7.10    │  421.6   │
  │          ove          │         │        │    │ 1  │         │        │          │               │           │          │
  ├───────────────────────┼─────────┼────────┼────┼────┼─────────┼────────┼──────────┼───────────────┼───────────┼──────────┤


```

This is on the PR:

```
  ┌───────────────────────┬─────────┬────────┬────┬────┬─────────┬────────┬──────────┬───────────────┬───────────┬──────────┐
  │         Label         │    #    │ Averag │ Mi │ Ma │  Std.   │ Error  │ Throughp │   Received    │   Sent    │   Avg.   │
  │                       │ Samples │   e    │ n  │ x  │  Dev.   │   %    │    ut    │    KB/sec     │  KB/sec   │  Bytes   │
  ├───────────────────────┼─────────┼────────┼────┼────┼─────────┼────────┼──────────┼───────────────┼───────────┼──────────┤
  │ GetTransactionsToAppr │  5000   │  152   │ 28 │ 46 │  82.16  │ 0.000% │ 28.88737 │     11.89     │   7.28    │  421.6   │
  │          ove          │         │        │    │ 4  │         │        │          │               │           │          │

```

It looks like the results are pretty similar. Also on some tests the PR results are a bit slower. We will soon have better reproducible tests.

I believe we can do better then the current code. More ideas in my upcoming review.

GalRogozinski:
I have performed another test on an isolated node that has a mainnetdb synced to ms #110932. Spamming with a GTTA of 3 with Jmeter I obtained the following results now:

Old CW:


Label | # Samples | Average | Min | Max | Std. Dev. | Error % | Throughput | Received KB/sec | Sent KB/sec | Avg. Bytes
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
GetTransactionsToApprove | 340 | 4124 | 828 | 9147 | 1868.22 | 1.471% | 1.21077 | 0.54 | 0.32 | 456.2
TOTAL | 340 | 4124 | 828 | 9147 | 1868.22 | 1.471% | 1.21077 | 0.54 | 0.32 | 456.2



New CW:



Label | # Samples | Average | Min | Max | Std. Dev. | Error % | Throughput | Received KB/sec | Sent KB/sec | Avg. Bytes
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
GetTransactionsToApprove | 500 | 1472 | 317 | 4712 | 759.86 | 0.000% | 3.12641 | 1.29 | 0.85 | 422.6
TOTAL | 500 | 1472 | 317 | 4712 | 759.86 | 0.000% | 3.12641 | 1.29 | 0.85 | 422.6


So the new implementation is approximately 3X faster

kwek20:
Whoa, did not expect that @GalRogozinski.
Guess we are going to use this in production code? Ill take another pass over it to check for some improvements

src/main/java/com/iota/iri/service/tipselection/impl/RecursiveWeightCalculator.java
```diff
@@ -0,0 +1,147 @@
...
+     * 
+     * @param txHash The tx we find the approvers of
+     * @param txToDirectApprovers The map we look in, and add to
+     * @param fallback The map we check in before going in the database, can be <code>null</code>
+     * @return
+     * @throws Exception
+     */
+    private ArrayDeque<Hash> getTxDirectApproversHashes(Hash txHash,  
+            Map<Hash, ArrayDeque<Hash>> txToDirectApprovers, 
+            Map<Hash, ArrayDeque<Hash>> fallback)
```
GalRogozinski: no need for fallback

src/main/java/com/iota/iri/service/tipselection/impl/RecursiveWeightCalculator.java
```diff
@@ -0,0 +1,147 @@
...
+            int rating = txHash.equals(entryPoint) ? hashWeight.size() + 1 : getRating(txHash, set);
+            hashWeight.put(txHash, rating);
+        }
+        return hashWeight;
+    }
+
+    private int getRating(Hash hash, Set<HashId> seenHashes) throws Exception {
+        int weight = 1;
+
+        ArrayDeque<Hash> approvers = getTxDirectApproversHashes(hash, txToDirectApprovers, null);
```
GalRogozinski: If instead of the field you will pass the parameter from `calculatingRatingDfs` you will also save on I/O calls

src/main/java/com/iota/iri/service/tipselection/impl/RecursiveWeightCalculator.java
```diff
@@ -0,0 +1,147 @@
...
+        for (Hash approver : approvers) {
+            if (!seenHashes.contains(approver)) {
+                seenHashes.add(approver);
+                weight += getRating(approver, seenHashes);
```
GalRogozinski: It is better to have something iterative rather than recursive calls. May prevent `StackOverflows`...
You are simply doing a `dfs`, so you can reuse the code of `calculateRatingDfs`

GalRogozinski:
@kwek20 
I just checked and even if you remove the field you get 3x better results

src/main/java/com/iota/iri/service/tipselection/impl/RecursiveWeightCalculator.java
```diff
@@ -0,0 +1,133 @@
...
+        Map<Hash, HashSet<Hash>> txToDirectApprovers = new HashMap<>();
+
+        Deque<Hash> stack = new ArrayDeque<>();
+        stack.push(entryPoint);
+
+        while (CollectionUtils.isNotEmpty(stack)) {
+            Hash txHash = stack.peekLast();
+
+            HashSet<Hash> approvers = getTxDirectApproversHashes(txHash, txToDirectApprovers);
+            if (null != approvers && (approvers.size() == 0 || hasAll(hashWeight, approvers, stack))) {
```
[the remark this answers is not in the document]
GalRogozinski: Do this, and no need to use `CollectionUtils`
Since we have guava now, let's try to get rid of apache

src/main/java/com/iota/iri/service/tipselection/impl/RecursiveWeightCalculator.java
```diff
@@ -55,53 +53,49 @@ public RecursiveWeightCalculator(Tangle tangle, SnapshotProvider snapshotProvide
...
-        
-        Deque<Hash> stack = new ArrayDeque<>();
-        Map<Hash, ArrayDeque<Hash>> txToDirectApprovers = new HashMap<>();
 
+        Map<Hash, HashSet<Hash>> txToDirectApprovers = new HashMap<>();
+
+        Deque<Hash> stack = new ArrayDeque<>();
         stack.push(entryPoint);
+
         while (CollectionUtils.isNotEmpty(stack)) {
```
GalRogozinski: you can use regular `!isEmpty()` let's get rid of `CollectionUtils`

src/main/java/com/iota/iri/service/tipselection/impl/RecursiveWeightCalculator.java
```diff
@@ -55,53 +53,49 @@ public RecursiveWeightCalculator(Tangle tangle, SnapshotProvider snapshotProvide
 
         // Estimated capacity per depth, assumes 5 minute gap in between milestones, at 3tps
         UnIterableMap<HashId, Integer> hashWeight = createTxHashToCumulativeWeightMap( 5 * 60 * 3 * depth);
```
GalRogozinski: Maybe call it `hashToWeight` or `hashWeightMap`.
This variable has an integer name

src/main/java/com/iota/iri/service/tipselection/impl/RecursiveWeightCalculator.java
```diff
@@ -0,0 +1,133 @@
...
+                if (nonDupes.add(hash)) {
+                    stack.add(hash);
+                }
+            }
+        }
+
+        return nonDupes.size();
+    }
+
+    private boolean hasAll(UnIterableMap<HashId, Integer> source, HashSet<Hash> requester, Deque<Hash> stack) {
```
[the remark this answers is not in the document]
GalRogozinski: The name of this method is unclear to me...
What are the reasons for the names `source` and `requester`?

src/main/java/com/iota/iri/service/tipselection/impl/RecursiveWeightCalculator.java
```diff
@@ -0,0 +1,133 @@
...
+                stack.removeLast();
+            } else {
+                stack.addAll(approvers);
+            }
+        }
+
+        return hashWeight;
+    }
+
+    private int getRating(HashSet<Hash> nonDupes, Map<Hash, HashSet<Hash>> txToDirectApprovers) throws Exception {
```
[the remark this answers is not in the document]
GalRogozinski: I don't understand the name nonDupes. Using `Set` is enough to convey non-dupes. Maybe us a better name?

src/main/java/com/iota/iri/service/tipselection/impl/RecursiveWeightCalculator.java
```diff
@@ -55,53 +53,49 @@ public RecursiveWeightCalculator(Tangle tangle, SnapshotProvider snapshotProvide
...
-        int weight = 1;
-
-        ArrayDeque<Hash> approvers = getTxDirectApproversHashes(hash, null);
-        for (Hash approver : approvers) {
-            if (!seenHashes.contains(approver)) {
-                seenHashes.add(approver);
-                weight += getRating(approver, seenHashes);
+    private int getRating(HashSet<Hash> nonDupes, Map<Hash, HashSet<Hash>> txToDirectApprovers) throws Exception {
+        Deque<Hash> stack = new ArrayDeque<>(nonDupes);
+        while (CollectionUtils.isNotEmpty(stack)) {
```
GalRogozinski: Get rid of `CollectionUtils`

src/main/java/com/iota/iri/service/tipselection/impl/RecursiveWeightCalculator.java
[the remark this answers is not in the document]
GalRogozinski: Can you explain to me the `hasAll` method, and what's its purpose?
I understood that it means that all the approvers of `txHash` had their ratings calculated.
But it took me time to understand this.

Btw, why are you doing it this way that you scan all the way to the tips and then you scan back down?
Will it not be more effective to just scan in one direction?

src/test/java/com/iota/iri/service/tipselection/impl/CumulativeWeightCalculatorTest.java
```diff
@@ -269,6 +283,7 @@ public void testTangleWithCircle2() throws Exception {
     }
 
     @Test
+    @Ignore
```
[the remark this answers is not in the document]
kwek20: It tests a broken thing in our previous system, collisions between HashIds of hashes. 
`transactionHash2` is a similar hash to the other one, and is therefore not included.
No idea why we ever did that, as it is literally skipping on transactions this way.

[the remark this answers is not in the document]
kwek20: Actually, id vote for removing the hashId entirely, having it might speed some things up, but if it blocks certain tx, its still not worth it. (in the test, tx1 and tx2 work because they reference to the same value in the end.)

An example on how to break it earlier is make tx3 and tx1 the same instead of 1 and 2. Then 3 should be value 1, and not 2, which will error

src/main/java/com/iota/iri/service/tipselection/impl/RecursiveWeightCalculator.java
```diff
@@ -75,21 +78,36 @@ public RecursiveWeightCalculator(Tangle tangle, SnapshotProvider snapshotProvide
...
-            HashSet<Hash> approvers = getTxDirectApproversHashes(stack.pollLast(), txToDirectApprovers);
+    /**
+     * Gets the rating of a set, calculated by checking its approvers
+     * 
+     * @param startingSet
+     * @param txToDirectApproversCache 
+     * @return
+     * @throws Exception
+     */
+    private int getRating(Set<Hash> startingSet, Map<Hash, Set<Hash>> txToDirectApproversCache) throws Exception {
```
GalRogozinski: Change the name to `getWeight`.

Optional:
I think it will be cleaner to add + 1 when you return at the end instead. Or conversely instead of `startingSet` you can have `startingHash` or `entryHash`. But then you will have that `getApprovers` call that you did before in `calculateRatingDfs`. It is presumably not so bad because you have the cache and the code will become more readable (you get the weight of the tx in the param).

src/test/java/com/iota/iri/service/tipselection/impl/CumulativeWeightCalculatorTest.java
```diff
@@ -283,7 +283,6 @@ public void testTangleWithCircle2() throws Exception {
     }
 
     @Test
-    @Ignore
```
GalRogozinski: Are we using the prefix again?
How come this is passing?

When I asked before I just didn't check what this unit test does.
Then you reminded and convinced me that this should be ignored and probably deleted later once we finalize tipsel.

src/main/java/com/iota/iri/service/tipselection/impl/CumulativeWeightCalculator.java
```diff
@@ -55,40 +58,58 @@ public RecursiveWeightCalculator(Tangle tangle, SnapshotProvider snapshotProvide
...
+                // Add the tx to the approvers list to count itself as +1 weight, preventing self-referencing
+                approvers.add(txHash);
+                
+                // calculate and add rating. Naturally the first time all approvers need to be looked up. Then its cached.
+                hashWeightMap.put(txHash, getRating(approvers, txToDirectApprovers));
+            } 
         }
 
-        return hashWeight;
+        // If we have a circular reference, its already added, otherwise we save a big calculation
```
GalRogozinski: circular -> self referencing

src/main/java/com/iota/iri/service/tipselection/impl/CumulativeWeightCalculator.java
```diff
@@ -55,40 +58,58 @@ public RecursiveWeightCalculator(Tangle tangle, SnapshotProvider snapshotProvide
...
+        if (!hashWeightMap.containsKey(entryPoint)) {
+            hashWeightMap.put(entryPoint, hashWeightMap.size() + 1);
```
GalRogozinski: nice

src/test/java/com/iota/iri/service/tipselection/impl/CumulativeWeightCalculatorTest.java
```diff
@@ -18,7 +18,6 @@
 import org.junit.AfterClass;
 import org.junit.Assert;
 import org.junit.BeforeClass;
-import org.junit.Ignore;
```
GalRogozinski: So we're still using transforming map
You want to avoid doing more tests for now?

Fine with that, we can change it later

jakubcech:
🎉

PR #1454, opened 2019-05-14
Added jacoco agent with java_options

sadjy:
# Description
Adding jacoco agent to TIAB config files used in python regression tests.

Fixes # (issue)

Configuration changes

# Checklist:

### DIFF
--- a/python-regression/tests/features/machine1/config.yml
+++ b/python-regression/tests/features/machine1/config.yml
@@ -12,7 +12,7 @@ defaults: &api_tests_config_files
   '--testnet',
   'true'
   ]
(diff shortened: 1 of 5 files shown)


python-regression/tests/features/machine1/config.yml
```diff
@@ -12,6 +12,8 @@ defaults: &api_tests_config_files
   '--testnet',
   'true'
   ]
+  java_options: -agentlib:jdwp=transport=dt_socket,server=y,address=8000,suspend=n -javaagent:/opt/jacoco/lib/jacocoagent.jar=destfile=/workdir/jacoco.exec,output=file,append=true,dumponexit=true
```
GalRogozinski: Just a little question about the `append` option.
Previously we appended the results of the regression tests to the unit tests. Currently we are not appending results but merging them later. What bothers me is that I don't know if `/wordir` is ephemeral. If it isn't then the `append` option may append new results on previous ones. If it gets deleted after the container is destroyed then it doesn't matter.

PR #1458, opened 2019-05-18
Fix: set arrival time in millis (not seconds) (#766)

### LINKED ISSUE #766: `setArrivalTime` publishes both in seconds and milliseconds

georgmittendorfer:
# Description

In API method storeTransactionStatement arrival date was set in seconds. Changed that to milliseconds. See cliri bug fix referenced in issue.

Fixes #766 
Replaces PR #1369

* Added tests
* Updated mockito (for new tests)
* Refactored incompatible mockito usage
* Made Iota.configuration private (instance access not needed and should be injected. Was easier to use this way in tests, too)
* Added documentation that arrival time is in millis

Attention: SnapshotServiceImpl#isOrphaned NOT changed. See comment in issue.

## Type of change

# How Has This Been Tested?

Unit tests that check that set arrival time is close to System.currentTimeMillis

# Checklist:

- [x] I have added tests that prove my fix is effective

### DIFF
--- a/src/test/java/com/iota/iri/service/APITest.java
+++ b/src/test/java/com/iota/iri/service/APITest.java
@@ -4,11 +4,12 @@ import com.iota.iri.TransactionValidator;
 import com.iota.iri.conf.IotaConfig;
 import com.iota.iri.controllers.TransactionViewModel;
 import com.iota.iri.service.snapshot.SnapshotProvider;
+import org.junit.Rule;
 import org.junit.Test;
-import org.junit.runner.RunWith;
 import org.mockito.Answers;
 import org.mockito.Mock;
-import org.mockito.junit.MockitoJUnitRunner;
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
 
 import java.util.Collections;
 
@@ -16,11 +17,13 @@ import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.ArgumentMatchers.anyInt;
 import static org.mockito.ArgumentMatchers.longThat;
 import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
 import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
 
-@RunWith(MockitoJUnitRunner.class)
 public class APITest {
+
+    @Rule
+    public MockitoRule mockitoRule = MockitoJUnit.rule();
 
     @Mock(answer = Answers.RETURNS_SMART_NULLS)
     private TransactionValidator transactionValidator;
--- a/src/test/java/com/iota/iri/service/APITest.java
+++ b/src/test/java/com/iota/iri/service/APITest.java
@@ -12,7 +12,12 @@ import org.mockito.junit.MockitoJUnitRunner;
 
 import java.util.Collections;
 
-import static org.mockito.Mockito.*;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyInt;
+import static org.mockito.ArgumentMatchers.longThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.verify;
 
 @RunWith(MockitoJUnitRunner.class)
 public class APITest {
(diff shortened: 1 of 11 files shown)


src/test/java/com/iota/iri/service/APITest.java
```diff
@@ -0,0 +1,56 @@
...
+import java.util.Collections;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyInt;
+import static org.mockito.ArgumentMatchers.longThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+import static org.mockito.Mockito.verify;
+
+@RunWith(MockitoJUnitRunner.class)
```
GalRogozinski: Hmm, now that you have upgraded Mockito this doesn't give a warning.

I agree that it is nicer than then using `@rule`, the problem is that we started using rules elsewhere in the code.
@kwek20, what do you think? You want to make the switch?

src/test/java/com/iota/iri/service/tipselection/impl/WalkerAlphaTest.java
```diff
@@ -58,8 +58,7 @@ public static void setUp() throws Exception {
         tangle.init();
 
         TailFinder tailFinder = Mockito.mock(TailFinder.class);
-        Mockito.when(tailFinder.findTail(Mockito.any(Hash.class)))
-                .then(args -> Optional.of(args.getArgumentAt(0, Hash.class)));
+        Mockito.when(tailFinder.findTail(Mockito.any(Hash.class))).then(args -> Optional.of(args.getArgument(0)));
```
GalRogozinski: I actually like it that the line was broken down. I found it more readable.
I need to play with the `format_settings.epf` file so it won't format the lines this way.

georgmittendorfer:
`For next time though if you want to do refactoring that is not in the scope of the issue please put it in a different PR.` -> OK! I knew you would be saying this, as we had the same discussion last time... :sweat_smile:

Reason (assuming you mean Iota.configuration): I implemented against an old revision (there it was in scope) but someone refactored the code in the meantime. After merging these changes  were left over and I was too lazy to revert and move them into another branch.

src/test/java/com/iota/iri/service/APITest.java
georgmittendorfer: Did this out of habit. Just tell me what you prefer an I will change it. Another option would be not to use the runner at all and call `MockitoAnnotations.initMocks(this);` in the `@Before`. That way the tests stay open for other Runners.

src/test/java/com/iota/iri/service/tipselection/impl/WalkerAlphaTest.java
georgmittendorfer: OK, will change that.

src/test/java/com/iota/iri/service/APITest.java
kwek20: I think using the `@Rule` is fine for now. If we ever decide to make a change for Junit 5, we will need to do an overhaul anyways, and we can decide then.

If you want to change to a different runner, `@RunWith` is available.
Please change it to the following:

```
@Rule
public MockitoRule mockitoRule = MockitoJUnit.rule();
```

georgmittendorfer: @kwek20 OK, will do!

georgmittendorfer:
There is something wrong with the new unit tests. Only some builds fail with:

`[ERROR] Tests run: 2, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 0.112 s <<< FAILURE! - in com.iota.iri.network.NodeTest
[ERROR] whenProcessReceivedDataSetArrivalTimeToCurrentMillis(com.iota.iri.network.NodeTest)  Time elapsed: 0.073 s  <<< ERROR!
java.lang.NullPointerException
	at com.iota.iri.network.NodeTest.whenProcessReceivedDataSetArrivalTimeToCurrentMillis(NodeTest.java:88)`

I had this once locally and it made `continuous-integration/travis-ci/pr — The Travis CI build failed` fail, too. WIll investigate but can't reproduce. Any ideas?

Issue #1477, opened 2019-06-02
LatestMilestoneTracker error while analyzing the milestone candidates


georgmittendorfer:
### Bug description

Exception in log file from LatestMilestoneTracker: `MilestoneException: unexpected error while analyzing the milestone candidates`

### IRI version

Patched IRI 1.7.0-DEV

### Steps To Reproduce

I don't know.

### Errors

Reason: java.lang.IndexOutOfBoundsException in MilestoneServiceImpl:

```
06/02 04:43:55.702 [Latest Milestone Tracker] ERROR LatestMilestoneTrackerImpl:465 - error while analyzing the milestone candidates
com.iota.iri.service.milestone.MilestoneException: unexpected error while analyzing the transaction XFRTRIJGPEAKJHZKKLLACOOGCVGBCTEGYFOVBKPAXCUYFRQQQEFVBOFYWLEHVQUYOXBDIG99EULS99999
        at com.iota.iri.service.milestone.impl.LatestMilestoneTrackerImpl.processMilestoneCandidate(LatestMilestoneTrackerImpl.java:170) ~[iri-1.7.0-DEV-6bde715.jar:1.7.0-DEV-6bde715]
        at com.iota.iri.service.milestone.impl.LatestMilestoneTrackerImpl.analyzeMilestoneCandidates(LatestMilestoneTrackerImpl.java:347) ~[iri-1.7.0-DEV-6bde715.jar:1.7.0-DEV-6bde715]
        at com.iota.iri.service.milestone.impl.LatestMilestoneTrackerImpl.latestMilestoneTrackerThread(LatestMilestoneTrackerImpl.java:274) ~[iri-1.7.0-DEV-6bde715.jar:1.7.0-DEV-6bde715]
        at com.iota.iri.utils.thread.BoundedScheduledExecutorService.lambda$wrapTask$6(BoundedScheduledExecutorService.java:823) ~[iri-1.7.0-DEV-6bde715.jar:1.7.0-DEV-6bde715]
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511) ~[na:1.8.0_172]
        at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:308) ~[na:1.8.0_172]
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$301(ScheduledThreadPoolExecutor.java:180) ~[na:1.8.0_172]
[...]

PR #1480, opened 2019-06-04
Fix: Prevents IndexOutOfBoundsException within the MilestoneServiceImpl

luca-moser:
# Description
Fixes an IndexOutOfBoundsException occurring in the MilestoneServiceImpl.

Fixes # (issue)
https://github.com/iotaledger/iri/issues/1477

## Type of change

# How Has This Been Tested?
Private net.

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/service/milestone/impl/MilestoneServiceImpl.java
+++ b/src/main/java/com/iota/iri/service/milestone/impl/MilestoneServiceImpl.java
@@ -175,9 +175,9 @@ public class MilestoneServiceImpl implements MilestoneService {
         }
 
         try {
-            if (MilestoneViewModel.get(tangle, milestoneIndex) != null) {
-                // Already validated.
-                return VALID;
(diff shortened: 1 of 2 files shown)


src/main/java/com/iota/iri/service/milestone/impl/MilestoneServiceImpl.java
```diff
@@ -175,9 +175,9 @@ public MilestoneValidity validateMilestone(TransactionViewModel transactionViewM
         }
 
         try {
-            if (MilestoneViewModel.get(tangle, milestoneIndex) != null) {
-                // Already validated.
-                return VALID;
+            MilestoneViewModel existingMilestone = MilestoneViewModel.get(tangle, milestoneIndex);
+            if(existingMilestone != null){
+                return existingMilestone.getHash().equals(transactionViewModel.getHash()) ? VALID : INVALID;
```
GalRogozinski: Just that you know for future code,
I usually like to use `Objects#equals` to handle the `null` case

In this case I'd rather have a `NullPointerException` so we can find the bug early on :-)
So it is all good

PR #1494, opened 2019-06-12
Revert "Fix: Return ip address instead of hostname in getNeighbors (#1276)

GalRogozinski:
# Description

This reverts #1276

Users complained that the fix may produce additional problems.
We will wait for #1393

Quoting from Discord:

> From what I have seen, I don't think it is good that the neighbors are returned (in getNeighbors) with their IP addresses rather than with the names with which they have been submitted to IRI.
> It is going to be very hard for node owners to associate the IP addresses with the neighbors they have added.
> Consider that most of the community node operators use GUI or other linux shell menu/tools based on getNeighbors to manage their peers.
> Node operators will be forced to a new shell, open up their IRI config where their neighbors are listed and try to determine one by one the peers (get each FQDN's IP, and match to getNeighbors output ....) 

## Type of change

# How Has This Been Tested?

IRI ran locally and `Get Neighobrs` was invoked.

# Checklist:


GalRogozinski:
Both tests were added in the PR so they got reverted automatically:
`NodeTest` - we can keep it maybe since I think it is good regardless of the revert
`GetNeighborResponseTest` - it will have to change since it depends on the reverted feature. Since this was suppose to be a simple revert I don't think it is in scope.

Anyhows, since networking will change soon I just let the tests get reverted. Do you want me bring back `NodeTest`?

luca-moser:
>Anyhows, since networking will change soon I just let the tests get reverted. Do you want me bring back NodeTest?
Ah ok, then it's fine.

PR #1505, opened 2019-07-10
Fix machine6 for networking

DyrellC:
# Description
Adds milestone issuance to the `machine6` regression tests to allow the nodes to start syncing now that the new network code does not request random tips from the other nodes. 

Fixes the failure of `machine6` with the new network rewrite

## Type of change

# How Has This Been Tested?
-Tested on local nodes using the new iri with the network rewrite.

# Checklist:


DyrellC:
@galrogo  unfortunately yes we do need to increase it, because now the syncing doesn't start automatically when the cluster is created like it used to. Now it has to be initiated manually with the milestone, and it needs the additional time to actually retrieve the transactions/milestones from the other node.

PR #1506, opened 2019-07-10
Fix: prevent self initialized connection being removed from reconnect pool too early

luca-moser:
# Description
If we initialize a connection to a neighbor ourselves, we should not remove that URI from the reconnect
pool up on the initialized (but not handshaked) connection, because it will lead to the node
not retrying a reconnect attempt in case the handshaking/read op. was faulty.

## Type of change

# Checklist:


GalRogozinski:
Fixes: #1507

PR #1520, opened 2019-07-15
Feature: Cache in between Tangle and  database

### LINKED ISSUE #935: Higher level cache above the standard RocksDB cache
### Description
Currently RocksDB caches data in bytes, this means that you may create two separate transaction objects and manipulate them in different threads. This can leads to inconsistencies when they're being saved to the database again. 

### Motivation
Local snapshots (?), possibly other bugs. 

### Requirements
- [ ] 

### Open Questions (optional)
Create scenarios where this is replicatable. To understand the importance of this issue - are we trying to solidify transactions that are already solid, for example?

kwek20:
# Description
Adds a cache that keeps transaction data in memory. This is required because the RocksDB cache does not work currently.

Fixes #935 

## Type of change

# How Has This Been Tested?
Added unit tests, ran on network.

# Checklist:
- [X My code follows the style guidelines for this project

### DIFF
--- a/src/main/java/com/iota/iri/controllers/TransactionViewModel.java
+++ b/src/main/java/com/iota/iri/controllers/TransactionViewModel.java
@@ -3,6 +3,7 @@ package com.iota.iri.controllers;
 import com.iota.iri.model.*;
 import com.iota.iri.model.persistables.*;
 import com.iota.iri.service.snapshot.Snapshot;
+import com.iota.iri.storage.DataCache;
 import com.iota.iri.storage.Indexable;
 import com.iota.iri.storage.Persistable;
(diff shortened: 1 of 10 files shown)


src/main/java/com/iota/iri/controllers/TransactionViewModel.java
```diff
@@ -1,19 +1,38 @@
 package com.iota.iri.controllers;
 
-import com.iota.iri.model.*;
-import com.iota.iri.model.persistables.*;
+import java.util.ArrayList;
```
kwek20: Whoeps i reformatted this by accident. Lets merge fast to not get too many conflicts :)

Issue #844, opened 2018-06-28
Solid milestones stop updating on an out-of-sync node once getTransactionsToApprove is received


rajivshah3:
### Bug description
My node is out-of-sync with the network. If I turn off `nginx`, therefore preventing API requests from coming to my node, the solid milestones will gradually increase as usual. If I turn on `nginx` and receive a `getTransactionsToApprove` request, the solid milestones will stop updating and resource usage increases significantly.

[Logs during getTransactionsToApprove request](https://pastebin.com/GrmtPCZR)

Before `getTransactionsToApprove`,  IRI used 70% CPU and 6.88 GB RAM. After `getTransactionsToApprove`, IRI used 262% CPU and 9.34 GB RAM.

### Hardware Spec
Ubuntu 16.04, 4vCPUs and 12GB RAM

### Steps To Reproduce
1. Set up IRI in a way that will allow you to block and unblock API requests without restarting IRI (using `nginx` proxy pass, for example)
2. Block API requests and then start IRI
3. Allow it to run for a few minutes. The solid milestones should be increasing
4. Unblock API requests and send a `getTransactionsToApprove` request
5. The solid milestones will stop updating, even after 5 minutes

### Expected behavior
Normal behavior, solid milestones should gradually increase

### Actual behavior
Solid milestones stop increasing after `getTransactionsToApprove`. The [transaction](https://thetangle.org/transaction/SDREQJXFDDFPMMORVAOOEIHXZOFBHINMJUJJLQMQLCOETTEXQTTTFFLYVJKCXHMREZDJSXEJBVJPZ9999) that IRI began looking for tips was a milestone issued almost 2 days ago. I'm not sure if this is the expected behavior.

### Errors
Upon stopping IRI:
```
06/28 22:07:56.479 [Shutdown Hook] DEBUG io.undertow - stopping undertow server io.undertow.Undertow@7d8995e
06/28 22:07:57.239 [pool-3-thread-4] INFO  com.iota.iri.network.Node - Shutting down Process Received Data Thread
06/28 22:07:57.240 [pool-3-thread-5] INFO  com.iota.iri.network.Node - Shutting down Reply To Request Thread
06/28 22:07:57.240 [pool-3-thread-1] INFO  com.iota.iri.network.Node - Shutting down Broadcaster Thread
06/28 22:07:57.479 [pool-3-thread-3] INFO  com.iota.iri.network.Node - Shutting down Neighbor DNS Refresher Thread
[...]

PR #1530, opened 2019-07-20
Fix: Rewrites request queue logic 

luca-moser:
# Description
* Bumps milestone solidification queue size to 20
* Decreases the milestone solidifier interval to 50 milliseconds
* Removes random chances of requesting milestones over normal transactions
* Removes separate milestone request queue
* Removes random drop chance and now always removes the transaction from the request queue when getting it
* Removes non used protocol config options
* Removes the `TransactionsRequesterWorkerImpl` as it serves no longer any purpose
* Massively speeds up synchronization.

183 milestones in ~3 minutes
https://ybin.me/p/945bc7782b03df1c#pwA95g/uJLLYezIfoOl8wBPXvaJ8QJaB0RTS14mwBpw=

150 milestones in ~4 minutes
https://ybin.me/p/7384053296409932#uMwaPYLjaDNW70N5XrtpTxtxILG+QOu5yBgnpCX+Zk4=

109 milestones in ~2 minutes
https://ybin.me/p/295dfc14449950b0#nVMltY/zRxv/JepGgZUZ8/QvDi7Q7g63n0sAkCm3BOM=

## Type of change

# How Has This Been Tested?

Setup a node and let it sync.

# Checklist:


src/main/java/com/iota/iri/network/TransactionRequester.java
```diff
@@ -103,61 +80,32 @@ void popEldestTransactionToRequest() {
...
             // repeat while we have transactions that shall be requested
-            while (requestSet.size() != 0) {
+            if (transactionsToRequest.size() != 0) {
                 // get the first item in our set for further examination
-                Iterator<Hash> iterator = requestSet.iterator();
+                Iterator<Hash> iterator = transactionsToRequest.iterator();
                 hash = iterator.next();
-
-                // if we have received the transaction in the mean time ....
-                if (TransactionViewModel.exists(tangle, hash)) {
```
GalRogozinski: So you're saying that our chances of receiving a tx that we already have are so low, that it will be a shame to incur I/O costs?

I definitely agree, especially now that we only request what we are missing

src/main/java/com/iota/iri/network/TransactionRequesterWorker.java
```diff
@@ -1,32 +0,0 @@
...
-     * Starts the background worker that automatically calls {@link #processRequestQueue()} periodically to process the
-     * requests in the queue.
-     */
-    void start();
-
-    /**
-     * Stops the background worker that automatically works through the request queue.
-     */
-    void shutdown();
-}
```
GalRogozinski: Yay!!
Code is gone :-)

src/main/java/com/iota/iri/network/impl/TransactionRequesterWorkerImpl.java
```diff
@@ -1,183 +0,0 @@
...
-    @VisibleForTesting
-    TransactionViewModel getTransactionToSendWithRequest() throws Exception {
-        Hash tip = tipsViewModel.getRandomSolidTipHash();
-        if (tip == null) {
-            tip = tipsViewModel.getRandomNonSolidTipHash();
-        }
-
-        return TransactionViewModel.fromHash(tangle, tip == null ? Hash.NULL_HASH : tip);
-    }
-}
```
GalRogozinski: Yay!!
Patchy worker is gone :-)

src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
```diff
@@ -30,12 +30,12 @@
     /**
      * Defines the amount of milestones that we "simultaneously" try to solidify in one pass.
      */
-    private static final int SOLIDIFICATION_QUEUE_SIZE = 10;
+    private static final int SOLIDIFICATION_QUEUE_SIZE = 20;
```
GalRogozinski: Question:
How did you come to the conclusion that 20 is better than 10?

Not that I think that 10 is such a good number... I am actually not sure what this value should be so any answer here will be good

src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
```diff
@@ -30,12 +30,12 @@
...
 
     /**
      * Defines the interval in which solidity checks are issued (in milliseconds).
      */
-    private static final int SOLIDIFICATION_INTERVAL = 100;
+    private static final int SOLIDIFICATION_INTERVAL = 50;
```
GalRogozinski: Can you just change in the description: 
"*Increases* the milestone solidifier interval to 50 milliseconds" -> "*Decreases* the milestone solidifier interval to 50 milliseconds"

Or you can write that you increase the rate, if you think "increase" sounds more positive...

I just got confused and so will others :-P

src/main/java/com/iota/iri/service/API.java
```diff
@@ -423,7 +423,7 @@ private AbstractResponse checkConsistencyStatement(List<String> transactionsList
      * @return <tt>false</tt> if we received at least a solid milestone, otherwise <tt>true</tt>
      */
     public boolean invalidSubtangleStatus() {
-        return (snapshotProvider.getLatestSnapshot().getIndex() == snapshotProvider.getInitialSnapshot().getIndex());
+        return latestMilestoneTracker.getLatestMilestoneIndex() -1 > snapshotProvider.getLatestSnapshot().getIndex();
```
GalRogozinski: Change Request:
I think the change is good. I just don't think it should be part of this PR.

Can you revert this and open a new PR to fix #844?
Also can you please change the javadoc of the method?

Thanks

src/main/java/com/iota/iri/network/TransactionRequester.java
luca-moser: Yes, no need to do any I/O.

src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
luca-moser: I figured it can solidify more milestones per solidifier step, so I simply doubled it.

src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
luca-moser: Changed

src/main/java/com/iota/iri/service/API.java
luca-moser: To be quite frank, this is super unnecessary from my perspective. Anyway, here you go: https://github.com/iotaledger/iri/pull/1533

src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
GalRogozinski: Sounds reasonable :-)

src/main/java/com/iota/iri/conf/NetworkConfig.java
```diff
@@ -36,13 +36,6 @@
      */
     boolean isAutoTetheringEnabled();
 
-    /**
-     * Default Value: {@value BaseIotaConfig.Defaults#P_REMOVE_REQUEST}
-     *
-     * @return {@value NetworkConfig.Descriptions#P_REMOVE_REQUEST}
```
GalRogozinski: Just noticed the Description still exists if you scroll down
Please remove it

luca-moser: Changed

src/main/java/com/iota/iri/service/API.java
GalRogozinski: There are several reasons for asking this:
1. When people (not just me) review a change that is supposed to change a certain behavior they usually want to focus only on the relevant changes to the specified behavior. If people will want to understand how #844 was solved, instead of seeing a small change they will see a big change. 
2. When you go back through history after a few months (or years) to understand when and why something has changed, you want history to be organized.
3. Smaller PRs get reviewed and merged faster in general. 
4. When one does smaller PRs he usually pays more attention to details
5. If a PR changes more than one behavior, and one of them causes a discussion or disagreement, then the other change also gets delayed.

You can maybe argue that in this case some points are less relevant, but in my experience it is best to abide by those rules of thumb because it hard to distinguish when those points will be relevant or not.

luca-moser: >When people (not just me) review a change that is supposed to change a certain behavior they usually want to focus only on the relevant changes to the specified behavior. If people will want to understand how #844 was solved, instead of seeing a small change they will see a big change.
I really don't get this. Is this really asking too much?

Please just make sure that both of the PRs are pulled into the same release as there was a reason why it was one unit of fixture for this particular bug (bad syncing).

GalRogozinski: > I really don't get this
I just think that under the umbrella term "syncing issues" you can have lots of changes. It doesn't make it all the same issue.

> Please just make sure that both of the PRs are pulled into the same release
This will happen

PR #1533, opened 2019-07-21
Fix: prevents certain HTTP API calls getting executed while the not is not synced

luca-moser:
# Description
Changes the criteria of an "invalid subtangle" in the `getTransactionsToApprove` call to: `latest known milestone - 1 > latestSnapshot index`. The reason to not use `latest known milestone - max depth > latestSnapshot index` (https://github.com/iotaledger/iri/issues/844#issuecomment-407339372) is because a max depth of 15 imposes too many milestones in order for the current TSA to finish in time. `-1` is used because a solid/synced node will ever be only one 1 milestone behind the most recent one for a short period of time, while the LSM tracker didn't pick it up. This is done, because calls to `getTransactionsToApprove` block the LSM tracker from setting a milestone as solid, thus, hindering the synchronization process to proceed (https://ybin.me/p/d27ff4a590b815bc#OrfJGQaeJy+DjJQZbQgn6cSgTkwn82LgkvhDxCSEfUY=).

Fixes # (issue)
https://github.com/iotaledger/iri/issues/844

## Type of change

# Checklist:


src/main/java/com/iota/iri/service/API.java
```diff
@@ -417,13 +417,14 @@ private AbstractResponse checkConsistencyStatement(List<String> transactionsList
...
+     * Checks whether the node is synchronized by comparing the latest known milestone index against the current
+     * latest solid milestone index. We allow for max. one milestone delta as a headroom, as a milestone is
+     * only applied in a set interval to the current ledger state.
      *
-     * @return <tt>false</tt> if we received at least a solid milestone, otherwise <tt>true</tt>
+     * @return true if the node is synced, false if not
      */
-    public boolean invalidSubtangleStatus() {
-        return (snapshotProvider.getLatestSnapshot().getIndex() == snapshotProvider.getInitialSnapshot().getIndex());
+    public boolean isNodeSynchronized() {
```
GalRogozinski: Good name change!

jakubcech: Indeed

jakubcech:
Great change! The validity of making these API calls when the node is not in sync aside, this should be in the same release as general improvements to the syncing mechanism.

src/main/java/com/iota/iri/service/API.java
```diff
@@ -417,13 +417,14 @@ private AbstractResponse checkConsistencyStatement(List<String> transactionsList
...
+        return snapshotProvider.getLatestSnapshot().getIndex() >= latestMilestoneTracker.getLatestMilestoneIndex() -1;
```
GalRogozinski: Looking now more closely (not rushing like last time):
The original logic covered the case where a node attempted to sync from the global/local snapshot and no milestones were yet to be received. So it covered the case where a node had communication problems with the neighbors, or the short time it takes to receive the first milestone during gossip.

The idea originally was to check that the node got solid on at least one milestone. The new logic will return `true` in this case since in the beginning the `latestSnapshot` should be equal to `latestMilestoneIndex` (the global/local genesis milestone index).

So it is an edge case, but I think it should be covered.

So I suggest two things:
1. Combine the 2 conditions
```suggestion
        return (snapshotProvider.getLatestSnapshot().getIndex() != snapshotProvider.getInitialSnapshot().getIndex()) && (snapshotProvider.getLatestSnapshot().getIndex() >= latestMilestoneTracker.getLatestMilestoneIndex() -1);
```
2. Adding a unit test for this method testing the conditions:
- before milestones are received
- after milestones are received but not synced
- after milestones are received and synced

GalRogozinski: Maybe don't commit the suggestion via github because it might create a long line :sweat_smile:

Issue #1541, opened 2019-07-25
Transactions needed to solidify milestones may be deemed invalid by their timestamp


luca-moser:
### Bug description
Given the new request queue logic via https://github.com/iotaledger/iri/pull/1530, transactions might be deemed to be invalid by their timestamp, because the timestamp validation logic bypasses the checks for transactions which are in the request queue. However, since transactions are removed automatically up on getting them from the request queue, the check to bypass the timestamp validation fails. This could lead to nodes only slowly syncing certain milestones because a tiny time frame must be hit where the transaction was put back into the request queue by the milestone solidifier and the transaction got received by a previous request.

### Solution
Add a `recentlyRequestedTxs` set (an unbounded set) to the TransactionRequester which memorizes which transactions were pulled form the request queue, in order to bypass the timestamp validation code by checking whether the transaction is in the unbounded set.
The entry is removed from `recentlyRequestedTxs` when the transaction was received.

PR #1543, opened 2019-07-25
Fix: Adds recently requested txs set to TransactionRequester

luca-moser:
# Description
Fixes transactions to be deemed invalid by their timestamp because the bypass via the `transactionRequester.isTransactionRequested()` doesn't work with the new request queue logic via https://github.com/iotaledger/iri/pull/1530

Fixes # (issue)
https://github.com/iotaledger/iri/issues/1541

## Type of change

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/network/TransactionRequester.java
+++ b/src/main/java/com/iota/iri/network/TransactionRequester.java
@@ -1,9 +1,7 @@
 package com.iota.iri.network;
 
 import java.security.SecureRandom;
-import java.util.Iterator;
-import java.util.LinkedHashSet;
-import java.util.Set;
+import java.util.*;
 
 import org.apache.commons.lang3.ArrayUtils;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


src/main/java/com/iota/iri/network/TransactionRequester.java
```diff
@@ -1,9 +1,7 @@
 package com.iota.iri.network;
 
 import java.security.SecureRandom;
```
GalRogozinski: Hmm on my IDE I see that it is not being used.
Since you changed imports automatically I would think that it would have been removed automatically.

It is out of context of this PR, so I don't care too much

src/main/java/com/iota/iri/network/TransactionRequester.java
```diff
@@ -20,6 +18,7 @@
 
     private static final Logger log = LoggerFactory.getLogger(TransactionRequester.class);
     private final Set<Hash> transactionsToRequest = new LinkedHashSet<>();
+    private final Set<Hash> recentlyRequestedTransactions = Collections.synchronizedSet(new HashSet<>());
```
GalRogozinski: The other option would be
`ConcurrentHashMap.newKeySet()` which is more performant under heavy contention.

However, I don't think we have too much contention here, so you probably made the more performant choice :+1:

Issue #1565, opened 2019-08-09
Allow merging data from multiple results from multiple persistence providers


ovanwijk:
### Description
For selective permanodes I want to use a separate persistence provider. The problem is that the Tangle.java returns the moment it finds data. This is fine for transactions because they are identified by their hash and thus both providers should return the same. But for addresses it might return different results. Like the selective permanode persistence layer might contain other information then the normal persistence layer. For addresses this needs to be merged before returning as a function.

The persistable interface already defines a method called 'merge()' that returns a boolean if it is mergable or not. Only implementations that implement 'Hashes' are mergable because they hold a list of hashes. Addresses is one of them.

I propose is to change the 'load' behavior in the Tangle.java class to merge results from all persistence providers when this is possible.

### Motivation
To support separate persistence providers.

### Requirements
Create a list of what you want this feature request to fulfill.

### Open Questions (optional)
Discuss on how to do this best, and if this is the location to change it.

### Am I planning to do it myself with a PR?
I could

ovanwijk:
In my own forked branch I solved it like this:

https://github.com/ovanwijk/iri/blob/27fa8219e79eba7b8c46292735951c116beecd32/src/main/java/com/iota/iri/storage/Tangle.java#L68-L96

Please note the introduction of two new functions isEmpty() and mergeTwo() to handle this properly.


As seen here: https://github.com/iotaledger/iri/blob/d658b1fd1edf58d4e9ff1ecf034aa25cf1cb3d02/src/main/java/com/iota/iri/storage/rocksDB/RocksDBPersistenceProvider.java#L153-L162

Is that .get will never return null.

Not all Persistable.bytes() implementations are safe to use. (milestone) 

If this is a suitable way to solve it I can create a separate branch just for this issue to handle it this way!

Love to hear a comment from: @GalRogozinski and/or @jakubcech

GalRogozinski:
I think the idea is fine. I am just not sure that any object that is "mergeable" in the sense that is currently defined should be "mergeable" in the way you define it. For example, should tags be merged from the other db?

If you agree with me then maybe create another method (can't think of a good name right now). You can even go for using annotations (I have a vision in my head to start putting annotations on persistable objects). I think the code will look nicer this way.

Also make sure you follow style guidelines:
https://github.com/iotaledger/iri/blob/dev/STYLEGUIDE.md#formatting

Issue #1307, opened 2019-01-28
Prune transaction metadata while pruning


GalRogozinski:
### Description
We store in the db a mapping between addresses and transaction hashes that spent/deposited iotas to the address.

When the pruning of transactions occur, we still keep the mapping of addresses to transactions.
We should delete the transaction hashes from the mapping that were pruned.

### Motivation
We load a huge number of irrelevant hashes that may cause OOM errors. This new feature should relieve the problem.

### Requirements
1. Once we are done with a pruning operation we should delete all the pruned transactions' hashes from the `AddressViewModel` db.

jakubcech:
✔️ by Trinity.

ovanwijk:
I agree with this as well! We should also consider the BundleViewModel, TagViewModel and AproveeViewModel. These are similar indexes.

As an idea, to prevent more costly constructions of the 'Hashes' class and loading into a set to delete it I already implemented an efficient method for just cutting out the 'indexable' part from the saved set on a separate branch.

https://github.com/ovanwijk/iri/blob/d69046319ed2656a2ef7f1f4bf03eaf2598ca517/src/main/java/com/iota/iri/storage/rocksDB/RocksDBPPPImpl.java#L311-L338

jakubcech:
We should break this down and include the different models, as it's already quite an expensive task at this point. 

>As an idea, to prevent more costly constructions of the 'Hashes' class and loading into a set to delete it I already implemented an efficient method for just cutting out the 'indexable' part from the saved set on a separate branch.

👍

PR #1579, opened 2019-08-22
Fix: Persistables merge and load functionality

ovanwijk:
# Description

- Introduced mergeTwo() on persistables to generically allow merging of different results form persistenceProviders
- Introduced isEmtpy() on persistables to allow checking of empty results from the persistenceProviders (result != null never works because .get() cannot return null)
- Combined those methodes in Tangle.load() so it now properly allows gets() from multiple persistence Providers.

Fixes # (issue)

#1565 

## Type of change

# How Has This Been Tested?

Ran all tests, nothing breaks. Code has been running in my own testnet successfully.

# Checklist:


src/main/java/com/iota/iri/model/persistables/Milestone.java
```diff
@@ -54,8 +54,18 @@ public void readMetadata(byte[] bytes) {
...
     }
 
+
     @Override
     public boolean merge() {
         return false;
     }
+    @Override
+    public Persistable mergeTwo(Persistable nrTwo){
+        return null;
```
GalRogozinski: Can you change to ` throw new OperationNotSupportedException("This object is not mergeable");

src/main/java/com/iota/iri/storage/Persistable.java
```diff
@@ -47,4 +48,19 @@
      * false if we should replace the old value
      */
     boolean merge();
+
+    /**
+     * Merges source object into this object.
+     * @param source
```
kwek20: the persistable that will be merged into the called persistable

src/main/java/com/iota/iri/storage/Persistable.java
```diff
@@ -47,4 +48,19 @@
...
+     * @return
```
kwek20: the updated persistable, which contains both. Or {@code null} if the persistable cannot be merged

src/main/java/com/iota/iri/storage/Tangle.java
```diff
@@ -82,13 +83,38 @@ public void shutdown() throws Exception {
...
+                return a.mergeInto(b);
+            }catch (OperationNotSupportedException e){
+                log.error("Error merging data: ", e);
+                log.error("Error merging data: ", e);
+                return null;
+            }
+        });
+
+        if (p == null) {
+            p = (Persistable) model.newInstance();
```
kwek20: We used to return null in this case, is there a reason you need an empty instance?

src/main/java/com/iota/iri/storage/Tangle.java
```diff
@@ -82,13 +83,38 @@ public void shutdown() throws Exception {
...
+                    outlist.add(result);
+                } else {
+                    // If it is a non-mergeable result then there is no need to ask another provider again.
+                    // return immediately
+                    return result;
                 }
             }
-            return out;
+        }
+        Persistable p = outlist.stream().reduce(null, (a, b) -> {
```
kwek20: If we were to cache the persistables, merging them would result in modification of the original. 
I  think its best to change the null into 'model.newInstance()'

src/main/java/com/iota/iri/storage/Tangle.java
```diff
@@ -82,13 +83,38 @@ public void shutdown() throws Exception {
...
+            if (a == null) {
+                return b;
+            }
+            try {
+                return a.mergeInto(b);
+            }catch (OperationNotSupportedException e){
+                log.error("Error merging data: ", e);
```
kwek20: Duplicate error logging.

src/main/java/com/iota/iri/storage/Tangle.java
ovanwijk: Actually there was never null returned here. Because persistence provider.get never returns null. It is the reason why this change is needed. But that code is in there for backward compatibility. Could go with a comment though.

kwek20: Aha, yea youre right. If you do the suggestion below, it will never be null anyways, so the check can be removed

src/main/java/com/iota/iri/storage/Tangle.java
GalRogozinski: and a missing space after `}`

src/main/java/com/iota/iri/storage/Tangle.java
ovanwijk: Well I think it should be possible to be null actually but the check is there 'outside' for the reason it looks 'wrong' maybe not the right way to do it, but neither is including class instantiation just because. : 
https://github.com/iotaledger/iri/blob/6769054eb05b07639d714d67eeb34d022e5fe1bd/src/main/java/com/iota/iri/storage/rocksDB/RocksDBPersistenceProvider.java#L154-L155

But really the right way would be to let the persistence provider return null if it didn't find anything and fix everything it breaks. But that will be a different issue then.

I would advise for maybe a new issue describing it and add it to the comments on why this is necessary.

GalRogozinski: https://github.com/iotaledger/iri/issues/1591

I chose to create an empty immutable constant object just to avoid NullPointerExceptions

src/main/java/com/iota/iri/storage/Tangle.java
ovanwijk: Is caching seriously considered now? Because there are a lot of arguments to be made to not do index caching (addresses, tags, bundles etc). Is it relevant now?

GalRogozinski: Well...
Currently it is evicted from our priority queue.
However, it is a good practice imo to write robust code that won't break.

We can come to a compromise where we write a comment on the caching issue so that if we ever start working on it we won't forget this change.

Btw, feel free to be explicit aout your arguments against caching

src/main/java/com/iota/iri/storage/Persistable.java
```diff
@@ -46,5 +47,21 @@
      * @return <tt>true</tt> if we should add the object to a list for an existing key. <tt>false</tt> if we should
      * false if we should replace the old value
      */
-    boolean merge();
```
GalRogozinski: I believe the javadocs here can also be updated. Now `canMerge` means two things:
1. We want to merge data on the same persistence provider
2. We want to merge data from another persistance provider

It happens so that when we want (1) we also want (2).
Anyhows the javadocs only pertain to (1).

docs are also important :-)

src/main/java/com/iota/iri/storage/Tangle.java
ovanwijk: Sure! There I go:

Currently, indexes are not being updated after deletion of the underlying data through snapshotting which is described here: https://github.com/iotaledger/iri/issues/1307.

So for any caching to represent data I think this is important to have fixed first. Then on caching: transaction caching is easy because the underlying data won't change. While for 'canMerge'/mergable persistables this can change. There are basically two approaches you can go here when an update arrives on the index(address/tags etc). 

1) Update it on disk and in memory/cache at the same time. 
2) Just evict it from the cache and update on disk.

Both shouldn't cause any trouble with that code.

Another problem would come from the ViewModel .store function and then the tangle no longer knows where to update this. This is generally true but: The .store function is never used, anywhere, not even in tests. Also imo a ViewModel should not have a store function (it is a View after all), that it is not used anywhere points to me this was not really intended either.

Indexes are a function of when a transaction is stored or deleted. Because of this, I think there will be no problem whatsoever with this code in the future. Even with caching.

GalRogozinski: Note that transaction metadata like `solidity` or `snapshotIndex` can change.

I think that if an update arrives on you can just update the cache. There will be a discrepancy between the disk and cache, but this won't matter too much. The cache should be evicted to disk in a way so that if there is a failure, then upon restarting the node should still be able to sync correctly and obtain from its neighbors all the lost data.

I agree with what you said about the `ViewModel`, and in fact I want to get rid of it entirely.

ovanwijk: So solidity and snapshotIndex are non-mergable persistables. (the only mergable persistables are indexes).

So when it comes to caching, it is a 'problem' on its own and not related to the 'mergablity' of an object?

PS When it comes to caching: updating on disk and just evicting from the cache and then repopulate the cache when it is required -for me- has always worked like a charm. There is some overhead but the simple approach usually makes it very manageable code-wise because there are much fewer edge-cases to handle.

src/main/java/com/iota/iri/storage/Persistable.java
ovanwijk: Yeah the doc addition surprised me a bit. I will edit it.

ovanwijk: However, for the docs: for my idea is that more generally says if two persistables of the same type can merge with each other.

"Specifies whether two objects of the same type can be merged and their merged result used as a single object. For storing in a persistence provider this means we can append items to an index and for retrieving this means we can merge the results of multiple storage providers"

src/main/java/com/iota/iri/storage/Tangle.java
```diff
@@ -82,13 +83,37 @@ public void shutdown() throws Exception {
...
             }
-            return out;
+        }
+        Persistable p = outlist.stream().reduce(null, (a, b) -> {
+            if (a == null) {
+                return b;
+            }
+            try {
+                return a.mergeInto(b);
+            }catch (OperationNotSupportedException e) {
```
GalRogozinski:
```suggestion
            } catch (OperationNotSupportedException e) {
```

missing space

GalRogozinski:
I see a buildkite regression test failing

PR #1588, opened 2019-08-30
Fix: Fixes regression introduced through the bundle validator refactor

luca-moser:
# Description
Through a redundant break statement, the balance diff creating method did return before it actually computed any balance diff, leading to every applied milestone to not change any addresses.

Weird is that no unit or regression tests catched this. We should have a regression test which mutates addresses and then does a `getBalances()` call to check that the address really was mutated (I didn't check whether we already have such regression test).

## Type of change

# How Has This Been Tested?

Applied this patch to my localsnapshot/spent-addresses RocksDB PR and verified that the mutated balances were not zero anymore.

# Checklist:


GalRogozinski:
Buildkite restarted now...
Once it passes this will be merged.

Since I see this as high priority then #1590 is also a priority

PR #1589, opened 2019-08-31
fixes issue #1566

### LINKED ISSUE #1566: Wrong data length calculation in AbstractHash
### Bug description
The class AbstractHash has the line in the constructor:

`System.arraycopy(source, sourceOffset, dest, 0, sourceSize - sourceOffset > source.length ? source.length - sourceOffset : sourceSize);`

It looks like a misspelling: "sourceSize - sourceOffset" instead of  "sourceSize + sourceOffset"

### IRI version
1.5.6 - 1.8.1

spangin:
# Description

Fixes #1566
Correct data setting into Hash object:
- source data is cut if it's too long
- 0s are added if source data is too short. I.e. padding for some data (e.g. Tag) can be removed.

## Type of change

# How Has This Been Tested?

All existing unit-tests are passed.

# Checklist:


src/main/java/com/iota/iri/model/AbstractHash.java
```diff
@@ -37,7 +37,7 @@ public AbstractHash() {
     public AbstractHash(byte[] source, int sourceOffset, int sourceSize) {
         if(sourceSize < SIZE_IN_TRITS) {
             byte[] dest = new byte[SIZE_IN_BYTES];
-            System.arraycopy(source, sourceOffset, dest, 0, sourceSize - sourceOffset > source.length ? source.length - sourceOffset : sourceSize);
+            System.arraycopy(source, sourceOffset, dest, 0, Math.min(dest.length, Math.min(source.length, sourceSize)));
```
GalRogozinski: I think you still need `source.length -sourceOffset`
```suggestion
            System.arraycopy(source, sourceOffset, dest, 0, Math.min(dest.length, Math.min(source.length - sourceOffset, sourceSize)));
```

I don't we currently really use this size optimization. But maybe we will so might as well add support.

Would you like to wrap this line in a method with `default` visibility, add the `@VisibleForTesting ` annotation, and add a unit task.
Since you are fixing a bug, we need to write a unit test so that the bug will never repeat itself.

spangin: Hash object is just a storage for data. The constructor parameters (`source`, `sourceOffset `and `sourceSize`) are description of the data source.
Hash object can't validate data. It can only check data syntax: the data array is not null and data coordinates (`sourceOffset `and `sourceSize`) are valid for this array. I.e. if syntax is incorrect, the hash object throws the corresponding exception: NullPointerException or ArrayIndexOutOfBoundsException.
If hash object tries to fix incorrect data source (`Math.min(source.length - sourceOffset, sourceSize)`), it can be a reason of incorrect data using and a reason of a bug that hard to find.
I.e. my point of view: if `sorceOffset + sorceSize > source.length`, ArrayIndexOutOfBoundsException has to be thrown to inform about the bug. Anyway you take a decision :)

GalRogozinski: I agree with you

GalRogozinski: You want to add the exception?

spangin: In this situation the adding 
`if (sourceOffset + sourceSize > source.length) throw new ArrayIndexOutOfBoundsException();`
to the code is incorrect, because IOTA uses
https://github.com/iotaledger/iri/blob/9f95ef7da5d6bc545555a7ad525cc64968111bc9/src/main/java/com/iota/iri/model/HashFactory.java#L79
and
https://github.com/iotaledger/iri/blob/9f95ef7da5d6bc545555a7ad525cc64968111bc9/src/main/java/com/iota/iri/model/HashFactory.java#L100
As you can see the methods don't get `sourceSize` as input parameter and use `Hash.SIZE_IN_TRITS` or `Hash.SIZE_IN_BYTES` as values for `sourceSize` to invoke the corresponding constructor.

BTW, I have checked the constructor with the code
`if (sourceOffset + sourceSize > source.length) throw new ArrayIndexOutOfBoundsException();`
and got errors for existing unit-tests:
```
TransactionViewModelTest.findShouldBeSuccessful:339 » ArrayIndexOutOfBounds
TransactionViewModelTest.findShouldReturnNull:352 » ArrayIndexOutOfBounds
```

As a result I suggest using the code I provided in this PR.

Issue #1536, opened 2019-07-21
getNeigbors 'domain' attribute should contain IP if empty


codepleb:
### Description
`getNeighbors` returns IPs and domains. Sometimes, for whatever reason, `domain` is not there. It should IMHO always be there and contain the IP in those cases.

### Motivation
On the tooling side, this causes a lot of nullchecks, that somehow look so unnecessary.

### Requirements
`IF domain === undefined THEN domain = address`

### Am I planning to do it myself with a PR?
No

jakubcech:
Is the domain field missing completely or is the issue with the returned value being domain="" ?

TeeVeeEss:
The domain field is missing completely, causing confusing output in tools like Iota peer manager. As soon as the domain name is not in the json, ipm shows 2 graphs of the neighbor: 1 with the ip-address as title and 1 with the dns-name as title. The one with the dns-name is shown as dead neighbor with 0 tx. The node operator acting on this visual sign could take the wrong conclusion and will delete a healthy nbr.

jakubcech:
+ regression test fixes.

jakubcech:
#1580

luca-moser:
I will apply following logic to the fix:
* if the neighbor is connected and there's no mapping to a domain, the IP address will be set in the `domain` field
* if the neighbor is connected and there's a mapping to a domain, the domain will be set in the `domain` field
* if the neighbor is not connected, the neighbor was added using an URI with a domain (tcp://example.com:15600) and during the call to `getNeighbors()` the domain can't be resolved to an IP address, the `domain` field will contain the domain name and the `address` field will be empty
* if the neighbor is not connected, the neighbor was added using an URI with an IP address (tcp://1.1.1.1:15600), the `domain` field will contain the IP address and the `address` field will contain the IP address + port combination
* if the neighbor is not connected, the neighbor was added using an URI with a domain (tcp://example.com:15600) and during the call to `getNeighbors()` the domain could be resolved to an IP address, the `domain` field will contain the domain name and the `address` field will contain the IP address + port combination.

Tools using `getNeighbors()` should check whether the given neighbor is connected and be agnostic to that. A peer/neighbor manager should map the domain to an IP address as the identifier for the object representing the connection and if there's no IP address to be resolved, use the domain or another mechanism as a placeholder. If the underlying IP address of a domain changes, you need to update the key of your tool which identifies the connection.

GalRogozinski:
Each one of the above bullet points is a regression test

PR #1604, opened 2019-09-06
Fix: Always set the domain field of a neighbor

luca-moser:
# Description
This PR alters the logic of neighbor properties and the `getNeighbors()` API call to fulfill following rules:
* if the neighbor is connected and there's no mapping to a domain, the IP address will be set in the `domain` field
* if the neighbor is connected and there's a mapping to a domain, the domain will be set in the `domain` field
* if the neighbor is not connected, the neighbor was added using an URI with a domain (tcp://example.com:15600) and during the call to `getNeighbors()` the domain can't be resolved to an IP address, the `domain` field will contain the domain name and the `address` field will be empty
* if the neighbor is not connected, the neighbor was added using an URI with an IP address (tcp://1.1.1.1:15600), the `domain` field will contain the IP address and the `address` field will contain the IP address + port combination
* if the neighbor is not connected, the neighbor was added using an URI with a domain (tcp://example.com:15600) and during the call to `getNeighbors()` the domain could be resolved to an IP address, the `domain` field will contain the domain name and the `address` field will contain the IP address + port combination.

The `domain` field of a neighbor always either contains the domain name of the initial URI or the IP address. It is never empty.

Additionally, this PR stops the node from doing reverse lookups of IP addresses to gather a domain name, as often a not desired domain will be set.

Fixes # (issue)
https://github.com/iotaledger/iri/issues/1536

## Type of change

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/network/NeighborRouter.java
+++ b/src/main/java/com/iota/iri/network/NeighborRouter.java
@@ -289,6 +289,8 @@ public class NeighborRouter {
             String domain = ipToDomainMapping.get(remoteAddr.getAddress().getHostAddress());
             if (domain != null) {
                 newNeighbor.setDomain(domain);
+            }else{
+                newNeighbor.setDomain(remoteAddr.getAddress().getHostAddress());
             }
             newNeighbor.send(Handshake.createHandshakePacket((char) networkConfig.getNeighboringSocketPort(),
                     byteEncodedCooAddress, (byte) protocolConfig.getMwm()));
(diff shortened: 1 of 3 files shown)


codepleb:
Thank you very much luca!

GalRogozinski:
Looks good!

If I am not mistaken, most of the regression tests for this bug can be created easily without adding any python. I think currently we can only add nodes via their IP though...

For the other cases where the neighbor is not added with the domain it should be a breeze

jakubcech:
👍 on removing the reverse lookup

GalRogozinski:
Even though simple regression tests should be added to bugs we solve (especially when they are trivial) I am opening #1611 so the team can focus on tip selection.

In the meanwhile I'll be merging this before we will have conflicts to solve.

PR #1608, opened 2019-09-10
Update regression node names, m6 times

DyrellC:
# Description
Simple update to node labels in regression tests for log artefact creation during buildkite testing. This will allow for easier debugging of tests in the future. This also includes a reduction in the machine6 wait times. 

## Type of change
-Update to regression files 

# How Has This Been Tested?
Purely syntax changes

### DIFF
--- a/python-regression/tests/features/machine2/config.yml
+++ b/python-regression/tests/features/machine2/config.yml
@@ -15,8 +15,5 @@ defaults: &transaction_tests_config_files
   java_options: -agentlib:jdwp=transport=dt_socket,server=y,address=8000,suspend=n -javaagent:/opt/jacoco/lib/jacocoagent.jar=destfile=/iri/jacoco.exec,output=file,append=true,dumponexit=true
 
 nodes:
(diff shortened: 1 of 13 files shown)


python-regression/tests/features/machine2/config.yml
```diff
@@ -15,8 +15,5 @@ defaults: &transaction_tests_config_files
   java_options: -agentlib:jdwp=transport=dt_socket,server=y,address=8000,suspend=n -javaagent:/opt/jacoco/lib/jacocoagent.jar=destfile=/iri/jacoco.exec,output=file,append=true,dumponexit=true
 
 nodes:
-  nodeA: #name
-    <<: *transaction_tests_config_files
-
-  nodeB:
```
GalRogozinski: You said purely a synactic change in the description.
Why is a node removed?

DyrellC: This node is not actually used in the test so it can be removed to save resources.

PR #1612, opened 2019-09-11
Fix: Ensure reference transaction is valid when alpha is 0

GalRogozinski:
## Description

In #1596 we introduced a bug. If one calls GTTA with the reference parameter, and no CW map was created, then we get an `InvalidAlgorithmParaemterException` since IRI assumes that we chose to promote a transaction that is not in the cone of our entrypoint milestone.

To fix this bug we do full validation on the `reference` transaction:
- It exists
- It is solid
- It is a tail (I don't think it was a requirement before, but as far as I am aware we should only promote tails)
- It is not below max depth
- It is consistent with our first tip.

## Type of change

# How Has This Been Tested?

Tests will be added before the merge

# Checklist:

- [ ] My code follows the style guidelines for this project

### DIFF
--- a/src/test/java/com/iota/iri/service/tipselection/impl/TipSelectorImplTest.java
+++ b/src/test/java/com/iota/iri/service/tipselection/impl/TipSelectorImplTest.java
@@ -1,13 +1,15 @@
 package com.iota.iri.service.tipselection.impl;
 
 import com.iota.iri.conf.BaseIotaConfig;
-import com.iota.iri.conf.MainnetConfig;
 import com.iota.iri.conf.TipSelConfig;
 import com.iota.iri.model.Hash;
 import com.iota.iri.model.HashFactory;
 import com.iota.iri.service.ledger.LedgerService;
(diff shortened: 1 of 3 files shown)


src/test/java/com/iota/iri/service/tipselection/impl/TipSelectorImplTest.java
```diff
@@ -0,0 +1,101 @@
...
+
+    @Mock
+    private Tangle tangle;
+
+    @Mock
+    private SnapshotProvider snapshotProvider;
+
+    @Mock
+    private TipSelConfig config;
+
```
GalRogozinski: All dependencies of TipSelector are mocked for future tests

GalRogozinski:
I deleted the regression test until changes to GTTA in PYOTA are made as defined by https://github.com/iotaledger/iota.lib.py/issues/235

GalRogozinski:
closed in favor of #1621 that will merge to release branch

PR #1633, opened 2019-10-01
RatingOne filter for alpha == 0

DyrellC:
# Description
Under the right spam conditions nodes have a tendency to spit out chains of transactions rather than a dag. The reason for this is that the alpha == 0 implementation does not filter out transactions added after the walk has started, and so, in higher tps environments, the spam will begin creating a chain of transactions due to the synchronised nature of the calls. This rating filter allows the DAG to spread out more evenly.

## Type of change

# How Has This Been Tested?
-Tested on ICC network as well as in private networks set up locally. 4 nodes with GTTA spammers on each, and the spaghettification is mitigated significantly in higher TPS environments. 

# Checklist:


luca-moser:
@DyrellC The buildkite pipeline is failing because some DB file checksums are wrong.

PR #1645, opened 2019-10-17
Fix: Correct walker behaviour for alpha == 0

DyrellC:
# Description

Currently the walker uses an `if(!Collections.EMPTY_MAP.equals(ratings))` check to determine if the alpha value is 0 in the walker. But since the ratings list is is provided by `ratingOne` this list is not empty. Therefore the node is still performing the walk as though the alpha is not 0. The improved behaviour from not doing the ratings calculation has been the result of the alpha 0 change, but having the walk skip the additional unnecessary calculations on the walk end should show another improvement to response time. 

## Type of change

# How Has This Been Tested?

-Set a break point in the walker to make sure that the walk is registering the correct alpha value and using the correct logic

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/service/tipselection/impl/WalkerAlpha.java
+++ b/src/main/java/com/iota/iri/service/tipselection/impl/WalkerAlpha.java
@@ -128,7 +128,7 @@ public class WalkerAlpha implements Walker {
         int approverIndex;
 
         //Check if ratings map is empty. If so, alpha was set to 0 and a random approver will be selected.
-        if(!Collections.EMPTY_MAP.equals(ratings)) {
+        if(alpha != 0) {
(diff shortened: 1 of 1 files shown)


src/main/java/com/iota/iri/service/tipselection/impl/WalkerAlpha.java
```diff
@@ -128,7 +128,7 @@ public Hash walk(Hash entryPoint, Map<Hash, Integer> ratings, WalkValidator walk
         int approverIndex;
 
         //Check if ratings map is empty. If so, alpha was set to 0 and a random approver will be selected.
```
kwek20: Youre probably going to stab me once or twice, but the comment above it is now wrong :)

Issue #1655, opened 2019-11-08
Node failing to solidify milestones correctly


DyrellC:
### Bug description
When trying to solidify over wide gaps, there is a possibility that the node will not properly solidify. This can be identified in two ways. Either the node will hang on the same milestone, producing an output as follows: 
```
Solidifying milestone #100 [20 / 397]
Solidifying milestone #100 [20 / 398]
Solidifying milestone #100 [20 / 399]
Solidifying milestone #100 [20 / 399]
``` 
or the solidifer fails to print a message, at all and the `latestSolidMilestone` remains the same while the `latestMilestone` continues to grow. In the latter instance, it can sometimes mean that there is one milestone in the `unsolidMilestonesPool` so no output is printed. In this case it is the same as the former scenario where the same milestone is being requested, and no further milestones are being added to the `unsolidMilestonesPool`. 

When investigating the milestones that were failing to solidify it appeared that some of the transactions were present in the db, but were not marked as milestones due to the other milestone in the bundle not being solid. In other instances the milestone was never found through `transactionValidator.checkSolidity` call.  This may be caused by a milestone being left behind/effectively orphaned during large spam events or splitting. In these cases the milestone would not be found when solidifying backwards through the tangle. 

### IRI version
v1.8.2

### Hardware Spec
Linux Mint, 8GB 4 cpu 160GB SSD  x 2 

#### Testnet
1. Start one node node with `https://s3.eu-central-1.amazonaws.com/iotaledger-dbfiles/dev/SyncTestDB.tar` and another with `https://s3.eu-central-1.amazonaws.com/iotaledger-dbfiles/dev/EmptyDB.tar`. Make sure the nodes are already neighbored. For faster syncing add extra solid nodes to the mix.  
2. Start nodes with the following configuration
`java -jar iri-1.* -p 14265 -t 15600 --zmq-enabled true --zmq-port 5556 --testnet true --testnet-coordinator EFPNKGPCBXXXLIBYFGIGYBYTFFPIOQVNNVVWTTIYZO9NFREQGVGDQQHUUQ9CLWAEMXVDFSSMOTGAHVIBH --testnet-no-coo-validation true --milestone-start 0 --mwm 1 --remote true --remote-limit-api "" --snapshot ./snapshot.txt -n 'your.neighbours.here'`
[...]

Issue #1447, opened 2019-05-13
Track milestones by index


GalRogozinski:
### Description
We want to track milestones according to index. Rather than loading all transactions that have the coordinator address we want to load all the transactions that have the current index we wish to track (use Obsolete_Tag).

### Motivation
1. Have smaller sets 
2. Perhaps will help with (investigating) #1434 

### Requirements
1. In `LatestMilestoneTrackerImpl#collectNewMilestones` load hashes  for the next milestone index you are looking for.
2. Get rid of the `seenMilestoneCandidate` set. 

### Open question
How does this affect the `milestoneTracker`?

luca-moser:
The current mechanism might also explain IRI consuming a modest amount of CPU. As seeon the ICC network, nodes start to consume a constant stream of CPU at around 50k milestones being published:
![image](https://user-images.githubusercontent.com/1385611/57718604-fda5c580-767d-11e9-81da-213908b7196c.png)

luca-moser:
The logic must be smarter for this to work. Consider someone setting up a new node at milestone 0 where the actual current one in the network is 1000. The new node doesn't have the milestones from 0-1000 and there is no way for it to request milestone 1 or 2 from the neighbors as it lacks the data in the db.

kwek20:
Yes I noticed my node with this fix didnt increase in LSMI at all, havent had time to check why. I will check your suggested solution in the PR!

GalRogozinski:
Now that compass depth is set to 0 I think this issue can be done.
This is because we will always have a chain of milestones.
When a node receives a new milestone we need to check that it is above the last index and then try to solidify it...

GalRogozinski:
A good logic can be the following:

Have 2 boot modes quick/normal:

In quick mode:
1. Try to load the latest MS
2. If it is solid and ledger state is updated then start querying the consecutive indexes from the DB until you can't find it, else solidify.
3. continue scanning for the next index every second (this is the current interval in the code, maybe we enlarge it in a different issue)

In normal mode we want to cover the following cases:
1. Maybe there are milestones that are not connected in a nice chain (happened before compass depth was set to 0). It could be that the latest MS is solid in this case but a previous one is not
2. Maybe the milestones are not consecutive (old DBs)

So what we should do in normal mode is:
1. Find latest MS
2. Start scanning by index from `initialSnapshot`.
3. Try to solidify any milestone on the way that is not solid
4. If one cannot find a milestone at a certain index that is below latest MS log a warning.
[...]

PR #1660, opened 2019-11-11
sort milestone candidates before analyzing

achabill:
# Description

Sort milestone candidates before analyzing to increase the likely hood of a new milestone being seen early.

Part of #1655 

## Type of change

# Checklist:

- [x] I have commented on my code, particularly in hard-to-understand areas


src/main/java/com/iota/iri/controllers/AddressViewModel.java
```diff
@@ -38,6 +41,34 @@ private AddressViewModel(Address hashes, Indexable hash) {
...
+     *
+     * @param tangle The tangle reference for the database to find the {@link Address} set in
+     * @param hash The hash identifier for the {@link Address} set that needs to be found
+     * @return The {@link AddressViewModel} controller generated
+     * @throws Exception Thrown if the database cannot load an {@link Address} set from the reference {@link Hash}
+     */
+    public static List<TransactionViewModel> loadAsSortedList(Tangle tangle, Indexable hash) throws Exception {
+        Address hashes = (Address)tangle.load(Address.class, hash);
+        return hashes.set.stream()
+                .map(item -> fromHash(tangle,item))
```
DyrellC: When you map with a `fromHash` call as you've done here, a tvm will always be created, which invalidates the `tvm != null` call below when sorting. What happens in this case is that the stream will throw an exception once the sorting hits a transaction that doesn't exist in the db (as is the case when a milestone has been removed from pruning). To fix this issue locally I replaced the mapping with: 
```
.map(
    item -> {
        try {
            if(tangle.exists(Transaction.class, item)) {
            return fromHash(tangle, item);
        } else {
            return null;
        }
    }catch(Exception e){
        return null;
    } 
})
```

This will make sure that if the transaction does not exist, a tvm will not be created for it. Of course this means that you will either need to not include these transactions in the collection before you analyse it in the `LatestMilestoneTrackerImpl` or have line 314 include a null check before trying to add to the queue's. 
i.e 
`if (tvm != null && seenMilestoneCandidates.add(tvm)) {`

DyrellC: @achabill  if you apply a change like this the code should pass. As is however it throws an exception the first time it encounters a pruned milestone. Which is the case in the regression test.

acha-bill: ok.

GalRogozinski:
I may be forced to look into #1447 now...
If we implement it, then this PR will become obsolete

src/main/java/com/iota/iri/controllers/AddressViewModel.java
```diff
@@ -38,6 +42,30 @@ private AddressViewModel(Address hashes, Indexable hash) {
         this.hash = hash;
     }
 
+    /**
+     * Constructor for an {@link Address} set controller from an existing {@link Address} set.
+     * The resulting hash list is sorted
```
GalRogozinski: This doc is unclear to me. This is not a java constructor...

Maybe write:
"Loads all transaction that mutate a certain address. It sorts them by the attachment timestamp"

src/main/java/com/iota/iri/controllers/AddressViewModel.java
```diff
@@ -38,6 +42,30 @@ private AddressViewModel(Address hashes, Indexable hash) {
...
+     *
+     * @param tangle The tangle reference for the database to find the {@link Address} set in
+     * @param hash The hash identifier for the {@link Address} set that needs to be found
```
GalRogozinski:
```suggestion
     * @param hash The address we are loading the transactions for
```

src/main/java/com/iota/iri/controllers/AddressViewModel.java
```diff
@@ -38,6 +42,30 @@ private AddressViewModel(Address hashes, Indexable hash) {
...
+     * @return The {@link AddressViewModel} controller generated
```
GalRogozinski: We are returning a list of transactions (actually transaction controllers unfortunately)

src/main/java/com/iota/iri/controllers/AddressViewModel.java
```diff
@@ -38,6 +42,30 @@ private AddressViewModel(Address hashes, Indexable hash) {
...
+     * @throws Exception Thrown if the database cannot load an {@link Address} set from the reference {@link Hash}
+     */
+    public static List<TransactionViewModel> loadAsSortedList(Tangle tangle, Indexable hash) throws Exception {
```
GalRogozinski: Just note (no need to change):
Usually it is best to use an interface like `Indexable` in method parameters.

In this case it would probably best to use something more specific like `AddressHash` because you are not going to be using other subtypes of `Indexable` in this method.
However at the moment casting will be neccessary to change this...

So this is just a point to think of before we start refactoring...

src/main/java/com/iota/iri/controllers/AddressViewModel.java
```diff
@@ -38,6 +42,30 @@ private AddressViewModel(Address hashes, Indexable hash) {
...
+        Address hashes = (Address) tangle.load(Address.class, hash);
+        return hashes.set.stream().map(item -> {
+            try {
+                if (tangle.exists(Transaction.class, item)) {
+                    return TransactionViewModel.fromHash(tangle, item);
+                }
+                return null;
```
GalRogozinski: why return null? Why not do a `.filter(item ->  (tangle.exists(Transaction.class, item))` so you won't have nulls in your list

src/main/java/com/iota/iri/controllers/AddressViewModel.java
```diff
@@ -38,6 +42,30 @@ private AddressViewModel(Address hashes, Indexable hash) {
...
+            } catch (Exception e) {
+                return null;
```
GalRogozinski: why are you swallowing exceptions? this shouldn't be happening

src/main/java/com/iota/iri/controllers/AddressViewModel.java
```diff
@@ -38,6 +42,30 @@ private AddressViewModel(Address hashes, Indexable hash) {
...
+            }
+        }).sorted(Comparator.comparing(tvm -> tvm != null ? tvm.getTransaction().attachmentTimestamp : 0))
```
GalRogozinski: I guess sorting by attachmentTimestamp is fine

maybe add a comment that this was an arbitrary choice (you also have timestamp which is part of the bundle essence)

DyrellC:
@acha-bill  still waiting on you to respond to Gal's review before I can review myself

src/main/java/com/iota/iri/controllers/AddressViewModel.java
```diff
@@ -38,6 +41,34 @@ private AddressViewModel(Address hashes, Indexable hash) {
...
+            }
+        }).map(item -> {
+            try {
+                return TransactionViewModel.fromHash(tangle, item);
+            } catch (Exception e) {
+                throw new RuntimeException(e);
+            }
+        }).sorted(Comparator.comparing(TransactionViewModel::getAttachmentTimestamp))
+                .collect(Collectors.toList());
+    }
```
GalRogozinski: I didn't run it, but the following compiles and looks nicer:
```
public static List<TransactionViewModel> loadAsSortedList(Tangle tangle, Indexable hash) throws Exception {
        Address hashes = (Address) tangle.load(Address.class, hash);
        return hashes.set.stream()
                .filter(ThrowingPredicate.unchecked(hash1 -> TransactionViewModel.exists(tangle, hash1)))
                .map(ThrowingFunction.unchecked(item -> TransactionViewModel.fromHash(tangle, item)))
                .sorted(Comparator.comparing(TransactionViewModel::getAttachmentTimestamp))
                .collect(Collectors.toList());
    }
```

GalRogozinski:
@karimodm 
This should now help with solidifying milestones
note that it is merged

PR #1663, opened 2019-11-14
Allow to configure benchmark from CLI

### LINKED ISSUE #1651: Make benchmarks faster
### Description
We need to speed up our benchmark tests while still providing reasonable results

### Motivation
We don't want the benchmarks clogging up our devnet resources

### Requirements
1. Make the number of forks configurable so one can control parallelism
2. Make the number of iterations configurable (warmup and measurement)

GalRogozinski:
# Description
We have added the following flags to benchmark runner:
`-Dforks` - How may time should the jvm fork. Running more than one fork will cause more iterations but should reset JVM optimization profiles. May reduce variance. Default Value: 1.
`-DwarmUpIterations` - How many warm up iterations we should run before we stop measuring. Default value: 5
`-DmeasurementIterations` - How many measurement iterations we should have. Default value: 10
`-Dthreads` - How many threads will run the test (concurrency). Default value: # of cores.

Example of how to run from CLI:
`mvn -Dtest=BenchmarkRunner#launchDbBenchmarks -Dforks=2 -DmeasurementIterations=6 -DwarmUpIterations=2 -Dthreads=3 test`

Fixes #1651 

## Type of change

- Test Enhancement (a non-breaking change which adds functionality)

# How Has This Been Tested?

Locally run the example line. The following was printed:
```
# Warmup: 2 iterations, 10 s each
# Measurement: 6 iterations, 10 s each
# Timeout: 10 min per iteration
# Threads: 3 threads, will synchronize iterations
# Benchmark mode: Average time, time/op
# Benchmark: com.iota.iri.benchmarks.dbbenchmark.RocksDbBenchmark.deleteBatch
# Parameters: (numTxsToTest = 10)
```

# Checklist:

### DIFF
--- a/src/test/java/com/iota/iri/benchmarks/BenchmarkRunner.java
+++ b/src/test/java/com/iota/iri/benchmarks/BenchmarkRunner.java
@@ -21,7 +21,8 @@ public class BenchmarkRunner {
                 .mode(Mode.AverageTime)
                 .timeUnit(TimeUnit.MILLISECONDS)
                 .warmupIterations(getWarmUpIterations(5))
-                .forks(getForks())
+                .forks(getForks(1))
+                .threads(getThreads())
                 .measurementIterations(getMeasurementIterations(10))
                 .shouldFailOnError(true)
                 .shouldDoGC(false)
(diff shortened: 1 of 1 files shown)


GalRogozinski:
True,
Devops said that they want to accelerate the test by utilizing more cpus

I am still waiting for their feedback

PR #1667, opened 2019-11-18
Fix: Updated db config loading to account for sections

### LINKED ISSUE #1624: Allow users to manually tweak RocksDb Options
### Description
Inside `mainnetdb` folder there is an Options file. It has all rocksdb configurations. Currently every time one restarts IRI, its contents is being replaced by hard coded options.
We want to ignore the hardcoded options in case the Options file is present.

### Motivation
One would like to tune RocksDb performance without changing java code.
Anyone in the community could fine settings that are good for him.

### Requirements
1. If options file is missing use hardcoded options
2. Else use options in file

kwek20:
# Description
Copy/pasting the OPTIONS file in rocksdb in place of the config file we use should work.
However previously, sections were not supported, resulting in `null` and default options beeing loaded.

This adds the ability to load from both a partial or a full OPTIONS file

Fixes #1624

## Type of change

# How Has This Been Tested?
Loaded both a OPTIONS file from another location, and a small config file

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/storage/rocksDB/RocksDBPersistenceProvider.java
+++ b/src/main/java/com/iota/iri/storage/rocksDB/RocksDBPersistenceProvider.java
@@ -569,7 +569,7 @@ public class RocksDBPersistenceProvider implements PersistenceProvider {
                 try (InputStream stream = new FileInputStream(config)){
                     // Map will contain DBOptions, TableOptions/BlockBasedTable, CFOptions
                     // Currently we only use DBOptions
-                    Map<String, Properties> map = parseINI(stream);
+                    Map<String, Properties> map = IotaIOUtils.parseINI(stream);
                     if (map.containsKey("DBOptions")) {
                         options = DBOptions.getDBOptionsFromProps(map.get("DBOptions"));
                     } else if (map.containsKey("default")) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/iota/iri/storage/rocksDB/RocksDBPersistenceProvider.java
```diff
@@ -598,4 +607,31 @@ private DBOptions createOptions(String logPath, String configFile) throws IOExce
...
+    // https://stackoverflow.com/a/41084504/4512850 -ish
+    private static Map<String, Properties> parseINI(InputStream stream) throws IOException {
+        Map<String, Properties> result = new HashMap<>();
+        
+        @SuppressWarnings("serial")
+        Properties p = new Properties() {
+
+            private Properties section;
+
+            public synchronized Object put(Object key, Object value) {
```
GalRogozinski: add `@Override` annotation please

src/main/java/com/iota/iri/storage/rocksDB/RocksDBPersistenceProvider.java
```diff
@@ -598,4 +607,31 @@ private DBOptions createOptions(String logPath, String configFile) throws IOExce
 
         return options;
     }
+    
+    // https://stackoverflow.com/a/41084504/4512850 -ish
+    private static Map<String, Properties> parseINI(InputStream stream) throws IOException {
```
GalRogozinski: I think you can make this public and move it to `IotaIOUtils`

Could be useful so why not?

PR #1685, opened 2019-12-16
fix broken getInclusionStates API call

### LINKED ISSUE #1350: GetInclusionStates api call is broken
### Bug description
Currently getInclusionStates api call is broken

### IRI version
1.6.0

### Steps To Reproduce
1.  Given:
```
milestone1 = RGAGQUXZ9VIVSKJHBUOYLDRZAVWHT9FWEBTDOILMZOEYXSGSZFEFLFYELJMQULYOGHHXGIF9ZHYDA9999
milestone2 = EAXDHOSARLDVKFLZMCJQMPI9MNCFZ9UMUXVRZKDKWXMWGASQNWWNYUWJBWRGCBVMO9WCSXCOBPIHZ9999
txBetweenM2_M3 = SHHCZGOPU9DB9HNMONJNOFRVQCRMFJNBP9OGQSRTMWSHVILTUSQKJBF9Y9QT9FGPSVXSVWDTBKNSZ9999
milestone3 = QYVQXYJBJEZCUPWLSHPIBYAXIPYULBKCDWGYH9XK9HGGEXLHCJAOEVV9WTJNHWXDIKNHQACQHESKA9999
pendingAfterM3 = A9JGAJIXRZW9KB9T9TODDXTWTFZLROWJDQZTVUNDENJQOEULME9YTFESCNAEMSLJJXCEZLEHF9SSZ9999
milestone4 = UOCCDROJVEIJLXBQFVO9ITKKDFTA9GPCQNTBIBEXBEGBRGFAGTZ9XSSDCKHCODJ9KAFDOLUGBPYGA9999
```
2. Then 
```
{"command": "getInclusionStates", "transactions": [milestone1, milestone2, txBetweenM2_M3, milestone3, milestone4], "tips": [pendingAfterM3]}
[ false, false, false, false, false ]
``` 
3.  
```
{"command": "getInclusionStates", "transactions": [milestone1, milestone2, txBetweenM2_M3, milestone3, milestone4], "tips": [txBetweenM2_M3]}
[ true, true, true, false, false ]
```
4.
```
{"command": "getInclusionStates", "transactions": [milestone1, milestone2, txBetweenM2_M3, milestone3, milestone4], "tips": [pendingAfterM3, txBetweenM2_M3]}
[ false, false, true, false, false ]
```

### Expected behaviour
Last command should return `[ true, true, true, false, false ]`

### Actual behaviour
`[ false, false, true, false, false ]`

acha-bill:
# Description

Updates `getInclusionStates` API call to only check the snapshot index of the given transactions. 
Tips are therefore irrelevant.

Fixes #1350 

## Type of change

# How Has This Been Tested?

- All existing unit tests pass
- API regression tests pass

# Checklist:


python-regression/tests/features/machine1/1_api_tests.feature
```diff
@@ -154,7 +154,7 @@ Feature: Test API calls on Machine 1
 
 		Then the response for "getInclusionStates" should return with:
 		|keys			|values			|type               |
-		|states			|False		        |bool               |
+		|states			|True		        |bool               |
```
GalRogozinski: really?
We will wait for @DyrellC to say whether this makes sense and wait for buildkite

src/main/java/com/iota/iri/service/API.java
```diff
@@ -752,122 +752,29 @@ private AbstractResponse getNodeAPIConfigurationStatement() {
...
      * @param transactions List of transactions you want to get the inclusion state for.
-     * @param tips List of tip transaction hashes (including milestones) you want to search for
      * @return {@link com.iota.iri.service.dto.GetInclusionStatesResponse}
      * @throws Exception When a transaction cannot be loaded from hash
      **/
     @Document(name="getInclusionStates")
-    private AbstractResponse getInclusionStatesStatement(
-            final List<String> transactions,
-            final List<String> tips) throws Exception {
+    private AbstractResponse getInclusionStatesStatement(final List<String> transactions ) throws Exception {
```
GalRogozinski: did you test what happens when someone passes tips still?
(it should work, but just checking)

python-regression/tests/features/machine1/1_api_tests.feature
acha-bill: ok

src/main/java/com/iota/iri/service/API.java
acha-bill: works. No difference

python-regression/tests/features/machine1/1_api_tests.feature
GalRogozinski: Btw, @DyrellC, this is what I was talking to you about needing to to properly document the DBs, currently it changed and I have no idea why by just looking at the code.

Besides that, @acha-bill, maybe it is worth while to have another scenario (or step) in this PR for `getInclusionStates`. We want to have both the `True` and `False` cases. 
We also want to make sure that we can accept a list of Hashes and not just one single hash.
You can wait for @DyrellC to explain to you what is going on with the DB

DyrellC: @galrogo this instance was actually the db that you had provided in the previous regression tests, but it does highlight this point yes. 
@acha-bill since the `getInclusionStates` call will no longer need the tip list, you can remove that from the input side of the test. You could also add the list example that @galrogo is asking for with a 2 for 1 by having another scenario that tests inclusion state on a list of randomly generated hashes and having that return a `boolList` of false. That would cover testing a list scenario as well as making sure we have false cases register false as well.

python-regression/util/static_vals.py
```diff
@@ -8,11 +8,14 @@
...
 
 TEST_HASH = "NMPXODIWSCTMWRTQ9AI9VROYNFACWCQDXDSJLNC9HKCECBCGQTBUBKVROXZLQWAZRQUGIJTLTMAMSH999"
+TEST_HASH_RES = [True]
 CONFIRMATION_REFERENCE_HASH = "ZSPRRUJRXHBSRFGCCPVHNXWJJKXRYZSAU9ZEGWFD9LPTWOJZARRLOEQYYWIKOPSXIBFD9ADNIVAHKG999"
 NULL_HASH = "999999999999999999999999999999999999999999999999999999999999999999999999999999999"
 
-
 TEST_TIP_LIST = ["SBKWTQWCFTF9DBZHJKQJKU9LXMZD9BMWJIJLZCCZYJFWIBGYYQBJOWWFWIHDEDTIHUB9PMOWZVCPKV999"]
+TEST_HASH_LIST = ["NMPXODIWSCTMWRTQ9AI9VROYNFACWCQDXDSJLNC9HKCECBCGQTBUBKVROXZLQWAZRQUGIJTLTMAMSH999",
+                  "INVALIDWSCTMWRTQ9AI9VROYNFACWCQDXDSJLNC9HKCECBCGQTBUBKVROXZLQWAZRQUGIJTLTMAMSH999"]
```
GalRogozinski: I suppose the second tx is a made up transaction.
Can we also add a real tx that is not included?

I think the scenarios should include:
1. A bundle tail that is approved
2. A bundle tail that is not approved
3. A non-tail that is approved
4. A non-tail that is not apporved
5. a made up tx

python-regression/tests/features/machine1/1_api_tests.feature
acha-bill: @DyrellC The python lib throws `expected 3 args` for `getInclusionStates`.
So we'll leave the `tips` param there for now.

python-regression/tests/features/machine1/1_api_tests.feature
```diff
@@ -154,7 +154,18 @@ Feature: Test API calls on Machine 1
 
 		Then the response for "getInclusionStates" should return with:
 		|keys			|values			|type               |
-		|states			|True		        |bool               |
+		| states | TEST_HASH_RES | staticValue |
```
GalRogozinski: **Only in case it is not too much work (no more than 5-10 minutes)**
I think that maybe for the responses we can put the actual list in the table instead of the static val.
I just don't like it that we mask what's going on too much.

With long hashes there is less of a choice because they are long and ugly...

If this requires work than just comment on the reply and ignore

python-regression/tests/features/machine1/1_api_tests.feature
```diff
@@ -154,7 +154,18 @@ Feature: Test API calls on Machine 1
...
+	#Values can be found in util/static_vals.py
+    Scenario: GetInclusionStates is called with transaction list
+	Given "getInclusionStates" is called on "nodeA-m1" with:
+	    | keys         | values         | type        |
+	    | transactions | TEST_HASH_LIST | staticValue |
+	    | tips         | TEST_TIP_LIST  | staticValue |
+
+	Then the response for "getInclusionStates" should return with:
+	    | keys   | values        | type        |
+	    | states | TEST_RESPONSE | staticValue |
```
GalRogozinski: ditto

python-regression/tests/features/machine1/1_api_tests.feature
GalRogozinski: Hmm I am adding communicating with clients to checklist

python-regression/tests/features/machine1/1_api_tests.feature
DyrellC: You can just use a boolList type here instead of loading a staticValue. Bool list will make an array of bools to compare with. So all you need to do is change the line to: 
`|states         | True            | boolList          |`
see line 113 in `python-regression/utils/test_logic/value_fetch_logic.py`

python-regression/tests/features/machine1/1_api_tests.feature
DyrellC: see above comment, same thing

python-regression/util/static_vals.py
acha-bill: Dropping scenario 2 and 4 as agreed with @galrogo .

python-regression/tests/features/machine1/1_api_tests.feature
acha-bill: @DyrellC add new type `boolListMixed` cause `boolList` could only convert 1 type.

python-regression/util/test_logic/value_fetch_logic.py
```diff
@@ -128,6 +128,16 @@ def fetch_bool_list(value):
...
 
 
+def fetch_bool_list_mixed(value):
+    """
+    Returns a list filled with bool conversions of the input string separated by space".
+    :param value: The input value
+    :return: The list of bool values
+    """
+
+    list = value.split()
```
GalRogozinski: Codacy Issue found: [Redefining built-in 'list'](https://app.codacy.com/manual/GalRogozinski/iri/pullRequest?prid=4682439)

python-regression/tests/features/machine1/1_api_tests.feature
DyrellC: Lovely :ok_hand:

PR #1694, opened 2019-12-19
Remove magic number from handshake

GalRogozinski:
# Description

While looking over networking code I found a magic number that confused me

## Type of change

- Readability fix 

# How Has This Been Tested?

If all unit tests and regression test pass it should suffice

# Checklist:


GalRogozinski:
Quick explanation:
`(maxLength - (maxLength - 60) + Protocol.SUPPORTED_PROTOCOL_VERSIONS.length` =  `60 + Protocol.SUPPORTED_PROTOCOL_VERSIONS.length`

It was unclear where 60 comes from, so it was replaced constants

src/main/java/com/iota/iri/network/protocol/Handshake.java
```diff
@@ -36,15 +43,14 @@
...
-        ByteBuffer buf = ByteBuffer.allocate(ProtocolMessage.HEADER.getMaxLength() + payloadLengthBytes);
-        Protocol.addProtocolHeader(buf, ProtocolMessage.HANDSHAKE, payloadLengthBytes);
+        ByteBuffer buf = ByteBuffer.allocate(ProtocolMessage.HEADER.getMaxLength() + PAYLOAD_LENGTH_BYTES);
+        Protocol.addProtocolHeader(buf, ProtocolMessage.HANDSHAKE, PAYLOAD_LENGTH_BYTES);
         buf.putChar(ownSourcePort);
         buf.putLong(System.currentTimeMillis());
         buf.put(ownByteEncodedCooAddress);
         buf.put(ownUsedMWM);
         buf.put(Protocol.SUPPORTED_PROTOCOL_VERSIONS);
+        //update PAYLOAD_LENGTH_BYTES if changed
```
acha-bill: Is this comment necessary?

GalRogozinski: So the programmer won't forget to update this value if in the future another `buf.put` is added
you think it is too obvious?

acha-bill: It's not "too" obvious. It's fine.

PR #1697, opened 2019-12-26
Feature: Cache Transaction Parents

### LINKED ISSUE #1689: Cache transactions' parents
### Description
Cache the Branch Approvee and the Trunk approvees (the transaction parents)

### Motivation
We want to cache data that helps us easily retrieve the parents for either validation or walks

### Requirements
We need to create the appropriate ViewModels before persisting to cache.
There should be an eviction policy similar to the TVM’s eviction policy. Evicted items should be weakly referenced.
We are handling immutable data here, so there is no concept of dirty
Before attempting to read data from the DB it must be first read from cache.
Flush to db upon shutdown

### Implementation suggestions
Instead of performing getSaveBatch and then tangle.saveBatch go via the ViewModels. This way we can relevant caches.
The problem with this solution is that it will disable us from using the writeBatch feature of RocksDb…
As far as I can see, we will only be able to solve this problem after we refactor the DB.

### Open Questions (optional)
Should we take this opportunity to cache anything else?

Blocked By: #1673

acha-bill:
# Description

_Please include a summary of the change. Include the motivation for the change._

Fixes #1689 

## Type of change

# How Has This Been Tested?

- Existing unit tests pass

# Checklist:


src/main/java/com/iota/iri/controllers/ApproveeViewModel.java
```diff
@@ -26,14 +28,20 @@ public ApproveeViewModel(Hash hash) {
         this.hash = hash;
     }
 
+    public ApproveeViewModel(ApproveeViewModel approveeViewModel) {
```
GalRogozinski: Codacy Issue found: [Public method and constructor comments are required](https://app.codacy.com/app/GalRogozinski/iri/pullRequest?prid=4733428)

src/main/java/com/iota/iri/controllers/TransactionViewModel.java
```diff
@@ -450,6 +466,17 @@ public boolean store(Tangle tangle, Snapshot initialSnapshot) throws Exception {
...
+        branchViewModel.addHash(hash);
+        approveeViewModelCache.put(getBranchTransactionHash(), branchViewModel);
+
+        ApproveeViewModel trunkViewModel = ApproveeViewModel.load(tangle, getTrunkTransactionHash());
+        trunkViewModel.addHash(hash);
+        approveeViewModelCache.put(getTrunkTransactionHash(), trunkViewModel);
+
+        ApproveeViewModel.load(tangle, hash);
+
         cachePut(tangle, this, hash);
```
GalRogozinski: maybe wrap this in a private method?

acha-bill: sure

PR #1699, opened 2019-12-30
Hotfix: Gyros - take transaction reattachments into account

GalRogozinski:
# Description

There is an edge case where IRI didn't account for a transaction that was shared between two distinct bundles. Once it marked it as "counted" in one bundle, it was ignored for the next bundle. This lead to a corrupt ledger state.

## Type of change

# How Has This Been Tested?

I ran it on a node that synced from hornet (1) and an isolated node that had a corrupt db (2).
1. It managed to sync from hornet :heavy_check_mark: 
2. It managed to restore the DB via the "reset corrupted milestone" mechanism

# Checklist:


GalRogozinski:
For clarity, when we have 2 bundle tails `R-0` and `0`, they both define distinct transfers. Transaction `3` is common to both of them and thus must be accounted twice.

image

GalRogozinski:
The corruption happened when milestone #1293082 was issued

PR #1702, opened 2020-01-04
Fix: Ensure proper creation of solid entrypoints

GalRogozinski:
# Description
Enforces the following definition for a solid entry point:

> A transaction is considered a solid entry point if it is a bundle tail that can be traversed down from a non-orphaned transaction that was approved by a milestone that is above the last local snapshot. Or if it is a bundle tail of a non-orphaned transaction that was approved by a milestone that is above the last local snapshot. 

Before non-tails could have been solid entry point, this caused problems in the case they were common to different bundles.

## Type of change

# How Has This Been Tested?
- My node was able to sync with the LS created.
- Using https://github.com/iotaledger/Snapshot.ixi I compared the ledger state of a node that is synced from last global snapshot to my synced node.

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/service/snapshot/impl/SnapshotServiceImpl.java
+++ b/src/main/java/com/iota/iri/service/snapshot/impl/SnapshotServiceImpl.java
@@ -510,7 +510,7 @@ public class SnapshotServiceImpl implements SnapshotService {
 
     /**
      * <p>
-     * This method determines if a transaction is orphaned.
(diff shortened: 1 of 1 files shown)


GalRogozinski:
We may want to release this quick as a hot fix.
My manual tests should suffice. A unit test can be created later...

Also this is not the most efficient implementation regarding db queries, but this can also be tended to later.

GalRogozinski:
After enough testing this will be merged w/o review in order to advance the release of the hotfix...
Can be later changed with subsequent PRs

PR #1707, opened 2020-01-14
Regression tests to python3

karimodm:
# Description

Fix regression tests that are now running on Python3.

# How Has This Been Tested?

By running our BuildKite pipelines.

https://buildkite.com/iota-foundation/iri-regression-tests/builds/1323


GalRogozinski:
@DyrellC 
I have merged so regression tests will start to pass, but we still need to update the README so that people will know to work with python 3 now

karimodm:
This has been tested against Python `3.8.1`, which, at present, comes within `python:3-alpine`

PR #1711, opened 2020-01-17
Added checks to make cache more robust

acha-bill:
# Checks to make caching more robust

## Type of change

# How Has This Been Tested?

- Existing unit tests pass

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/cache/impl/CacheImpl.java
+++ b/src/main/java/com/iota/iri/cache/impl/CacheImpl.java
@@ -76,9 +76,6 @@ public class CacheImpl<K, V> implements Cache<K, V> {
 
     @Override
     public void put(K key, V value) {
-        if (getSize() == cacheConfiguration.getMaxSize()) {
-            evict();
-        }
         // new entry
         if (strongStore.put(key, value) == null) {
             evictionQueue.offer(key);
--- a/src/main/java/com/iota/iri/cache/impl/CacheImpl.java
+++ b/src/main/java/com/iota/iri/cache/impl/CacheImpl.java
@@ -49,11 +49,12 @@ public class CacheImpl<K, V> implements Cache<K, V> {
             return null;
         }
         V value = strongStore.get(key);
-        if (value == null && weakStore.containsKey(key)) {
-            put(key, weakStore.get(key));
-            value = strongStore.get(key);
-            weakStore.remove(key);
-        }
+
+        // if (value == null && weakStore.containsKey(key)) {
+        // put(key, weakStore.get(key));
+        // value = strongStore.get(key);
+        // weakStore.remove(key);
+        // }
         if (value != null) {
             cacheHit();
         } else {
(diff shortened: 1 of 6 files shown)


src/main/java/com/iota/iri/controllers/TransactionViewModel.java
```diff
@@ -995,7 +995,8 @@ public static void cacheEvict(Tangle tangle) throws Exception {
                 TransactionViewModel tvm = cache.get(hash);
                 if (tvm != null) {
                     if (!tvm.getIsCacheEntryFresh()) {
-                        tvm.updateDB(tangle, tvm.getTransaction(), tvm.getHash(), "");
+                        List<Pair<Indexable, Persistable>> batch = tvm.getSaveBatch();
+                        tangle.saveBatch(batch);
```
GalRogozinski: hmmm,
I didn't say to do this thing... now I am thinking if it makes sense :thinking: 

If it is the first time we are saving the tx, then it must be this way and you fixed a serious bug...

For the updates, not so much because the data will never change

since if I am not mistaken, you just fixed a bug so let's leave it at that

GalRogozinski:
Since we updated to python3, can you please merge the `cache` branch and your PR branch with dev so that tests will pass?

src/main/java/com/iota/iri/controllers/TransactionViewModel.java
GalRogozinski: I think you got confused...
when I told you to use `saveBatch` it was in order to save a batch of transaction view models at once

acha-bill: Ah. How will `saveBatch(tvmList)` be used?

GalRogozinski: The `saveBatch` command can do a batch write of any data.
In your other PR #1713, when you evict a certain number of txs at once, you can do it via a save batch.

Then you are doing less DB queries,  less use of JNI, and possibly leveraging RocksDB optimizations

src/main/java/com/iota/iri/cache/impl/CacheImpl.java
```diff
@@ -49,11 +49,12 @@ public V get(K key) {
...
-            put(key, weakStore.get(key));
-            value = strongStore.get(key);
-            weakStore.remove(key);
-        }
+
+        // if (value == null && weakStore.containsKey(key)) {
+        // put(key, weakStore.get(key));
+        // value = strongStore.get(key);
+        // weakStore.remove(key);
+        // }
```
GalRogozinski: Something here is missing, either we want to make values that we get more sticky...
Or we don't...

Anyhows, we have to check weakStore on cache get no?

src/main/java/com/iota/iri/service/transactionpruning/jobs/MilestonePrunerJob.java
```diff
@@ -234,6 +235,7 @@ private void cleanupMilestoneTransactions() throws TransactionPruningException {
                             throw new RuntimeException(e);
                         }
                     }
+                    TransactionViewModel.cacheDelete(getTangle(), (Hash) element.low);
```
GalRogozinski: This shows how hacky and bad the design is

It is of course not your fault... and it show how important it is to refactor the persistence layer (or DAL layer if we want to start speaking java)...

Anyhows, 
There is also `UnconfirmedSubtanglePrunerJob` that you missed


From here we learn that we need more regression/unit tests to be added...
Let's fix this and we can do this after the release

acha-bill: oh, yeah.
We definitely need more regression tests to fish these out

PR #1726, opened 2020-01-26
check for null key and value before cache.put

acha-bill:
# Description

Added null check before cache put. Fixes 
```
01/26 20:54:46.064 [Milestone Solidifier] ERROR MilestoneSolidifier:465 - Error while solidifying milestone #1071451
java.lang.NullPointerException: null
	at java.util.concurrent.ConcurrentHashMap.putVal(ConcurrentHashMap.java:1011) ~[na:1.8.0_181]
	at java.util.concurrent.ConcurrentHashMap.put(ConcurrentHashMap.java:1006) ~[na:1.8.0_181]
	at com.iota.iri.cache.impl.CacheImpl.put(CacheImpl.java:107) ~[iri-PR-e1526d2-e8e23c4c.jar:PR-e1526d2-e8e23c4c]
	at com.iota.iri.cache.impl.CacheImpl.get(CacheImpl.java:78) ~[iri-PR-e1526d2-e8e23c4c.jar:PR-e1526d2-e8e23c4c]
	at com.iota.iri.controllers.TransactionViewModel.fromHash(TransactionViewModel.java:158) ~[iri-PR-e1526d2-e8e23c4c.jar:PR-e1526d2-e8e23c4c]
	at com.iota.iri.TransactionValidator.checkSolidity(TransactionValidator.java:264) ~[iri-PR-e1526d2-e8e23c4c.jar:PR-e1526d2-e8e23c4c]
	at com.iota.iri.service.milestone.impl.MilestoneSolidifierImpl.isSolid(MilestoneSolidifierImpl.java:344) [iri-PR-e1526d2-e8e23c4c.jar:PR-e1526d2-e8e23c4c]
```

## Type of change

# How Has This Been Tested?

- existing unit tests pass

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/cache/impl/CacheImpl.java
+++ b/src/main/java/com/iota/iri/cache/impl/CacheImpl.java
@@ -102,9 +102,9 @@ public class CacheImpl<K, V> implements Cache<K, V> {
 
     @Override
     public void put(K key, V value) {
-        if (key == null || value == null) {
-            return;
-        }
+        Objects.requireNonNull(key, "Cache key cannot be null");
+        Objects.requireNonNull(value, "Cache value cannot be null");
+
         if (getSize() >= cacheConfiguration.getMaxSize()) {
             release();
         }
--- a/src/main/java/com/iota/iri/cache/impl/CacheImpl.java
+++ b/src/main/java/com/iota/iri/cache/impl/CacheImpl.java
@@ -100,6 +100,9 @@ public class CacheImpl<K, V> implements Cache<K, V> {
 
     @Override
     public void put(K key, V value) {
+        if (key == null || value == null) {
+            return;
+        }
         if (getSize() >= cacheConfiguration.getMaxSize()) {
             release();
         }
--- a/src/main/java/com/iota/iri/cache/impl/CacheImpl.java
+++ b/src/main/java/com/iota/iri/cache/impl/CacheImpl.java
@@ -73,10 +73,12 @@ public class CacheImpl<K, V> implements Cache<K, V> {
         }
         V value = strongStore.get(key);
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/iota/iri/cache/impl/CacheImpl.java
```diff
@@ -100,6 +100,9 @@ public int getSize() {
 
     @Override
     public void put(K key, V value) {
+        if (key == null || value == null) {
+            return;
```
GalRogozinski: Something is strange to me... if we look at `get()` we see
``` java
if (value == null && weakStore.containsKey(key)) {
            value = weakStore.get(key);
            put(key, value);
            weakStore.remove(key);
        }
```
This `put` failed meaning that that the condition returned true...
I guess what happened is that `weakStore.containsKey(key)` returned true -> GC kicked in -> failure at put

I am afraid that you are just masking the problem here... paving the way to more latent bugs
I would say that fixing `get()` as so would be better:

``` java
if (value == null) {
            value = weakStore.get(key);
            if (value != null) {
               put(key, value);
               weakStore.remove(key);
        }
```

GalRogozinski: perhaps it is worth checking if at other places we forget that GC can kick in...

acha-bill: Makes sense.

GalRogozinski: > I am afraid that you are just masking the problem here
I would delete this change

Even though you are preventing, the error from happening you are just hiding the underlying cause
I would rather have an exception thrown in testing so we can do a proper fix when this happens

PR #1729, opened 2020-01-27
fix issue where updated tx was not saved to DB

acha-bill:
# Description

Fixes the following
- A tx could be cached and updated but not saved to DB during `cacheRelease` that was only called during tangle shutdown
- Move release code from the cache to VMs. Ensures that txs are saved to DB before releasing.
- Ensure to release all `releaseCount` txs from the cache
- Release all txs from cache during tangle shutdown.
- `dirty` tx kept saving to DB (increasing size) because it was still dirty even after being brought back to live from weakstore after release.

## Type of change

# How Has This Been Tested?

- Existing unit tests pass

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/cache/impl/CacheImpl.java
+++ b/src/main/java/com/iota/iri/cache/impl/CacheImpl.java
@@ -105,9 +105,6 @@ public class CacheImpl<K, V> implements Cache<K, V> {
         Objects.requireNonNull(key, "Cache key cannot be null");
         Objects.requireNonNull(value, "Cache value cannot be null");
 
-        if (getSize() >= cacheConfiguration.getMaxSize()) {
-            release();
-        }
         // new entry
         if (strongStore.put(key, value) == null) {
             releaseQueue.offer(key);
--- a/src/test/java/com/iota/iri/controllers/TransactionViewModelTest.java
+++ b/src/test/java/com/iota/iri/controllers/TransactionViewModelTest.java
@@ -441,7 +441,7 @@ public class TransactionViewModelTest {
 
         Hash hash = tvms[0].getHash();
         TransactionViewModel tvm = new TransactionViewModel((Transaction) tangle.load(Transaction.class, hash), hash);
-        Assert.assertTrue(tvm.isMilestone());
+        Assert.assertTrue("TVM should be a milestone", tvm.isMilestone());
     }
 
     @Test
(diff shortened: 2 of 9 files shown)


acha-bill:
ok. This idea does not work. Closing for now.
Maybe I was tired or something.

src/main/java/com/iota/iri/cache/impl/CacheImpl.java
```diff
@@ -105,9 +105,6 @@ public void put(K key, V value) {
         Objects.requireNonNull(key, "Cache key cannot be null");
         Objects.requireNonNull(value, "Cache value cannot be null");
 
-        if (getSize() >= cacheConfiguration.getMaxSize()) {
-            release();
```
GalRogozinski: This is certainly a bug

GalRogozinski:
I think what you did makes sense
How did you test exactly?

acha-bill:
This actually works.

src/test/java/com/iota/iri/controllers/TransactionViewModelTest.java
```diff
@@ -411,4 +419,66 @@ public void firstShouldFindTx() throws Exception {
...
+            trytes = TransactionTestUtils.nextWord(trytes);
+            tvms[i] = TransactionTestUtils.createTransactionWithTrytes(trytes);
+            tvms[i].store(tangle, snapshot);
+        }
+        tvms[0].isMilestone(tangle, snapshot, true);
+        tvms[1].isMilestone(tangle, snapshot, true);
+
+        Hash hash = tvms[0].getHash();
+        TransactionViewModel tvm = new TransactionViewModel((Transaction) tangle.load(Transaction.class, hash), hash);
+        Assert.assertTrue(tvm.isMilestone());
```
GalRogozinski: message is missing

src/test/java/com/iota/iri/controllers/TransactionViewModelTest.java
```diff
@@ -411,4 +419,66 @@ public void firstShouldFindTx() throws Exception {
...
+        }
+
+        // tangle shutdown
+        TransactionViewModel.cacheReleaseAll(tangle);
+
+        for (int i : tvmIndicesToUpdate) {
+            Hash hash = tvms[i].getHash();
+            TransactionViewModel tvm = new TransactionViewModel((Transaction) tangle.load(Transaction.class, hash),
+                    hash);
+            Assert.assertTrue(tvm.isMilestone());
```
GalRogozinski: message is missing

PR #1737, opened 2020-01-30
Fix reference transaction logic

DyrellC:
# Description
A fix for transaction issuance referencing another transaction in the regression tests. The machine 4 tests are reliant on this and have been exhibiting failures using the pool logic. This replaces the pool logic with simple synchronous api calls to ensure the test follows through properly. 

## Type of change

# How Has This Been Tested?
- Ran tests locally for machine 4 tests

# Checklist:

### DIFF
--- a/python-regression/tests/features/steps/response_handling_steps.py
+++ b/python-regression/tests/features/steps/response_handling_steps.py
@@ -24,6 +24,8 @@ def compare_thread_return(step, api_call):
     # Exclude duration from response list
     if 'duration' in response_list:
         del response_list['duration']
+    if 'info' in response_list:
+        del response_list['info']
     response_keys = response_list.keys()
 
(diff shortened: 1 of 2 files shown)


python-regression/tests/features/steps/response_handling_steps.py
```diff
@@ -24,6 +24,8 @@ def compare_thread_return(step, api_call):
     # Exclude duration from response list
     if 'duration' in response_list:
         del response_list['duration']
+    if 'info' in response_list:
```
kwek20: Whats this for dyrell?

DyrellC: This is for checkConsistency negative responses. The length of arguments wasn't matching because info was not included in the expected responses list

PR #1740, opened 2020-01-31
Cache: Check for empty entries before adding them to cache

DyrellC:
# Description
During creation of TVM's the `TransactionViewModel.fromHash()` function will try to fetch the object from the cache first. If the object doesn't exist in the cache it will then try to fetch it from the db. If it doesn't exist in the db a new transaction is formed for the TVM, and then subsequently placed in the cache. This can happen in several places but there is a compounding effect during syncing as a transaction may be stored, but the trunk and branch of that transaction are not. Each time a `fromHash()` is called on those objects they are found in the cache as empty objects. So long as the empty object is found in the cache, new transactions will not be allowed to be stored until the empty object has been released. This cascades over into the `ApproveeViewModel` as well by filling the cache with empty AVM's.  

This patch makes sure to only store relevant TVM's and AVM's which helps to correct the synchronisation problem. 

Fixes: synchronisation issue in caching branch 

## Type of change

# How Has This Been Tested?
- Local testing for synchronisation between nodes sped up significantly after restoring the call

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/controllers/ApproveeViewModel.java
+++ b/src/main/java/com/iota/iri/controllers/ApproveeViewModel.java
@@ -74,7 +74,7 @@ public class ApproveeViewModel implements HashesViewModel {
         }
 
         approveeViewModel = new ApproveeViewModel((Approvee) tangle.load(Approvee.class, hash), hash);
-        if (cache != null) {
+        if (cache != null && approveeViewModel.getHashes().size() > 0) {
             cachePut(tangle, approveeViewModel, hash);
         }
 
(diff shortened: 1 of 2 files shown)


acha-bill:
This does not seem to help the sync speed on my end.

Secondly, the updated TX is either already in cache or in DB. 
Except for somewhere along the sync process, we rely on the update to already be in DB

acha-bill:
Nice 👍

src/main/java/com/iota/iri/controllers/ApproveeViewModel.java
```diff
@@ -74,7 +74,7 @@ public static ApproveeViewModel load(Tangle tangle, Indexable hash) throws Excep
         }
 
         approveeViewModel = new ApproveeViewModel((Approvee) tangle.load(Approvee.class, hash), hash);
-        if (cache != null) {
+        if (cache != null && approveeViewModel.getHashes().size() > 0) {
```
GalRogozinski: I usually like using the `isEmpty()`
Usually we also need to check that the collection is not null (but in this case it won't ever be null)

Issue #1734, opened 2020-01-28
Add more unit tests for caching


acha-bill:
### Description

We need to cover more unit tests scenarios for caching.

### Motivation
We currently seeing some new bugs.
We could see these bugs earlier if our unit tests maybe covered a broader range of scenarios.

### Requirements
The following tests can be added

1. Check that cache size never grows past max configured size
2. A tx that is updated in the cache should be evicted to DB before being released from cache.
3. After prunning the DB with cache on, the DB shouldn't grow slowy i.e txs that have been deleted from DB shouldn't be written back from cache.
4. etc.

### Open Questions (optional)
What other tests can we add?

GalRogozinski:
@acha-bill, if you are done with this move it to done please

PR #1741, opened 2020-02-04
Added TVM cache delete tests

acha-bill:
# Description

Added unit tests to cover tx delete scenarios with cache.

Part of #1734 

## Type of change

# How Has This Been Tested?

- New and existing unit tests pass

# Checklist:

### DIFF
--- a/src/test/java/com/iota/iri/controllers/TransactionViewModelTest.java
+++ b/src/test/java/com/iota/iri/controllers/TransactionViewModelTest.java
@@ -8,6 +8,7 @@ import com.iota.iri.cache.impl.CacheManagerImpl;
 import com.iota.iri.conf.MainnetConfig;
 import com.iota.iri.crypto.SpongeFactory;
 import com.iota.iri.model.Hash;
+import com.iota.iri.model.HashFactory;
 import com.iota.iri.model.TransactionHash;
 import com.iota.iri.model.persistables.Transaction;
 import com.iota.iri.service.snapshot.Snapshot;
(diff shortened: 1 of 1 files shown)


src/test/java/com/iota/iri/controllers/TransactionViewModelTest.java
```diff
@@ -481,4 +482,36 @@ public void updatedTxShouldBeSavedToDbBeforeTangleShutdown() throws Exception {
...
+
+        Transaction tx = new Transaction();
+        Hash hash = HashFactory.TRANSACTION
+                .create("TEST9TRANSACTION9TEST9TRANSACTION9TEST9TRANSACTION9TEST9TRANSACTION9TEST999999999");
+        tx.address = hash;
+        TransactionViewModel tvm = new TransactionViewModel(tx, hash);
+
+        tvm.store(tangle, snapshot);
+        tvm.delete(tangle);
+        Assert.assertNull("TVM should be null", cache.get(hash));
```
kwek20: i think that if you move this line to the other test, and delete this test its also sufficient
no need to do an almost equal test twice.

Unless you see a case where `deletedTxFromDbShouldDeleteFromCache` fails but `deletedTxShouldNotBeBroughtBackToCache` succeeds?

GalRogozinski: I am still waiting for @acha-bill's response.
But imo it makes sense that there are two tests, because the second test came to ensure that a bug we had won't come back...

And there is a case where one test fails and the other won't. If the second fail test and the first one won't we will know immediately that the problem was `TransactionViewModel.fromHash` is to blame

However, Brord has a point that since we unfortunately are not mocking `tangle` atm this test is expensive. The situation where we have poorly designed unit tests is less of a priority because they still work fine...

So I'll let @achabill make the final call

acha-bill: I get what you mean @kwek20 .
But 2 tests makes it explicit. Like Gal said, there was an issue we had were deleted tx coming back to cache when `fromHash` was called.
So the 2 tests look fine for me.

PR #1744, opened 2020-02-05
Don't reply if there is nothing to request even if you are not synced

GalRogozinski:
# Description
We were seeing an issue of requesting too many needless transactions that clogged the pipeline queue.
So now a node won't reply even if it is not in sync

## Type of change

# How Has This Been Tested?

TBD

# Checklist:


GalRogozinski:
This will make you not request transactions from a synced node...
It is very bad so it will be closed in favor of #1745

PR #1758, opened 2020-02-18
Optimize cache release

acha-bill:
# Description

Sometimes the node will crash during travis tests. Profiling CPU and Memory usage highlighted increased cpu/memory usage.

These changes aim to optimize the cache release routine which we found was the major culprit.

## Type of change

# How Has This Been Tested?

Running travis test locally and profiling CPU/Memory. Results collected with `htop`

```
no cache (1.8.4)    = %cpu / %memory
                    = 0.5  / 1.0 (approx. 300mb avg)

cache_size - release_count = %cpu / %mem
1,000  - 10         = 12  / 1.1
5,000  - 100        = 9   / 1.2
5,000  - 500        = 7   / 1.2
5,000  - 1,000      = 7   / 1.3
10,000 - 1,000      = 0.5 / 1.2
10,000 - 2,000      = 0.5 / 1.2
1,000  - 50         = 12  / 1.1
1,000  - 80         = 12  / 1.1

//This PR
1,000  - 10         = 0.5  / 1.1 (approx. 330mb avg)

Device:
Macbook 2018
RAM: 16gb
CPU: i7, 6 cores
```
# Checklist:


src/main/java/com/iota/iri/controllers/ApproveeViewModel.java
```diff
@@ -179,16 +179,20 @@ private static void cacheDelete(Tangle tangle, Indexable hash) {
      */
     private static void cacheRelease(Tangle tangle) throws Exception {
         Cache<Indexable, ApproveeViewModel> cache = tangle.getCache(ApproveeViewModel.class);
-        Queue<Indexable> releaseQueueCopy = cache.getReleaseQueueCopy();
+        Queue<Indexable> releaseQueue = cache.getReleaseQueue();
 
-        for (int i = 0; i < cache.getConfiguration().getReleaseCount(); i++) {
-            Indexable hash = releaseQueueCopy.poll();
+        int releaseCount = 0;
```
GalRogozinski: I think we can even do without this `releaseCount` because we know what it will be thanks to `ReleaseCount`

acha-bill: `releaseCount` is the number of items released. We need it to make sure we only release the specified `config.getReleaseCount` items.

acha-bill: If you are suggesting instead that we should remove the whole release count feature and have the cache at `1k/1`, then once the cache is full, it'll do 1 `release` before every `put`.
Is that not more expensive?

GalRogozinski: It means that for every put we do a release, yes
I am not sure why it would be more expensive... It would smooth out cpu effort over time
If you're sure that it is more expensive, please explain here

If not you can try it out :-)

acha-bill:
Same benchmark results after removing release count.
1,000  - 10         = 0.5  / 1.1 (approx. 330mb avg)

src/main/java/com/iota/iri/controllers/TransactionViewModel.java
```diff
@@ -976,66 +975,45 @@ public String toString() {
...
+    private static void cacheRelease(Tangle tangle, Cache<Indexable, TransactionViewModel> cache) throws Exception {
+        Indexable hash = cache.getNextReleaseKey();
+        releaseNext(tangle, cache, hash);
+    }
+
+    private static void releaseNext(Tangle tangle, Cache<Indexable, TransactionViewModel> cache, Indexable hash) throws Exception {
+        TransactionViewModel tvm = cache.get(hash);
+        if (tvm != null && tvm.shouldPersist()) {
+            tvm.setShouldPersist(false);
+            cache.put(hash, tvm);
```
GalRogozinski: why do you need this line?

src/main/java/com/iota/iri/controllers/TransactionViewModel.java
```diff
@@ -976,66 +975,45 @@ public String toString() {
...
-                    cache.put(hash, tvm);
-                    batch.addAll(tvm.getSaveBatch());
-                }
-            }
+    private static void cacheRelease(Tangle tangle, Cache<Indexable, TransactionViewModel> cache) throws Exception {
+        Indexable hash = cache.getNextReleaseKey();
+        releaseNext(tangle, cache, hash);
+    }
+
+    private static void releaseNext(Tangle tangle, Cache<Indexable, TransactionViewModel> cache, Indexable hash) throws Exception {
```
GalRogozinski: Maybe a better name?
`persistAndReleaseNext` ? (I don't know if it is so good)

src/main/java/com/iota/iri/controllers/TransactionViewModel.java
```diff
@@ -976,66 +975,45 @@ public String toString() {
...
-                TransactionViewModel tvm = cache.get(hash);
-                hashesToRelease.add(hash);
-                if (tvm != null && tvm.shouldPersist()) {
-                    tvm.setShouldPersist(false);
-                    cache.put(hash, tvm);
-                    batch.addAll(tvm.getSaveBatch());
-                }
-            }
+    private static void cacheRelease(Tangle tangle, Cache<Indexable, TransactionViewModel> cache) throws Exception {
+        Indexable hash = cache.getNextReleaseKey();
```
GalRogozinski: I would put this inside the implementation of the method below

src/main/java/com/iota/iri/controllers/TransactionViewModel.java
```diff
@@ -976,66 +975,45 @@ public String toString() {
     private static void cachePut(Tangle tangle, TransactionViewModel transactionViewModel, Hash hash) throws Exception {
         Cache<Indexable, TransactionViewModel> cache = tangle.getCache(TransactionViewModel.class);
         if (cache.getSize() >= cache.getConfiguration().getMaxSize()) {
```
GalRogozinski: Maybe this should be changed into `while` on all implementations? Just to make this more robust

src/main/java/com/iota/iri/controllers/TransactionViewModel.java
DyrellC: I agree, it doesn't look  like `cacheRelease()` is a necessary method, nor does it look like it's necessary to get the next release key, as the `cache.releaseNext()` call poll's the next releasable key anyways. So we're just fetching the key an extra time each time for no reason. You could instead set `cache.releaseNext()` to return a bool and that would let you check if the next hash is null by returning false from within the cache if it is. No more need for the `getNextReleaseKey()` that way.

src/main/java/com/iota/iri/controllers/TransactionViewModel.java
```diff
@@ -976,66 +975,45 @@ public String toString() {
...
-                TransactionViewModel tvm = cache.get(hash);
-                hashesToRelease.add(hash);
-                if (tvm != null && tvm.shouldPersist()) {
-                    tvm.setShouldPersist(false);
-                    cache.put(hash, tvm);
-                    batch.addAll(tvm.getSaveBatch());
-                }
-            }
+        Indexable hash = cache.getNextReleaseKey();
+        while(hash != null) {
```
DyrellC: This could be replaced with a bool return as described below and save the need to make the extra getNextReleaseKey calls to the cache.

src/main/java/com/iota/iri/controllers/TransactionViewModel.java
acha-bill: so that in the case where this released item is brought back from `weakStore`, it does not write to DB again.

src/main/java/com/iota/iri/controllers/TransactionViewModel.java
acha-bill: It's better than `releaseNext` though. :)

src/main/java/com/iota/iri/controllers/TransactionViewModel.java
acha-bill: In order to persist and release, we need to get the next release key and check `shouldPersist` before.
I agree that we can use the `bool` return in the case of `cacheReleaseAll`

src/main/java/com/iota/iri/controllers/TransactionViewModel.java
acha-bill: I don't follow

GalRogozinski: All the notification from the migration tool made me miss this...

Imagine that due to some weird scenario we have `cache.getSize() == cache.getConfiguration().getMaxSize() + 100`

I don't see how this can happen but still, code changes

So if you put a while you will solve this :-)?

You can also argue back that this shouldn't happen, and we're just attempting to mask bugs (which is also a valid statement)

kwek20: I dont see how this can happen gal,Could you describe this scenario? And if it happens, we should have a unit test for it :)

src/main/java/com/iota/iri/controllers/TransactionViewModel.java
```diff
@@ -988,10 +988,10 @@ private static void cachePut(Tangle tangle, TransactionViewModel transactionView
      */
     public static void cacheReleaseAll(Tangle tangle) throws Exception {
         Cache<Indexable, TransactionViewModel> cache = tangle.getCache(TransactionViewModel.class);
-        Indexable hash = cache.getNextReleaseKey();
-        while(hash != null) {
-            releaseNext(tangle, cache, hash);
-            hash = cache.getNextReleaseKey();
+        while(true) {
```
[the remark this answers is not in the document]
DyrellC: I agree this can be a one liner.

src/main/java/com/iota/iri/controllers/TransactionViewModel.java
GalRogozinski: currently it can't happen
the point is that code changes

What the code does now is not what it will do tomorrow

acha-bill: Is it not better to throw an exception in `cache.put` if you attempt to put when it's full?

GalRogozinski: this will indeed make the problem easier to spot!!!
So yes

acha-bill: But then we'll be double-checking size constraints for each cache put. A check in `TVM.cachePut` and check again in `cache.put`.
I think we should leave it as it is. :)

GalRogozinski: hmm... I don't think those checks are so expensive though

acha-bill: We’ve come back to this:

https://iotafoundation.slack.com/archives/GRVHJAA9F/p1580815146089200
> so I added another cache unit test: cacheSizeShouldNotGrowPastMaxSize and it’s failing.
Because if cache is full and then a get is called for a recently released item that’s still in weakstore, it puts it back in strongstore.
So there’s a +1 on max size there. :disappointed:
We cannot release in cache.put because we have to make sure its released to DB if necessary (from the VM)

so now I’m getting some cacheExceptions from unit tests due to the above
Since we are now releasing 1 at a time, it means cache is always full.
So if cache.get is called on an item that is in weakstore, it’ll try to add it to already full cache which will throw Exception.

What I think we should do, for now, is to use the `while` option @GalRogozinski  proposed.
Then open an issue for refactoring that can be looked at a later time

PR #1761, opened 2020-02-20
sort trytes before doing pow in attachToTangle API call

### LINKED ISSUE #1708: allow reversed order for attachToTangle
### Bug description
If you send the attachToTangle API call with the txs of a bundle not reversed (index 0 first element) IRI will do the PoW in the wrong order, resulting in a invalid consistency for the bundle index 0:<https://thetangle.org/transaction/MDGEQZZMQGECFLCZLNKRCZHHWA9AJWQILEM9KXXJZOVGNEJFYZTZGDUMBXOJCEIBXKRNDGJHCFID99999> children tx is index 1, but it should be the trunk tx

### IRI version
1.8.4-RELEASE

### Steps To Reproduce
1. Send attachToTangle request with multiple txs and tail txs as first element

### Expected behaviour
IRI should reverse the order before doing the PoW

### Actual behaviour
IRI does the PoW in the wrong order

acha-bill:
# Description

Sort transaction trytes in descending order by `currentIndex` so that the order of trytes in `attachToTangle` API call does not matter.

Fixes #1708 

## Type of change

# How Has This Been Tested?

Existing unit tests pass

# Checklist:


GalRogozinski:
@acha-bill, @kwek20 , and @DyrellC 
note the `getInclusionStates` fail here again

This is also a problem that didn't happen before so often... and now it happens again
I am not merging yet, I want to talk about it (again I know)

PR #1774, opened 2020-02-29
Only cache after exists check passes

GalRogozinski:
# Description

When a node sees a transaction it may have seen before it won't insert it into the cache to not override the existing transaction.

Fixes a problem where we may be switching instances of txs and thus not setting them solid, because we skip the `updateStatus` method

## Type of change

# How Has This Been Tested?

TBD

# Checklist:


GalRogozinski:
Current Testing doesn't show results for the devnet problem
But maybe the PR is still good?

PR #1775, opened 2020-03-01
Regression: Update pruning test 

### LINKED ISSUE #1337: Regression test: DB Pruning works correctly on a pruning enabled node
### Description
We want to have a regression tests that verifies that a node is pruning data below pruning delay and doesn't prune data within the delay. 

Note: this regression test is running into the issue of having to reach the set PRUNING_DELAY first. 

### Motivation
We need to make sure that the data configured by the node to present in the DB is always present, not more, not less. 

### Scenarios
1. I start up a new node from local snapshot files.
2. The node synchronizes and the DB holds the transaction data specified by the PRUNING_DELAY + DEPTH interval.

### How to test for this
1. Start up a permanode. 
2. Start a node with default configuration that is connected to the permanode. The node synchronizes and starts taking local snapshots at the SYNCED interval.
3. The node doesn't hold more transactions transactions than defined by PRUNING_DELAY + DEPTH, given the current PRUNING_INTERVAL. 
4. The node holds all transactions in its DB that are within PRUNING_DELAY + DEPTH. Query at different portions of the interval.

DyrellC:
# Description
Adds the requested before and after tests to confirm pruning occurred. 

Fixes #1337 

## Type of change

# How Has This Been Tested?
- Regression tests pass locally and with kubernetes

# Checklist:


DyrellC:
@galrogo conflicts resolved

PR #1777, opened 2020-03-03
Make Transaction Mutable Fields Atomic 

DyrellC:
# Description
OUTDATED: In the current cache implementation, the assumption of a read-only environment for `ApproveeViewModel`'s has led to the removal of db storage for Approvee objects. As a result, it is possible for both walking and transaction solidity propagation to be halted/hindered for transactions in the event that a given transaction's `ApproveeViewModel` has been swept from weak store by the Garbage Collector. 

CURRENT: Transactions being passed through the cache should ensure the thread safety of the corresponding mutable fields to ensure that the logic does not ignore a parsed or solid transaction due to a race condition. Also, within the TVM store, we should be checking if the object exists before performing any caching and store logic. This is handled with the inclusion of #1774 .

OUTDATED: Fixes Dev-Ops issue #99 

## Type of change

# How Has This Been Tested?
- Requires testing in a larger network like devnet to confirm that it fixes the issue
- Locally passes regression and unit tests

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/TransactionValidator.java
+++ b/src/main/java/com/iota/iri/TransactionValidator.java
@@ -278,6 +278,7 @@ public class TransactionValidator {
         }
         if (solid) {
             updateSolidTransactions(tangle, snapshotProvider.getInitialSnapshot(), analyzedHashes);
+            analyzedHashes.forEach(this::addSolidTransaction);
         }
         analyzedHashes.clear();
         return solid;
--- a/src/main/java/com/iota/iri/controllers/TransactionViewModel.java
+++ b/src/main/java/com/iota/iri/controllers/TransactionViewModel.java
@@ -212,7 +212,7 @@ public class TransactionViewModel {
 
         this.hash = hash;
         weightMagnitude = this.hash.trailingZeros();
-        transaction.type = FILLED_SLOT;
+        transaction.type.set(FILLED_SLOT);
     }
 
     /**
@@ -508,7 +508,7 @@ public class TransactionViewModel {
      * @return The current type of the transaction.
      */
     public final int getType() {
-        return transaction.type;
+        return transaction.type.get();
     }
 
     /**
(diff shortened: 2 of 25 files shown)


acha-bill:
Added Gals #1774 into the mix

GalRogozinski:
I only looked now
I actually don't understand this

Approvees are:
1. immutable
2. stored when we save a transaction (so we don't need to have approvee store do anything....)

src/main/java/com/iota/iri/TransactionValidator.java
```diff
@@ -278,6 +278,7 @@ public boolean checkSolidity(Hash hash, int maxProcessedTransactions) throws Exc
         }
         if (solid) {
             updateSolidTransactions(tangle, snapshotProvider.getInitialSnapshot(), analyzedHashes);
+            analyzedHashes.forEach(this::addSolidTransaction);
```
GalRogozinski: interesting... :nerd_face:

src/main/java/com/iota/iri/controllers/TransactionViewModel.java
```diff
@@ -98,7 +98,7 @@
      * @throws Exception Thrown if the database fails to save the batch of data.
      */
     public static void fillMetadata(Tangle tangle, TransactionViewModel transactionViewModel) throws Exception {
-        if (transactionViewModel.getType() == FILLED_SLOT && !transactionViewModel.transaction.parsed) {
+        if (transactionViewModel.getType() == FILLED_SLOT && !transactionViewModel.transaction.parsed.get()) {
```
GalRogozinski: type should also be atomic

src/main/java/com/iota/iri/network/pipeline/ReplyStage.java
```diff
@@ -146,6 +146,16 @@ public ProcessingContext process(ProcessingContext ctx) {
...
+        // we didn't have the requested transaction (random or explicit) from the neighbor but we will immediately reply
+        // with the latest known milestone and a needed transaction hash, to keep up the ping-pong
+        try {
+            final TransactionViewModel msTVM = TransactionViewModel.fromHash(tangle,
+                    latestMilestoneTracker.getLatestMilestoneHash());
+            neighborRouter.gossipTransactionTo(neighbor, msTVM, false);
+        } catch (Exception e) {
+            e.printStackTrace();
+        }
+
```
GalRogozinski: This should be reverted again I think?

acha-bill: yeah

PR #1786, opened 2020-03-10
added bundle validity rules

### LINKED ISSUE #1746: Change bundle validity rules for enabling more robust node architecture
### Description
(I've been instructed to write this issue (and not an RFC))

Enforce that bundles:
* must only approve tail transactions
* all transactions within the bundle approve via their branch the trunk transaction of the head transaction.

#### About rule 1
The ontology concept describes an abstraction which allows a node software to create different views from the origin Tangle, in order to enable graphs or data structures to simplify processing transactions, bundles or other data in a meaningful way.

For example, consider a bundle tangle which itself is represented by a graph made up of `Bundle` objects holding for example fields like this:
* tail hash
* diffs on addresses (meaning the address with the corresponding mutation caused by this bundle)
* refs (a set of hashes to other approved `Bundle`s)
* ...

Out of such objects, a graph can be built which is basically an abstraction providing immediate information which can be used for tip-selection or other functionalities. In such "bundle tangle", an invalid bundle for example wouldn't even occur, making tip-selection easier (and faster) as the graph represents a validated graph of bundles (meaning the graph only grows when a new valid bundle becomes solid). One can take this ontology concept even further and for example construct a walker tangle out of the bundle tangle, where the graph represents only the subset of vertices which are relevant for tip-selection (for example by removing anything below max depth).

In order to make such bundle tangle ontology work however, it is required that bundles only approve other tail transactions, as otherwise it is not possible to construct such graph: if a bundle could approve a non tail transaction, it wouldn't be clear what the actual tail of that bundle is (remember, one can reattach a tail directly onto the origin bundle infinite times).

#### About rule 2

Enforcing that a bundle can max approve two other transactions enables the node software to work with less logic or even removes potential vulnerabilities, as it eases the complexity of allowing a lot of cones to be approved by a single atomic transfer. Our client libraries already build bundles in such fashion per default, so it would make sense to let the node software also be only agnostic to such bundles.

### Requirements
In order to enforce this rules, the bundle validation logic has to be adjusted to deem bundles invalid where:
* the branch transaction hash of the non head transactions within a bundle, is not the same as the trunk transaction hash of the head transaction.
* the trunk and branch transactions approved by the bundles are non tails.

Note that the second validation logic enforces, that a bundle can only be validated when it becomes solid (which shouldn't really change anything as a non-solid bundle in itself is also useless).

### Open Questions (optional)
Rule 2 doesn't directly affect the ontology concept, however, it allows it to keep the same "a vertex approves two other vertices", as a `Bundle` object in such bundle tangle would only need to keep max two other references.

No.

acha-bill:
# Description

Adds 2 new rules for bundle validity. A bundle is valid if:

- It only approves tail transactions
- All transactions within the bundle approve via their branch the trunk transaction of the head transaction.

Fixes #1746 

## Type of change

# How Has This Been Tested?

- Existing unit tests pass that cover newly added rules
- Added new tests for the specific rules

# Checklist:


src/main/java/com/iota/iri/BundleValidator.java
```diff
@@ -379,6 +403,37 @@ public static boolean isInconsistent(Collection<TransactionViewModel> transactio
...
+     * A bundle is invalid if The branch transaction hash of the non head transactions within a bundle, is not the same
+     * as the trunk transaction hash of the head transaction.
+     *
+     * @param bundleTxs list of transactions that are in a bundle.
+     * @return Whether the bundle tx chain is valid.
+     */
+    public static Validity validateBundleTransactionsApproval(List<TransactionViewModel> bundleTxs){
+        Hash headTrunkTransactionHash = bundleTxs.get(bundleTxs.size() - 1).getTrunkTransactionHash();
+        for(int i = 0; i < bundleTxs.size() - 1; i++){
+            if(!bundleTxs.get(i).getBranchTransactionHash().equals(headTrunkTransactionHash)){
```
GalRogozinski: usually I don't like all those concatenated gets becuase they mask `nulls`
but I think here no nulls would appear anyhows

src/test/java/com/iota/iri/BundleValidatorTest.java
```diff
@@ -79,6 +79,30 @@ public void checkInconsistencyOfConsistentBundle() {
...
+        };
+        List<TransactionViewModel> transactions = persistAndMapTxs(trytes);
+        assertEquals("should be valid because bundle approves tails", BundleValidator.Validity.VALID,  BundleValidator.validateBundleTailApproval(tangle, transactions));
+    }
```
GalRogozinski: 2 requests:
1. Can you add some comments explaining the bundles you created?
2. Can you add tests that check for invalidity?

about (1), I like less how the tests in general in this class are written. When Luca refactored the bundle validator I didn't want to delay him and it slipped in. We will not rewrite them to make sure they are clear any time soon, but comments on new tests will be nice :-)

acha-bill:
Manual tests:

rule 1: It only approves tail transactions

```
trunk = tail, branch = tail : true
branch = trunk = non tail: invalid bundle
trunk = non tail, branch = non tail: invalid bundle
trunk = tail, branch = non tail:  invalid bundle
```

rule 2:   All transactions within the bundle approve via their branch the trunk transaction of the head transaction
Created a good bundle with 3 transactions, edited it and set it's `tx[1].branch = some old tx`. 
Then validity for this new bundle.

Db and source here: https://github.com/acha-bill/iri_bundle_validity

Issue #1615, opened 2019-09-12
When running getTips on an isolated node an empty array is returned


GalRogozinski:
# Bug description
If you run a node that is not receiving any transactions, the tip cache (`TipsViewModel`) is not populated. so `getTips` api call will return an empty array.

Not sure whether the solution is to persist tips, or maybe scan for them upon starting IRI.

### IRI version
1.8.1

### Steps To Reproduce
1. Run a node in isolation
2. Call `getTips`: 
```
curl http://localhost:14265 \
-X POST \
-H 'Content-Type: application/json' \
-H 'X-IOTA-API-Version: 1' \
-d '{"command": "getTips"}'
```

### Expected behaviour
List of tips should be retruned

### Actual behaviour
An empty list is returned:
` "hashes": [],`

jakubcech:
We'll try and figure out if anyone's actually using it for some scenario and if not we can just deprecate it.

GalRogozinski:
According to Discord we should remove it

PR #1787, opened 2020-03-11
Do not return unsolid tips for API call

DyrellC:
# Description
The current `GetTips` api call returns both solid and unsolid tips. This can lead to chaining of transactions off of unsolid tips, which will not be seen by the gtta walks due to the current solidification logic. Transactions are placed into the `TipsViewModel` before any attempts to solidify the transaction are conducted. As such, in the event of an out of order transaction, a tip that can be returned in the `TipsViewModel` may not be solid by the time it is requested by the API call. If that is the case, the next transaction attached on top of it will result in a chain of transactions built upon a transaction that has yet to be marked solid. All transactions attached on top of that will also fail to solidify unless the propagation thread in the `TransactionValidator` happens to see the root transaction of this chain. 

Fixes the current issue seen on devnet, and will likely assist in increasing the confirmation rate on mainnet as many users are using `getTips()` instead of `getTransactionsToApprove()` to fetch their trunks and branches right now.

## Type of change

# How Has This Been Tested?
- We have set up an internal network comprising of a chain of connected nodes to emulate the devnet environment. With the change present, chains of unsolid transactions do not form as they did before. 

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/controllers/TipsViewModel.java
+++ b/src/main/java/com/iota/iri/controllers/TipsViewModel.java
@@ -76,10 +76,6 @@ public class TipsViewModel {
         Set<Hash> hashes = new HashSet<>();
         synchronized (sync) {
             Iterator<Hash> hashIterator;
-            hashIterator = tips.iterator();
-            while (hashIterator.hasNext()) {
-                hashes.add(hashIterator.next());
-            }
 
             hashIterator = solidTips.iterator();
             while (hashIterator.hasNext()) {


GalRogozinski:
Due to discord messages and #1615 maybe we will change it to delete this method (or class entirely)

DyrellC:
In the meantime it's important to tell people that they should not use getTips as is then. There's still spam on devnet that follows the pattern of getTips calls which is dragging down confirmation of transaction branches

PR #1802, opened 2020-03-17
Check validity rules on tip-sel and check-consistency only

### LINKED ISSUE #1746: Change bundle validity rules for enabling more robust node architecture

GalRogozinski:
# Description

We want the rules added in #1786 to only be used when we are doing tip-selection and check consistency calls. This is so that IRI won't give an error when calculating the ledger state on old milestones that didn't adhere to the rule.

We have a boolean flag in the `validate` method that takes care of the 2 modes, that forces the extra checks.

Fixes #1746 

## Type of change

# How Has This Been Tested?
Unit test have been added

- One unit test fails validation when the extra checks flag is set to `true`, on a bundle that will otherwise be valid.
- When the flag is set to `false` we verify that the extra checks are not being called

# Checklist:


src/main/java/com/iota/iri/service/ledger/impl/LedgerServiceImpl.java
```diff
@@ -133,7 +133,7 @@ public boolean isBalanceDiffConsistent(Set<Hash> approvedHashes, Map<Hash, Long>
         }
         Set<Hash> visitedHashes = new HashSet<>(approvedHashes);
         Map<Hash, Long> currentState = generateBalanceDiff(visitedHashes, tip,
-                snapshotProvider.getLatestSnapshot().getIndex());
+                snapshotProvider.getLatestSnapshot().getIndex(), true);
```
acha-bill: Forcing extra rules here cannot break backward compatibility?

GalRogozinski: Please verify that when we apply milestone we don't hit this code
Only in tip-selection we do

GalRogozinski:
@acha-bill 
Taking a discussion from slack here about the behavior of `GetBalances`.

I think this doesn't hurt previously confirmed txs because the snapshot is being inspected directly here:
```java
 for (final Hash address : addressList) {
                Long value = snapshotProvider.getLatestSnapshot().getBalance(address);
                if (value == null) {
                    value = 0L;
                }
                balances.put(address, value);
            }
```

PR #1806, opened 2020-03-20
White flag

### LINKED ISSUE #1684: Protocol Change: Ignore Conflicts (white-flag)
### Description
If a milestone approves conflicting transactions simply ignore one. Do this by ordering the approved cone in a deterministic way.
See:  https://iota.cafe/t/conflict-white-flag-mitigate-conflict-spamming-by-ignoring-conflicts/233

### Motivation
1. The abilty to merge conflicting subtangles should ensure a high CTPS.
2. Pave the way to different faster tipselection algorithms

### Requirements
1. Traverse all approved transactions in deterministic manner
2. Find conflicts as you traverse
3. Allow temporary overdrafts like before. For example if one spends, 10 iotas from an empty address and then later someone funds this address, it should be accepted.
4. Add a flag to the transaction structure that tells us whether
5. In case you are not approving conflicts, the resulting change should be the same as before.

### Proposed algorithm
1. When a milestone comes in start traversing in DFS from stopping at the *leafs* (transactions that has only approved parents).
2. Start applying bundles to the diffmap only as you go up the recursion stack (when you climb up back from the leafs)
3. Every time you apply to the diffmap check that it doesn't conflict with the last ledger state. If it does, store the conflicting bundle in a temporary set you put aside.
4. Overdraft handling: After you are traversing the graph attempt to apply again the conflicting bundles. Remove the ones successfuly applied from the set. Repeat this step until the size stopped changing.

### Implementation tips
For requirement 4, one can maybe use a negative `snapshotIndex` (make sure to use absolute value when reading it), or use the Bitmasks trick like we do in flags. This way we won't change the transaction schema.

acha-bill:
# Description

Ability to confirm conflicting bundle while applying on the first to the ledger.
The order of bundles is determined by a topological sort via DFS preferring trunk first.

Fixes #1684 

## Type of change

# How Has This Been Tested?

- Existing unit and regression tests pass
- Added new regression test where a milestone approves 2 double-spend bundles and asserts that even though only 1 affects the ledger, they are both confirmed.

# Checklist:


src/main/java/com/iota/iri/service/ledger/impl/LedgerServiceImpl.java
```diff
@@ -163,54 +163,76 @@ public boolean isBalanceDiffConsistent(Set<Hash> approvedHashes, Map<Hash, Long>
...
+            try {
+                Map<Hash, Long> currentState = new HashMap<>();
+                Hash transactionPointer = stack.peek();
+                TransactionViewModel transactionViewModel = TransactionViewModel.fromHash(tangle, transactionPointer);
+                TransactionViewModel trunk = transactionViewModel.getTrunkTransaction(tangle);
+                TransactionViewModel branch = transactionViewModel.getBranchTransaction(tangle);
+
+                boolean approvedTrunk = (trunk.snapshotIndex() > 0) && (trunk.snapshotIndex() != milestoneIndex);
+                boolean approvedBranch = (branch.snapshotIndex() > 0) && (branch.snapshotIndex() != milestoneIndex);
+                if ((trunk.isEmpty()  || visitedTransactions.contains(trunk.getHash()) || approvedTrunk) &&
```
GalRogozinski: One should note that an empty transaction is the same as the genesis transaction (genesis transaction has all 9's trytes, and all 9s Hash). Then it is fine...

If the transaction was empty because the trunk tx is simply missing (not solid) then we have a problem...

I know you went according to the RFC, I just added a comment on it as well...

src/main/java/com/iota/iri/controllers/TransactionViewModel.java
```diff
@@ -863,6 +863,30 @@ public void isMilestone(Tangle tangle, Snapshot initialSnapshot, final boolean i
...
+    }
+
+    /**
+     * Checks if a transaction is empty.
+     * @return True if empty. False otherwise.
+     */
+    public boolean isEmpty(){
+        return Arrays.equals(getBytes(), new byte[SIZE]);
+    }
+
```
GalRogozinski: Currently we are just check for `TransactionViewModel.PREFILLED_SLOT`
I am not saying that this is smart (I didn't make this weird design)

But it will be more awful if we have more than one way to do this check...
So I wouldn't have this method now

src/main/java/com/iota/iri/service/ledger/impl/LedgerServiceImpl.java
```diff
@@ -163,54 +162,80 @@ public boolean isBalanceDiffConsistent(Set<Hash> approvedHashes, Map<Hash, Long>
...
+            try {
+                Map<Hash, Long> currentState = new HashMap<>();
+                Hash transactionPointer = stack.peek();
+                TransactionViewModel transactionViewModel = TransactionViewModel.fromHash(tangle, transactionPointer);
+                TransactionViewModel trunkTransactionViewModel = transactionViewModel.getTrunkTransaction(tangle);
+                TransactionViewModel branchTransactionViewModel = transactionViewModel.getBranchTransaction(tangle);
+
+                boolean approvedTrunk = (trunkTransactionViewModel.snapshotIndex() > 0) && (trunkTransactionViewModel.snapshotIndex() != milestoneIndex);
+                boolean approvedBranch = (branchTransactionViewModel.snapshotIndex() > 0) && (branchTransactionViewModel.snapshotIndex() != milestoneIndex);
+                if ((trunkTransactionViewModel.getType() == TransactionViewModel.PREFILLED_SLOT  || visitedTransactions.contains(trunkTransactionViewModel.getHash()) || approvedTrunk) &&
```
DyrellC: Perhaps providing labels for these conditions as booleans instead of rewriting them in the if and else statements here could help to declutter this section. There are a lot of conditions being thrown around very close together here that could make it harder to follow. For example another boolean for each the trunk and branch types to be reused below in the else statement. More descriptive labels or a few extra comments could go a long way towards making the logic easier for others to follow.

python-regression/tests/features/machine3/3_transaction_tests.feature
```diff
@@ -87,3 +87,47 @@ Feature: Test transaction confirmation
             | keys   | values | type          |
             | states | False  | boolListMixed |
 
+    Scenario: Conflicting bundles are ignored
```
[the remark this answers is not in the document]
kwek20: Already done in my PR, theyre in the disabled regression tests file in machine 3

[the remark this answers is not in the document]
GalRogozinski: was just about to tag you here @kwek20 but I see you are very alert :-)

src/main/java/com/iota/iri/service/ledger/impl/LedgerServiceImpl.java
```diff
@@ -175,6 +175,18 @@ public boolean isBalanceDiffConsistent(Set<Hash> approvedHashes, Map<Hash, Long>
...
+                Hash genesisHash = Hash.NULL_HASH;
+                if(isEmptyTrunk  && !allowGenesisReference){
+                    return null;
+                 }
+                //proceed only if empty trunk or branch is genesis
+                if((isEmptyTrunk && !trunkTransactionViewModel.getHash().equals(genesisHash))||
+                (isEmptyBranch && !branchTransactionViewModel.getHash().equals(genesisHash))){
+                    return null;
+                }
+
```
GalRogozinski: 1. In the event that trunk or branch have a non `9s` hash then this is wrong
2. I believe (I think I missed it in the last review so my bad) that the `9s` genesis hash is in the `solidEntryPointsSet`

So I think the flag is unnecessary (because we should always allow genesis reference)
I think we should always return null if it is `PREFILLED_SLOT`

acha-bill: Since we return `null` if `branch` or `trunk` is `PREFILLED_SLOT`. 
Is this the only definition of a `leaf`?
```
boolean isLeafTrunk = visitedTransactions.contains(trunkTransactionViewModel.getHash()) || isApprovedTrunk;
```

GalRogozinski: Yes, when `visitedTransactions` contain all SEPs and genesis

src/main/java/com/iota/iri/service/ledger/impl/LedgerServiceImpl.java
```diff
@@ -164,55 +163,84 @@ public boolean isBalanceDiffConsistent(Set<Hash> approvedHashes, Map<Hash, Long>
...
+                    stack.pop();
                 }
-
-            } catch (Exception e) {
+                else if(!isLeafTrunk){
+                    stack.addFirst(trunkTransactionViewModel.getHash());
+                }else if(!isLeafBranch) {
+                    stack.addFirst(branchTransactionViewModel.getHash());
+                }
+            }catch(Exception e){
```
GalRogozinski, line 240:
```suggestion
            } catch(Exception e) {
```

PR #1768, opened 2020-02-27
Remove threshold from getbalances API call

acha-bill:
# Description

Removes the unused `threshold` parameter from `getBalances` API call.

Fixes #1723 

## Type of change

- Bug fix (a non-breaking change which fixes an issue)

# How Has This Been Tested?

- Existing unit tests pass
- Manually called `getBalances` on a node with and without threshold.

# Checklist:

- [x] My code follows the style guidelines for this project
- [x] I have performed a self-review of my own code
- [x] New and existing unit tests pass locally with my changes

PR #1808, opened 2020-03-25
Removed threshold

kwek20:
Forgotten from #1768


GalRogozinski:
Thanks,
It is kind of weird that #1768 was merged to `release-v1.9.0` and this one is to `dev`
Can we just put it in order please?
You can force push here after you fix the branch

Thanks

PR #1817, opened 2020-03-29
Fix: release next key

### LINKED ISSUE #1816: Releasing the wrong key
### Bug description
In order to release from cache we first peek, then save, then release
```java
 private static boolean cachePersistAndReleaseNext(Tangle tangle, Cache<Indexable, TransactionViewModel> cache) throws Exception {
        Indexable hash = cache.getNextReleaseKey();
        TransactionViewModel tvm = cache.get(hash);
        if (tvm != null && tvm.shouldPersist()) {
            //We update and cache this item again so that in the case where this released item is brought back from
            // weakStore, it does not write to DB again.
            tvm.setShouldPersist(false);
            cache.put(hash, tvm);
            tangle.save(tvm.getTransaction(), hash);
        }
        return cache.releaseNext();
    }
```

However due to a context switches and multi-threading between peek and releaseNext() may not use the same hash 1.9.0
### IRI version
1.9.0

GalRogozinski:
# Description

Ensures that only the saved key is released

Fixes #1816 

## Type of change

# How Has This Been Tested?

There are current unit tests.
I am suspecting that this caused solidification problem.
More tests should commence

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/cache/Cache.java
+++ b/src/main/java/com/iota/iri/cache/Cache.java
@@ -46,10 +46,11 @@ public interface Cache<K, V> {
 
     /**
      * Release the specified key from the cache and puts it in the weak store.
-     * 
+     *
      * @param key
+     * @return {@code true} if released or {@code false} if key was never in cache
      */
-    void release(K key);
+    boolean release(K key);
 
     /**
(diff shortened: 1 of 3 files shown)


GalRogozinski:
I think the current name and docs describe the method correctly

PR #1805, opened 2020-03-19
Add transaction solidifier

DyrellC:
# Description
This contains the bulk of changes from #1646 and relies on #1804. Transaction solidification should not be contained within the `TransactionValidator` and depending on the number of transactions required for a walk, using the `checkSolidity` call within the `TransactionValidator`, as is done in the `MilestoneSolidifier`, can take a significant portion of time. In order to mitigate this, the solidification and propagation logic has been placed into its own class running in its own thread independent of the validation logic. Instead of making direct calls to `checkConsistency` a queue is formed so as to not lock up any other threads, and the solidity of a transaction will first be checked via the view model before it is placed into the queue.

This is a step towards introducing the milestone pipeline refactoring.

 ## Type of change
- Enhancement (a non-breaking change which adds functionality)

# How Has This Been Tested?
- All corresponding tests pass: Regression, Unit and Synchronisation

# Checklist:
- [x] My code follows the style guidelines for this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #1819, opened 2020-03-31
reset getInclusion reg test assertion to what it should be

acha-bill:
# Description

In `Value Transactions are confirmed`, 10 txs are issued on a node and all are referenced by a milestone. All the txs are supposed to be confirmed. Yet, the last was not. This was happening on buildkite.
I thought there was a race condition on buildkite. So I *fixed the test* instead. This was an effort to allow other tests to pass.

This PR resets it to what it should have been in the first place.

## Type of change

# Checklist:


GalRogozinski:
I see buildkite fails with this...
Is it dependent on Dyrell's #1805 ?

By the way are you getting mail notifications when buildkite fails?

acha-bill:
No. It does not depend on #1805 
Buildkite is acting up again.

As you can see, today this PR passed https://github.com/iotaledger/iri/pull/1820 while all other  have been failing for the above reason before today. Weird

I don't receive emails from buildkite.

Issue #1674, opened 2019-11-28
Feature: Update Milestone Solidification


DyrellC:
### Description
The current milestone solidification logic relies on a 3 step process that is rather disjointed. Milestones are found in the `LatestMilestoneTracker` (LMT) through an address search and then processed through milestone/bundle validity checks before being placed into the `MilestoneSolidifier` (MS) where it will then go through a separate process of transaction requesting and solidity checking to fetch missing reference transactions for milestone candidates. When a milestone is solidified through this process a milestone object is created in the DB. Lastly the `LatestSolidMilestoneTracker` (LSMT) searches sequentially for the next milestone object by index and updates the nodes solidity index by index. This strikes me as an unnecessarily disconnected process. With the new `TransactionProcessingPipeline`(TPP) following the last network rewrite, we can utilise the processing stages to more directly handle milestones and milestone reference requests to streamline milestone solidification and process them in a more connected order that carries the hash references straight through to the LSMT checks. Currently the LMS checks rely on transaction addresses and hash references to pass to the MS, whereas the LSMT is reliant on the milestone index, which is only created when the MS successfully solidifies the transaction. The MS does not rely on index while solidifying and may produce the occasional false `INVALID` or `INCOMPLETE` which then stops a specific Milestone from solidifying correctly. 

### Motivation
To process milestone objects more efficiently and improve solidification and request logic. 

### Requirements
[ ] New transaction requesting logic using the TPP
[ ] New milestone stages and payload processing through the TPP
[ ] Rewrite/Consolidate milestone solidification pipeline with better milestone hash/index mapping

### Am I planning to do it myself with a PR?
Yes

GalRogozinski:
Relates to #1448

DyrellC:
I would like to outline my proposal for changes to milestone processing. With the TPP in place and the new solidification process coming into play, we now have a good opportunity to consolidate behaviour into a less disjointed flow. The solidification of milestones can be handled from within the TPP rather than as an external set of threads which each run independently of one another. 
As is currently,  the processing flow looks something like this: 
![image](https://user-images.githubusercontent.com/34013004/70567969-2f1f3a80-1b54-11ea-8f39-ac9d0d5f7d87.png)

With the addition of the solidification stage it's starting to become more streamlined as follows: 
![image](https://user-images.githubusercontent.com/34013004/70568024-45c59180-1b54-11ea-9d49-485764faad24.png)

[...]

PR #1821, opened 2020-04-01
Add transaction solidifier queue

DyrellC:
# Description
Decouple the milestone and transaction solidification processes to handle transaction solidification asynchronously from milestone checking. Currently each time the `MilestoneSolidifier` checks the solidity of a transaction, it does the `TransactionSolidifier`'s `checkSolidity` call directly. This can cause a hanging when processing longer sync queues, which can delay the `MilestoneSolidifier` processing newer received milestone transactions. In order to decouple this, the `MilestoneSolidifier` will check the solid state of a transaction and add it to a processing queue in the `TransactionSolidifier` so that there are no hanging moments during milestone processing. This is a step introduced within the process of updating milestone handling as per #1674.

Follow up to #1805 
 
## Type of change

# How Has This Been Tested?
- All previously created tests still work, new unit tests work, and regression tests still pass locally (with the exception of the already known bug for machine3) 

# Checklist:


src/main/java/com/iota/iri/service/validation/impl/TransactionSolidifierImpl.java
```diff
@@ -87,6 +104,43 @@ public void shutdown() {
...
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    public boolean addMilestoneToSolidificationQueue(Hash hash){
+        try{
+            TransactionViewModel tx = fromHash(tangle, hash);
+            if(tx.isSolid()){
```
GalRogozinski, line 132:
```suggestion
            if (tx.isSolid()) {
```

src/main/java/com/iota/iri/service/validation/impl/TransactionSolidifierImpl.java
```diff
@@ -87,6 +104,43 @@ public void shutdown() {
...
+                transactionPropagator.addToPropagationQueue(hash);
+                return true;
+            }
+            addToSolidificationQueue(hash);
+            return false;
+        }catch(Exception e){
```
GalRogozinski, line 138:
```suggestion
        } catch (Exception e) {
```

src/main/java/com/iota/iri/service/validation/impl/TransactionSolidifierImpl.java
```diff
@@ -216,6 +280,12 @@ private void addToBroadcastQueue(TransactionViewModel tvm) {
         }
     }
 
+    @VisibleForTesting
+    Set<Hash> getSolidificationQueue(){
```
GalRogozinski, line 284: It is funny that you call it `getQueue` but you return a `Set`

src/test/java/com/iota/iri/service/validation/impl/TransactionSolidifierImplTest.java
```diff
@@ -98,11 +98,22 @@ public void verifyTxIsNotSolid() throws Exception {
         assertFalse("Expected transaction to fail solidity check", txSolidifier.checkSolidity(tx.getHash()));
     }
 
+    @Test
+    public void getSolidificationQueue() throws Exception {
```
GalRogozinski, line 102: You are testing `add` to solidification queue not `get`, so better rename

src/main/java/com/iota/iri/service/validation/impl/TransactionSolidifierImpl.java
DyrellC: A fair point, i'll rename

PR #1822, opened 2020-04-01
Add milestone stage

DyrellC:
# Description
This is step one towards the migration of milestone handling logic as per #1674. In this phase we add a milestone stage to the processing pipeline to handle milestone transactions as they are received rather than relying on the `LatestMilestoneTracker` to pull them from the db. This should mitigate the issue demonstrated on devnet, when a node has a high number of milestone transactions present in the db, and it takes a long time to pull them out. New milestones were not being seen and processed fast enough, so to fix that, we handle them straight from the received stage. 

This is a follow up to #1821 

Fixes #1688    
Also relevant to #1448 and #1797 

## Type of change
- Enhancement (a non-breaking change which adds functionality)

# How Has This Been Tested?
- Existing unit and regression tests have continued to pass with this migration

# Checklist:
- [x] My code follows the style guidelines for this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #1823, opened 2020-04-01
Consolidate milestone logic

DyrellC:
# Description
This is the second phase to #1674. Currently we have 3 separate classes handling different parts of the solidification process for milestones, each with their own thread, handling milestone objects in different ways. The `LatestMilestoneTracker` loads all transactions with the coo address form the db each time it runs through a cycle to find new unprocessed milestones and pass them to the `MilestoneSolidifier`, which is resource and time consuming. The `MilestoneSolidifier` holds several mappings of milestone transactions (hash : index) to be placed into processing after an analysis determines the oldest 20 milestones in the solidification queue. It then puts these transactions through calls to determine the solidity of a transaction, and removes the milestone from the queue once it has been determined that the transaction is solid (and therefore a milestone object has been created through the `TransactionSolidifier.checkSolidity` ->  Bundle Validation -> Ledger services). Lastly the `LatestSolidMilestoneTracker` iterates through milestone objects by index and checks that the milestone exists, then logging the change and performing updates to snapshots etc. This disjointed methodology not only creates confusion when following the logic of the code, it also has some performance drawbacks that can be mitigated by handling the milestones through the transaction processing pipeline, and a single solidification class instead of 3 to map potential milestone objects without holding unnecessary mappings that take up additional memory.   

This is a follow up to #1822 

## Type of change

# How Has This Been Tested?
- Existing regression and unit tests pass with the migration

# Checklist:


src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
```diff
@@ -1,339 +1,666 @@
...
+            } catch(Exception e) {
+                log.error("Error running milestone solidification thread", e);
+            }
+        }
+    }
 
     /**
-     * Holds all unsolid milestones that shall be solidified (the transaction hash mapped to its milestone index).
+     * Scan through milestones in the {@link #unsolidMilestones} queue and refill the {@link #solidificationQueue} queue.
+     * It then iterates through this queue to determine iff the milestone has already been seen and validated. If it is
```
GalRogozinski, line 142:
```suggestion
     * It then iterates through this queue to determine if the milestone has already been seen and validated. If it is
```

src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
```diff
@@ -1,339 +1,666 @@
...
+            solidificationQueue = newSolidificationQueue;
+        }
+
+        // update the oldest milestone in queue
+        oldestMilestoneInQueue = null;
+        for (Map.Entry<Hash, Integer> currentEntry : solidificationQueue.entrySet()) {
+            if (seenMilestones.containsKey(currentEntry.getValue())) {
+                // remove candidate from pool
+                unsolidMilestones.remove(currentEntry.getKey());
+                solidificationQueue.remove(currentEntry.getKey());
```
GalRogozinski, line 170: why are those values can be in the queue??
Maybe better to check upon insertion to the queues?

src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
```diff
@@ -1,339 +1,666 @@
...
-    private static final int SOLIDIFICATION_INTERVAL = 100;
+    private Map<Hash, Integer> unsolidMilestones = new ConcurrentHashMap<>();
+    private Map<Hash, Integer> solidificationQueue = new ConcurrentHashMap<>();
+    private Map<Integer, Hash> seenMilestones = new ConcurrentHashMap<>();
+
+    private Map.Entry<Hash, Integer> oldestMilestoneInQueue = null;
+
+    private int latestMilestoneIndex;
+    private Hash latestMilestoneHash;
+    private int latestSolidMilestone;
```
GalRogozinski, line 41: adding all those state fields is usually a red flag...
In a multithreaded environment you want to avoid this as much as possible

src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
```diff
@@ -1,339 +1,666 @@
...
      */
-    private final SnapshotProvider snapshotProvider;
+    private int repairBackoffCounter = 0;
 
     /**
-     * Holds a reference to the transactionSolidifier which allows us to issue solidity checks.
+     * An indicator for whether or not the solidifier has processed from the first batch of seen milestones
      */
-    private final TransactionSolidifier transactionSolidifier;
+    private boolean initialized = false;
```
GalRogozinski, line 67: also a bit scary

python-regression/tests/features/machine1/1_local_snapshots_tests.feature
```diff
@@ -53,7 +53,7 @@ Feature: Test Bootstrapping With LS
     #Default for test is to issue 10324
     When milestone 10324 is issued on "nodeA-m1"
     #Give the node time to finish syncing properly, then make sure that the node is synced to the latest milestone.
-    And we wait "120" second/seconds
+    And we wait "20" second/seconds
```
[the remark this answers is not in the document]
DyrellC: No the failure in buildkite was due to me removing the `scanMilestonesInQueue()` from the main thread. The failure was due to the pruning not starting because the initialized flag wasn't set to true until a new milestone became present. You can tell by looking at the failures in the tests. They were pertaining to a failure of the node to take the localsnapshot on startup and waiting for the first issued milestone. I just pushed a change to re-include that call. Tests should pass again after that with the reduced times still.

src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
```diff
@@ -1,339 +1,666 @@
...
+                    logChange(snapshotProvider.getInitialSnapshot().getIndex());
+
+                    syncProgressInfo.setSyncMilestoneStartIndex(snapshotProvider.getInitialSnapshot().getIndex());
+                }
+
+                solidifySeenMilestones();
+                processSolidifyQueue();
+                checkLatestSolidMilestone();
+
+                if (latestMilestoneIndex > latestSolidMilestone) { solidifyLog();}
```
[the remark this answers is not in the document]
DyrellC: I did not :sweat_smile: i'll pair it with some other syntax changes and push later

src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
```diff
@@ -1,339 +1,666 @@
...
+     * the {@link #solidificationQueue}.
+     *
+     * If upon scanning the {@link #unsolidMilestones} queue is empty and {@link #initialized} is false, set
+     * {@link #initialized} to true.
      */
-    private final Map<Hash, Integer> unsolidMilestonesPool = new ConcurrentHashMap<>();
+    private void scanMilestonesInQueue() {
+        // refill the solidification queue with solidification candidates sorted by index
+        if (unsolidMilestones.size() > 0) {
+            Map<Hash, Integer> newSolidificationQueue = new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
DyrellC: Thinking about it more, should be able to remove that and just clear the original queue and refill it, no need to make the new queue. Although the queue only holds a mapping of 10 hashes to integer pairings, so the memory footprint is minor for each object, but none is better than a little in this instance so i'll update that.

src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
DyrellC: To help alleviate some of the concern i've switched the logic to now use atomic integers and atomic references to try and shield some of the possible issues that could surface.

src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
DyrellC: Also switched to atomic, this was there before as well.

src/main/java/com/iota/iri/network/pipeline/MilestoneStage.java
```diff
@@ -57,48 +53,52 @@ public MilestoneStage(Tangle tangle, MilestoneSolidifier milestoneSolidifier,
...
-            if (!transactionSolidifier.addMilestoneToSolidificationQueue(milestone.getHash()) &&
-            isFirstInBundle) {
-                milestoneSolidifier.add(milestone.getHash(), payload.getMilestoneIndex());
-            } else {
+            // Add milestone tails to the milestone solidifier, if transaction is solid, add to the propagation queue
+            if (isTail) {
+                milestoneSolidifier.addMilestoneCandidate(milestone.getHash(), newMilestoneIndex);
+            }
+
+            if (transactionSolidifier.addMilestoneToSolidificationQueue(milestone.getHash())) {
```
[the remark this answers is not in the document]
DyrellC: You make a good point, I agree it should only be used from within the mielstoneSolidifier.

src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
DyrellC: I'm going to move it to the validation check, since that's the only place a transaction ends up in the seenMilestones pool. This was left over logic from when I had insertion to the seenMilestones pool from the `MilestoneStage`.

PR #1826, opened 2020-04-05
Solidify transactions we walk across while doing GTTA

### LINKED ISSUE #1818: Solidify unsolid transactions encountered during the walk
### Description
while running the walker due to GTTA calls we should solidify transactions we encounter on a separate thread

### Motivation
Due to network/node unreliability some transactions may not become solid on a node but they may appear as solid on other nodes.

If the node is public, this may cause users to not have their transactions confirmed...

### Requirements
1. Upon calling GTTA we try to solidify unsolid transactions the walker encounters, on a separate thread
2. It shouldn't affect any other API calls (i.e. CheckConsistency)

GalRogozinski:
# Description

While doing GTTA we will attempt to solidify transactions we walk across.
This will make sure that transactions that we missed out on for whatever reason will solidify

Fixes #1818

## Type of change

# How Has This Been Tested?

I ran both 1.8.5 and a canary with this change.
I called GTTA in a loop on both with JMETER

GTTA with this change was still faster compared to 1.8.5.
Throughput: 176.662 ops/minute compared to 71.603 ops/minute

However, profiling showed that solidifying takes 30 % of cpu time...

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/service/tipselection/impl/AsyncTipSelSolidifier.java
+++ b/src/main/java/com/iota/iri/service/tipselection/impl/AsyncTipSelSolidifier.java
@@ -2,6 +2,7 @@ package com.iota.iri.service.tipselection.impl;
 
 import com.iota.iri.TransactionValidator;
 import com.iota.iri.model.Hash;
+import com.iota.iri.service.milestone.MilestoneSolidifier;
 import com.iota.iri.service.tipselection.TipSelSolidifier;
 
 import java.util.Set;
@@ -40,7 +41,7 @@ public class AsyncTipSelSolidifier implements TipSelSolidifier {
             solidExecutor.submit(() -> {
                 try {
                     log.debug("attempting to solidify transaction {}", transactionHash);
-                    transactionValidator.checkSolidity(transactionHash);
+                    transactionValidator.checkSolidity(transactionHash, MilestoneSolidifier.SOLIDIFICATION_TRANSACTIONS_LIMIT);
                 } catch (Exception e) {
                     log.error("Failed to solidify transaction during a walk", e);
                 }
--- a/src/main/java/com/iota/iri/service/tipselection/impl/AsyncTipSelSolidifier.java
+++ b/src/main/java/com/iota/iri/service/tipselection/impl/AsyncTipSelSolidifier.java
@@ -1,0 +1,50 @@
+package com.iota.iri.service.tipselection.impl;
+
+import com.iota.iri.TransactionValidator;
+import com.iota.iri.model.Hash;
+import com.iota.iri.service.tipselection.TipSelSolidifier;
+
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ThreadFactory;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class AsyncTipSelSolidifier implements TipSelSolidifier {
+
+
+    private static final Logger log = LoggerFactory.getLogger(AsyncTipSelSolidifier.class);
+
+    private final TransactionValidator transactionValidator;
+    private final Set<Hash> transactionToSolidify = ConcurrentHashMap.newKeySet();
+    private final ExecutorService solidExecutor;
+
+    public AsyncTipSelSolidifier(TransactionValidator transactionValidator) {
+        this.transactionValidator = transactionValidator;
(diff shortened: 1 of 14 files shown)


src/main/java/com/iota/iri/service/tipselection/impl/AsyncTipSelSolidifier.java
```diff
@@ -0,0 +1,51 @@
...
+    private final TransactionValidator transactionValidator;
+    private final Set<Hash> transactionToSolidify = ConcurrentHashMap.newKeySet();
+    private final ExecutorService solidExecutor;
+
+    public AsyncTipSelSolidifier(TransactionValidator transactionValidator) {
+        this.transactionValidator = transactionValidator;
+        ThreadFactory threadFactory = new ThreadFactoryBuilder()
+                .setNameFormat(this.getClass().getSimpleName() + " %d")
+                .build();
+        solidExecutor = Executors.newCachedThreadPool(threadFactory);
```
GalRogozinski, line 31: Perhaps it needs to be changed to another executor?

src/main/java/com/iota/iri/service/tipselection/impl/AsyncTipSelSolidifier.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+
+    @Override
+    public void solidify(Hash transactionHash) {
+        // Don't try to solidify transactions that are in the process of being solidified
+        //Problem: we are not checking transactions in the past cone here
+        if (transactionToSolidify.add(transactionHash)) {
```
DyrellC, line 39: As you pointed out this has the potential to get DOS attacked if there is enough power behind the spammers being thrown at the node. Does it make sense maybe to have a limit to the maximum number of objects within the set?

src/main/java/com/iota/iri/service/tipselection/impl/AsyncTipSelSolidifier.java
```diff
@@ -0,0 +1,51 @@
...
+            solidExecutor.submit(() -> {
+                try {
+                    log.debug("attempting to solidify transaction {}", transactionHash);
+                    transactionValidator.checkSolidity(transactionHash);
```
DyrellC, line 43: As Brord pointed out, we probably shouldn't allow this to run up to MAX_INTEGER and limit it to something like 10k analysed transactions

src/main/java/com/iota/iri/service/tipselection/impl/TipSelectorImpl.java
```diff
@@ -11,11 +11,7 @@
 import com.iota.iri.model.Hash;
 import com.iota.iri.service.ledger.LedgerService;
 import com.iota.iri.service.snapshot.SnapshotProvider;
-import com.iota.iri.service.tipselection.EntryPointSelector;
-import com.iota.iri.service.tipselection.RatingCalculator;
-import com.iota.iri.service.tipselection.TipSelector;
-import com.iota.iri.service.tipselection.WalkValidator;
-import com.iota.iri.service.tipselection.Walker;
+import com.iota.iri.service.tipselection.*;
```
DyrellC, line 14: Not really a review on this PR specifically, but can we do something about Codacy freaking out about these kinds of imports?

GalRogozinski: yeah I know...
I need to find time to take care of this, so much to do...

src/main/java/com/iota/iri/service/tipselection/impl/AsyncTipSelSolidifier.java
GalRogozinski: I will use the same solidification limit as the milestone solidifier
If it is too much here, it is too much for milestones

src/main/java/com/iota/iri/service/tipselection/impl/AsyncTipSelSolidifier.java
GalRogozinski: When I tested this I counted the number of AsyncSolidifier threads created when I bombed the node with Jmeter spam

I think there are more chances of bombing the CPU out before getting an oom from this

GalRogozinski:
@kwek20 

> What made you chose this dummy solidifier instead of a null check? Probably same speed after optimizing, but just wondering. I would generally not chose the empty class approach.

I am just wired to use polymorphism instead of `if-else`. In some cases it makes way shorter readable code. In this case, it matters less and maybe the `null` check is better. The "advantage" here is that now you are ready to inject other imps in other places the WalkValidator is called from. But we probably won't use it. Anyhow, this change will be changed by Dyrell and he can choose another approach

> Theres no issue linked
Fixed :-)

PR #1827, opened 2020-04-07
Update Broadcast Stage logic 

DyrellC:
# Description
When the transaction solidifier was introduced, we updated how transactions are passed through to broadcast, and under initial review we opted to fetch broadcast transactions at the end of the solidification stage, to pass them forward to the broadcast stage in the event the original solidifying transaction is not solid. This sequential loading of transactions from the broadcast queue does not forward transactions fast enough. To correct this, the logic has been moved to the broadcast stage instead. The broadcast stage will complete the transaction forwarding as it was previously designed to do, and the broadcast stage will check if a transaction still exists to broadcast, and if there is, it will pass it to the broadcast stage queue until there is nothing left to pull out of the set. 

Fixes an error that occurred in high tps environments that would cause a delay in the forwarding of transactions via broadcast 

## Type of change

# How Has This Been Tested?
Using the same topology that the problem was first observed in (within the ICC network) transactions, specifically milestone transactions, are being forwarded in a timely fashion. 

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/network/pipeline/BroadcastStage.java
+++ b/src/main/java/com/iota/iri/network/pipeline/BroadcastStage.java
@@ -3,6 +3,7 @@ package com.iota.iri.network.pipeline;
 import com.iota.iri.controllers.TransactionViewModel;
 import com.iota.iri.network.NeighborRouter;
 import com.iota.iri.network.neighbor.Neighbor;
+import com.iota.iri.service.validation.TransactionSolidifier;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/iota/iri/network/pipeline/BroadcastStage.java
```diff
@@ -54,6 +58,15 @@ public ProcessingContext process(ProcessingContext ctx) {
...
         }
 
+        // Check the transaction solidifier to see if there are solid transactions that need to be broadcast.
+        // If so, forward them to the BroadcastStageQueue to be processed.
+        TransactionViewModel transactionToBroadcast;
+        if((transactionToBroadcast = transactionSolidifier.getNextTxInBroadcastQueue()) != null){
+            ctx.setNextStage(TransactionProcessingPipeline.Stage.BROADCAST);
+            ctx.setPayload(new BroadcastPayload(payload.getOriginNeighbor(), transactionToBroadcast));
+            return ctx;
+        }
```
GalRogozinski, line 68: hmm I thought you were going to do a `while` loop

I guess what you did here works as well
Not sure about performance what is better?

Did you test?

DyrellC: I did test it in ICC and there were no issues with this approach. Originally I was going to do a while, but I was worried about holding up the forwarding of received transactions. With this approach it will mesh the received and the solidified transactions in the queue for the broadcasts.

PR #1832, opened 2020-04-09
Fix visibility problem on shouldPersist and refactor tvm update methods

acha-bill:
# Description

Txs that were being quick set solid were not being persisted upon release from cache.
This change ensures that updates are always persisted by assuming responsibility to always call `tvm.update`
Also, fixes is a possible race condition with `tvm.shouldPersist` and cache metric fields.

## Type of change

# How Has This Been Tested?

- Existing and new unit tests that assert persistence pass.
- Buildkite will run reg tests :)

# Checklist:

### DIFF
--- a/src/main/java/com/iota/iri/controllers/TransactionViewModel.java
+++ b/src/main/java/com/iota/iri/controllers/TransactionViewModel.java
@@ -11,6 +11,7 @@ import com.iota.iri.utils.Converter;
 import com.iota.iri.utils.Pair;
 
 import java.util.*;
+import java.util.concurrent.atomic.AtomicBoolean;
 
 /**
  * Controller class for {@link Transaction} sets. A {@link TransactionViewModel} stores a {@link HashesViewModel} for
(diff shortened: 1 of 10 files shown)


DyrellC:
I agree with you @galrogo :+1:

src/main/java/com/iota/iri/controllers/TransactionViewModel.java
```diff
@@ -805,12 +805,15 @@ public static void updateSolidTransactions(Tangle tangle, Snapshot initialSnapsh
      *
      * Used by the {@link com.iota.iri.TransactionValidator} to quickly set the solidity of a {@link Transaction} set.
      *
+     * @param tangle          The tangle reference for the database.
+     * @param initialSnapshot snapshot that acts as genesis
      * @param solid The solidity of the transaction in the database
      * @return True if the {@link Transaction#solid} has been updated, False if not.
      */
-    public boolean updateSolid(boolean solid) throws Exception {
+    public boolean updateSolid(Tangle tangle, Snapshot initialSnapshot, boolean solid) throws Exception {
```
DyrellC, line 813: I like this change the most, it just simplifies a bit of a convoluted means of achieving the same thing as we had before.

kwek20:
I have asked @luca-moser for feedback on the original decision. Might give us some insights.
Either way, making it volatile would reduce speed across threads quite a bit, as every time it changes, it writes all fields in the TVM to global memory. so AtomicBool is actually nicer here.

PR #1842, opened 2020-04-27
Logs and bug patches for milestone solidification 

DyrellC:
# Description of change
The addition of the milestone stage and consolidation of milestone logic has resulted in a couple of small bugs that need patching before we merge with dev. 

The first is that milestones were being registered in the milestone stage, which meant that the latest milestone object would be updated prior to validation. Removing this from the milestone stage fixes that issue. 

Secondly the thread did not have a sleep timer at all, and it was eating up more cpu and memory than necessary. A 500 millisecond sleep mitigates this issue without damaging performance. 

Thirdly I added a couple of additional logs to improve understanding of the code. 

## Type of change
- Bug Fix

Ran nodes on an internal test network and profiled them to make sure they worked, and didn't overload the cpu as much. 

## Change checklist

### DIFF
--- a/src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
+++ b/src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
@@ -144,7 +144,7 @@ public class MilestoneSolidifierImpl implements MilestoneSolidifier {
             unsolidMilestones.entrySet().stream()
                     .sorted(Map.Entry.comparingByValue())
                     .forEach(milestone -> {
-                        if (solidificationQueue.size() < MAX_SIZE) {
+                        while (solidificationQueue.size() < MAX_SIZE) {
                             solidificationQueue.put(milestone.getKey(), milestone.getValue());
                         }
                     });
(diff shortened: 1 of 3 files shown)


src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
```diff
@@ -260,18 +269,24 @@ private void bootStrapSolidMilestones() throws Exception {
...
         setLatestSolidMilestone(milestoneIndex);
-
-        AddressViewModel.load(tangle, config.getCoordinator()).getHashes().forEach(hash -> {
+        Set<Hash> milestoneTransactions = AddressViewModel.load(tangle, config.getCoordinator()).getHashes();
+        int processed = 0;
+        int index;
+        for (Hash hash: milestoneTransactions) {
             try {
-                int index;
+                processed += 1;
```
GalRogozinski, line 277: not that I care or that you should change
but why not `++processed`?

src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
```diff
@@ -260,18 +269,24 @@ private void bootStrapSolidMilestones() throws Exception {
...
                 if ((index = milestoneService.getMilestoneIndex(TransactionViewModel.fromHash(tangle, hash))) >
                         getLatestSolidMilestoneIndex()) {
                     addMilestoneCandidate(hash, index);
                 }
+
+                if (processed % 1000 == 0 || processed % milestoneTransactions.size() == 0){
+                    log.info("Bootstrapping milestones: [ " + processed  + " / " + milestoneTransactions.size() + " ]");
+                }
```
GalRogozinski, line 285: We also have the `IntervalLogger`, but it is fine
you don't have to change :-)

PR #1845, opened 2020-04-29
Fix: max propagate up to 10 transactions for solidification

DyrellC:
# Description of change
After removing the cache with the dev branch merge, the propagation process takes a little longer to complete, and holds up the transaction solidifier processing time. To mitigate this, the propagation queue will be processed in batches to avoid the issue of holding up transactions in the solidification queue from being processed. Also, a small bug fix to the milestone solidifier to make sure we are registering new milestones that are not solid but valid.  

## Type of change

Tests that failed the merge into dev are passing now.  

## Change checklist

### DIFF
--- a/src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
+++ b/src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
@@ -197,6 +197,7 @@ public class MilestoneSolidifierImpl implements MilestoneSolidifier {
             switch(validity) {
                 case VALID:
                     milestoneCandidate.isMilestone(tangle, snapshotProvider.getInitialSnapshot(), true);
+                    registerNewMilestone(getLatestMilestoneIndex(), milestoneIndex, milestoneHash);
                     if (milestoneCandidate.isSolid()) {
                         removeFromQueues(milestoneHash);
                         addSeenMilestone(milestoneHash, milestoneIndex);
@@ -205,9 +206,7 @@ public class MilestoneSolidifierImpl implements MilestoneSolidifier {
                     }
                     break;
                 case INCOMPLETE:
-                    if(milestoneIndex > getLatestMilestoneIndex()) {
-                        registerNewMilestone(getLatestMilestoneIndex(), milestoneIndex, milestoneHash);
-                    }
+                    registerNewMilestone(getLatestMilestoneIndex(), milestoneIndex, milestoneHash);
                     transactionSolidifier.addToSolidificationQueue(milestoneHash);
                     break;
                 case INVALID:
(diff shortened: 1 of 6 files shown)


src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
```diff
@@ -205,9 +206,7 @@ private void processSolidifyQueue() throws Exception {
                     }
                     break;
                 case INCOMPLETE:
-                    if(milestoneIndex > getLatestMilestoneIndex()) {
-                        registerNewMilestone(getLatestMilestoneIndex(), milestoneIndex, milestoneHash);
-                    }
+                    registerNewMilestone(getLatestMilestoneIndex(), milestoneIndex, milestoneHash);
```
GalRogozinski, line 209: why are you also registering milestones that may be fake?

src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
```diff
@@ -197,6 +197,7 @@ private void processSolidifyQueue() throws Exception {
             switch(validity) {
                 case VALID:
                     milestoneCandidate.isMilestone(tangle, snapshotProvider.getInitialSnapshot(), true);
+                    registerNewMilestone(getLatestMilestoneIndex(), milestoneIndex, milestoneHash);
```
GalRogozinski, line 200: I am actually not sure why it is a bug to wait until the milestone is solid before registering...
Care to elaborate?

src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
acha-bill: The check has been moved to `registerNewMilestone`

src/main/java/com/iota/iri/service/validation/impl/TransactionSolidifierImpl.java
```diff
@@ -387,8 +395,11 @@ public void addToPropagationQueue(Hash hash) throws Exception{
 
         @VisibleForTesting
         void propagateSolidTransactions() {
-            while(!Thread.currentThread().isInterrupted() && solidTransactions.peek() != null) {
+            int processed = 0;
+            while(!Thread.currentThread().isInterrupted() && solidTransactions.peek() != null
+                    && processed < PROPAGATION_QUEUE_MAX_PROCESS) {
```
acha-bill, line 400: Nice change

src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
DyrellC: That was not the bug as it happens. The issue was in the propagation stalling out, but since I had made this change prior to fixing that I was under the false assumption that this part of it was necessary. As it turns out it does not look to be the case, so I've reverted it to only register milestones when being added to the seen milestone queue. Same as before, but a little cleaner with the check in the `registerNewMilestone` method.

PR #1848, opened 2020-05-05
Update milestone bootstrapping

DyrellC:
# Description of change
Milestone bootstrapping in the solidifier was registering solid milestones before the snapshot provider, and therefore the milestones were not being applied correctly. This caused a discrepancy between the `getNodeInfo` latest solid milestone vs the `MilestoneSolidifier` latest solid milestone. It also stopped milestone candidates that were between these 2 separate values from being properly processed. The change here ensures that any valid and incomplete milestone candidates are properly placed into the queue for processing.

As an addition the `updateQueues()` method has been removed as it is unnecessary logic. 

## Type of change

Choose a type of change, and delete any options that are not relevant.

Canary has been syncing continuously on Mainnet from a Local Snapshot 22k milestones behind.  

## Change checklist

### DIFF
--- a/src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
+++ b/src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
@@ -259,18 +259,7 @@ public class MilestoneSolidifierImpl implements MilestoneSolidifier {
      * @throws Exception
      */
     private void bootStrapSolidMilestones() throws Exception {
-        boolean solid = true;
-        int milestoneIndex = snapshotProvider.getInitialSnapshot().getIndex();
-
-        while (solid) {
-            milestoneIndex += 1;
-            if (MilestoneViewModel.get(tangle, milestoneIndex) == null) {
-                solid = false;
-                milestoneIndex -= 1;
-            }
-        }
-
-        setLatestSolidMilestone(milestoneIndex);
+        setLatestSolidMilestone(snapshotProvider.getLatestSnapshot().getIndex());
         Set<Hash> milestoneTransactions = AddressViewModel.load(tangle, config.getCoordinator()).getHashes();
         int processed = 0;
         int index;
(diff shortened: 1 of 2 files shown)


src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
```diff
@@ -281,9 +270,16 @@ private void bootStrapSolidMilestones() throws Exception {
...
                         .getMilestoneIndex(tvm)) > getLatestSolidMilestoneIndex()) {
-                    addMilestoneCandidate(hash, index);
-                }
+                    MilestoneValidity validity = milestoneService.validateMilestone(tvm, index);
+                    if (validity == MilestoneValidity.VALID) {
+                        tvm.isMilestone(tangle, snapshotProvider.getInitialSnapshot(), true);
+                        registerNewMilestone(getLatestMilestoneIndex(), index, tvm.getHash());
+                    }
 
+                    if (validity != MilestoneValidity.INVALID) {
```
GalRogozinski, line 279:
```suggestion
                   else if (validity != MilestoneValidity.INVALID) {
```

GalRogozinski: unless we should add candidates we registered for some reason?

src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
```diff
@@ -319,23 +315,9 @@ public void addMilestoneCandidate(Hash milestoneHash, int milestoneIndex) {
         if (!unsolidMilestones.containsKey(milestoneHash) && !seenMilestones.containsKey(milestoneIndex) &&
                 milestoneIndex > getLatestSolidMilestoneIndex()) {
             unsolidMilestones.put(milestoneHash, milestoneIndex);
-            updateQueues(milestoneHash, milestoneIndex);
```
GalRogozinski, line 322: shouldn't we still call `scanMilestonesInQueue`?

src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
GalRogozinski: this is wrong and I now understand

src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
DyrellC: That we should. A scan happens after each round of processing, but it's still best to scan on new additions as well.

PR #1856, opened 2020-05-07
Feature: Sync check added to broadcast stage

kwek20:
# Description of change

Extracts our sync checker to a service, and uses that service in the BroadcastStage of the Network pipeline.

This is required because if you get all your new transactions from 1 neighbour, and have 1 other neighbour which doesnt have the transaction you need, there is no chance of syncing.

If that second neighbour sends you a very low amount of new transactions(compared to the other neighbour) you will still sync, but very slow.

This has been introduced due to our removal of "ping pong" in #1745 

## Type of change

The stuck node got unstuck in the scenario described above

### DIFF
--- a/src/main/java/com/iota/iri/service/milestone/impl/MilestoneInSyncService.java
+++ b/src/main/java/com/iota/iri/service/milestone/impl/MilestoneInSyncService.java
@@ -6,6 +6,8 @@ import com.iota.iri.service.snapshot.SnapshotProvider;
 
 import com.google.common.annotations.VisibleForTesting;
 
+import java.util.concurrent.atomic.AtomicBoolean;
+
 /**
  * 
  * A node is defined in sync when the latest snapshot milestone index and the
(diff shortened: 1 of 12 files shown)


src/main/java/com/iota/iri/service/milestone/impl/MilestoneInSyncService.java
```diff
@@ -0,0 +1,73 @@
...
+     * To prevent jumping back and forth in and out of sync, there is a buffer in between.
+     * Only when the latest milestone and latest snapshot differ more than this number, we fall out of sync
+     */
+    @VisibleForTesting
+    static final int LOCAL_SNAPSHOT_SYNC_BUFFER = 5;
+    
+    /**
+     * If this node is currently seen as in sync
+     */
+    private boolean isInSync;
```
GalRogozinski, line 31: Even though currently we have one thread per stage, let's do this correctly and make this variable volatile/atomic.. since supposedly with a small change we can have more than one broadcast thread

kwek20:
@GalRogozinski can we merge this, so i can easily merge 1.8.6 onto the progressbar PR?

PR #1860, opened 2020-05-13
Remove address scan in bootstrap

DyrellC:
# Description of change
No longer scan milestones via the coo address. Instead walk up from the latest solid milestone and add the existing milestones to the solidifier queue to process. 
 
## Type of change

Took an older local snapshot and ran the node with a db that is up to date. The milestones above the LS were processed and solidified faster and bootstrapping took much less time. 

## Change checklist

### DIFF
--- a/src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
+++ b/src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
@@ -272,6 +272,7 @@ public class MilestoneSolidifierImpl implements MilestoneSolidifier {
         int index = getLatestSolidMilestoneIndex();
         MilestoneViewModel nextMilestone;
 
+        log.info("Starting milestone bootstrapping...");
         while (milestonePresent) {
             index++;
             if ((nextMilestone = MilestoneViewModel.get(tangle, index)) != null) {
@@ -282,6 +283,7 @@ public class MilestoneSolidifierImpl implements MilestoneSolidifier {
             }
         }
 
+        log.info("Done bootstrapping milestones.");
         initialized.set(true);
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
```diff
@@ -268,33 +268,22 @@ private void solidifyLog() {
...
-            } catch(Exception e) {
-                log.error("Error processing existing milestone index", e);
+        boolean milestonePresent = true;
+        int index = getLatestSolidMilestoneIndex();
+        MilestoneViewModel nextMilestone;
+
+        log.info("Starting milestone bootstrapping...");
+        while (milestonePresent) {
+            index++;
+            if ((nextMilestone = MilestoneViewModel.get(tangle, index)) != null) {
```
GalRogozinski, line 278: fyi you can do it with less lines :-)
` if ((nextMilestone = MilestoneViewModel.get(tangle, ++index)) != null) {`

src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
```diff
@@ -268,33 +268,22 @@ private void solidifyLog() {
...
+                registerNewMilestone(getLatestMilestoneIndex(), index, nextMilestone.getHash());
+                addMilestoneCandidate(nextMilestone.getHash(), nextMilestone.index());
+            } else {
+                milestonePresent = false;
```
GalRogozinski, line 282: also you can use `break` and then save on a variable

it doesn't really matter here
maybe it matters on low-level code sometimes
just drawing your attention to different ways to do stuff

you don't have to change unless you want to

GalRogozinski:
closed in favor of #1861

PR #1862, opened 2020-05-19
Fix milestone solidification on reboot

DyrellC:
# Description of change
If a sync is midway through and the node is shut off, it's possible for milestone transactions to get lost in the mix on restart. The transaction object may be present but the milestone object has not been generated. This means that if the node tries to solidify a transaction that references this milestone, it will always skip requesting this transaction as it is already stored. 

This stops the milestones from making it into the solidification queue again. By making the `TransactionSolidifier` request milestones regardless of solidity, and adding an address scanning method for when synchronization stalls out on reboot, we can mitigate most issues that would pop up revolving around syncing and restarting during a sync. 

Also lowers the max size of the `TransactionSolidifier` processing queue to ensure we solidify from bottom up in a more efficient manner. 

## Type of change

I have started, stopped and restarted many times on mainnet from a Local Snapshot 13k milestones back. Stops and starts resumed syncing in almost all instances. 

## Change checklist

### DIFF
--- a/src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
+++ b/src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
@@ -297,10 +297,14 @@ public class MilestoneSolidifierImpl implements MilestoneSolidifier {
             }
         }
 
-        if (!transactionSolidifier.addMilestoneToSolidificationQueue(seenMilestones.get(lowestIndex)) &&
-                (seenMilestones.size() + unsolidMilestones.size()) <
-                        (getLatestMilestoneIndex() - getLatestSolidMilestoneIndex())) {
-            scanAddressHashes();
+        if (lowestIndex <= getLatestSolidMilestoneIndex()) {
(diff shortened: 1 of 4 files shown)


src/main/java/com/iota/iri/service/milestone/impl/MilestoneSolidifierImpl.java
```diff
@@ -281,6 +288,57 @@ private void bootStrapSolidMilestones() throws Exception {
...
 
+    /**
+     * Checks the seen milestones queue for the lowest index available, and tries to solidify it. If it is already
+     * solid but not syncing, there are milestone objects in the db that have not been processed through the solidifier
+     * (likely due to shutting down during synchronisation). If this is the case, an address scan is performed to
+     * find and process all present milestone transactions through the milestone solidifier.
+     */
+    private void checkOldestSeenMilestoneSolidity() {
+        int lowestIndex = 0;
+        for (int index: seenMilestones.keySet()) {
```
GalRogozinski, line 299: just that you know you could have done `seenMilestones.keySet().stream().min(Integer::compareTo)`

PR #1870, opened 2020-06-02
Update readme to encourage users to use hornet

GalRogozinski:
# Description of change

Encourage users to use hornet

## Type of change

- Documentation Fix

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,8 @@
 
 ---
 ## Notice
-This project is currently being deprecated in favor of [hornet](https://github.com/gohornet/hornet). Users are encouraged to migrate.
+This project is currently being deprecated in favor of [HORNET](https://github.com/gohornet/hornet). Users are encouraged to migrate.
+In order to migrate your existing DB to HORNET please follow the [guide](https://docs.iota.org/docs/hornet/1.0/guides/migrating-from-iri).
 
 ## About
 
--- a/README.md
+++ b/README.md
@@ -28,6 +28,8 @@
 </p>
 
 ---
+## Notice
+This project is currently being deprecated in favor of [hornet](https://github.com/gohornet/hornet). Users are encouraged to migrate.
 
 ## About
 


JakeSCahill (commented):
Looks good. 

Do we expect users to migrate in a particular way? Or do we recommend just deleting all current database files and starting from the latest snapshot?

GalRogozinski:
Good question @JakeSCahill 
I think we should supply the instructions
I would give both options

@acha-bill (or @achabill, don't remember the handle he currently uses), created the migration procedure. Let's wait for him to reply here

Then let's create a migration document, and add it as a link to the README

acha-bill:
Migration tool is here: https://github.com/acha-bill/iri-db-migration

@taraszka was testing this on some nodes.
Maybe we want to wait for his feedback before proceeding with documentation.

JakeSCahill:
So if I'm reading this right, you need to have a certain snapshot on the Hornet node already?

If you have very old transactions on your IRI node, Hornet won't be able to process them?

acha-bill:
>If you have very old transactions on your IRI node, Hornet won't be able to process them?

Yes. If txs from IRI are older than the local snapshot hornet is running from.

If hornet starts from global snapshot, then time validation will always pass.

JakeSCahill:
From what I can see, there aren't many DB dumps available: https://dbfiles.iota.org/?prefix=mainnet/hornet/

So the oldest transactions may only be from May of this year?

JakeSCahill (approved):

JakeSCahill:
Nevermind my last comment.

I have an open PR for our documentation portal, which includes a migration guide: https://github.com/iotaledger/documentation/pull/722

You can see it on staging here: https://docs.iota.works/docs/hornet/1.0/guides/migrating-from-iri

Maybe we want to link to that?

@acha-bill could you ping me when the tool is ready?

GalRogozinski:
So @acha-bill the tool works well?
Can we add a link to the the docs and close the PR?

taraszka:
@GalRogozinski - let's wait. The migration tool almost finished the migration (LSMI/LMI: 1439876/1452434) so we will have a really large database of hornet available (since index: 1050000) soon.
I'd suggest waiting for hornet-0.4.1 and then encourage users to use hornet (it should be available more/less same time when the first database of hornet be available at dbfiles.iota.org)

Thanks

